Issue #139, opened 2021-03-20
Talk app integration


mejo-:
Ideas:

* maybe automatically create a Talk conversation for each new Collective
* link to this conversation from Collectives
* a button to open a page in the Talk conversation
* allow to open a collective (with page list) in a talk conversation

juliushaertl:
@mejo- We had the idea to dive into this topic for 31, but no concrete discussion yet. Maybe something to think about a bit more so we can come up with some ideas on what we could implement in that area.

g-work:
When I add a group to a circle/team, add this team to a room and then remove the group from the circle, the group members stay participants of the room. Can someone confirm this?

See https://github.com/nextcloud/spreed/issues/13569

PR #442, opened 2022-12-13
Draft: create and open a talk room for the current collective

mejo-:
In GitLab by @azul on Aug 9, 2021, 11:01

It looks like we need to keep track of the room ourselves as there's no way to create and later identify a room for a circle other than going through the participants list of every room.

See #139.


mejo-:
Let's close this for now, given that we have other means to link collectives and talk rooms with the new Teams feature in Nextcloud 29.

PR #443, opened 2022-12-13
WIP: Periodically update page content for read mode from editor session

mejo-:
In GitLab by @mejo- on May 30, 2022, 14:00

If we have an editor with an open session, periodically (every two
seconds) update the page content from that session.

That way, users can stay in read mode when they merely want to follow
what others type.


mejo-:
Let's close this PR. This will be tackled differently by switching to the Text Editor API soonish.

PR #445, opened 2022-12-13
Build parent path without queries

### LINKED ISSUE #377: Performance issues
In GitLab by @ognarb on Aug 16, 2022, 12:06

I profiled a bit collectives on our perftesting instance and /index.php/apps/collectives/_api/12/_pages call is quite bad. One of the reasons for this is that the parent are not cached in PageService::getParentPageId. This then created 900 DB requests

One idea how to fix it is to do something similar to https://github.com/nextcloud/end_to_end_encryption/pull/316 (who was also affected by the same bug) or maybe even try to finish https://github.com/nextcloud/server/pull/31793 so that apps don't have to the caching themselves and duplicate the caching (this is also quite a bit more risky)

mejo-:
In GitLab by @juliushaertl on Aug 18, 2022, 13:28

- Build parent node name from path without getting the parent node 
- Pass parent node where available to avoid extra db queries
- Cache root storage id per request to avoid duplicate queries

Cuts down the queries in half on our perftesting instance for the handbook example collective: 948 -> 541

Related upstream fix in server, but is not fully applicable here due to the recursive directory iteration https://github.com/nextcloud/server/pull/33602

Fixes #377

### DIFF
--- a/lib/Model/PageInfo.php
+++ b/lib/Model/PageInfo.php
@@ -87,16 +87,18 @@ class PageInfo extends Entity implements JsonSerializable {
 	public function fromFile(File $file, int $parentId, ?string $lastUserId = null, ?string $lastUserDisplayName = null, ?string $emoji = null, ?string $subpageOrder = null): void {
 		$this->setId($file->getId());
 		// Set folder name as title for all index pages except the collective landing page
+		$dirName = dirname($file->getInternalPath());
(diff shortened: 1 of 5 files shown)


juliushaertl:
@mejo- Tests pass now and this would be ready for review. I just can't request your review due to the migration from gitlab ;)

PR #448, opened 2022-12-13
Add app config to specify default app folder location

tcitworld:
If the admin does it it's expected they know what they're doing and the mount point should work.


mejo- (approved):
Thanks for the PR, looks sensible :ok_hand:

Issue #138, opened 2021-03-20
Open pages+collectives in collectives app from files app


mejo-:
When opening a collective page from the files app, the page should be opened in our collectives app automatically.

Since we don't have a special mimetype as selector, it needs to be chosen depending on the path.

To be decided if we want this per default or as an alternative option from the files app menu.

PR #461, opened 2022-12-21
Display info with a link to Collectives in Files app

mejo-:
Inside collectives folders in the Files app, display an info box and a button to navigate to the Collectives app instead.

Fixes: #138

### 📝 Summary

* Resolves: #138

#### 🖼️ Screenshots

2022-12-21T15:23:32,786517496+01:00

### 🏁 Checklist

### DIFF
--- a/src/helpers/files.js
+++ b/src/helpers/files.js
@@ -1,0 +1,56 @@
+import Vue from 'vue'
+import FileListInfo from '../views/FileListInfo.vue'
+import axios from '@nextcloud/axios'
+import { generateOcsUrl } from '@nextcloud/router'
+
+const FilesCollectivesPlugin = {
+	el: null,
+
+	async getCollectivesFolder() {
+		const response = await axios.get(generateOcsUrl('apps/collectives/api/v1.0/settings/user/user_folder'))
+		return response.data.ocs.data
+	},
+
+	attach(fileList) {
+		if (fileList.id !== 'files' && fileList.id !== 'files.public') {
+			return
+		}
+
+		this.el = document.createElement('div')
+		this.el.id = 'files-collectives-info-wrapper'
+		fileList.registerHeader({
+			id: 'collectives',
+			el: this.el,
+			render: this.render.bind(this),
+			// Higher order than Text rich workspaces
+			order: -1,
+		})
+	},
+
+	async render(fileList) {
+		const collectivesFolder = await this.getCollectivesFolder()
+		const View = Vue.extend(FileListInfo)
+		Vue.prototype.t = window.t
+		Vue.prototype.n = window.n
+		const vm = new View({
+			propsData: {
+				collectivesFolder,
+				path: fileList.getCurrentDirectory(),
+			},
+		}).$mount(this.el)
+
+		fileList.$el.on('urlChanged', data => {
+			vm.path = data.dir.toString()
+		})
+
+		fileList.$el.on('changeDirectory', data => {
+			vm.path = data.dir.toString()
+		})
+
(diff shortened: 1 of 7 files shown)


src/helpers/files.js
```diff
@@ -0,0 +1,49 @@
+import Vue from 'vue'
+import store from '../store/store.js'
```
mejo-, line 2: I'm a bit unsure whether it's worth pulling in the full Collectives store just for using `GET_COLLECTIVES_FOLDER` from it (see below) or whether it's better to duplicate a bit of code and reimplement `GET_COLLECTIVES_FOLDER` in order to not load the Collectives store each time the Files app is loaded :thinking:

src/views/FileListInfo.vue
```diff
@@ -0,0 +1,131 @@
...
+
+	mounted() {
+		this.getCollectivesFolder().then(() => {
+			this.setActive()
+		})
+	},
+
+	methods: {
+		...mapActions({
+			dispatchGetCollectivesFolder: GET_COLLECTIVES_FOLDER,
```
mejo-, line 74: This is what the store is needed for.

src/helpers/files.js
juliushaertl: Yep, I'd say it rather makes sense to inject an initial state value when you load the additional files app script and just use that value then.

src/views/FileListInfo.vue
```diff
@@ -0,0 +1,131 @@
...
+		}
+	},
+
+	computed: {
+		...mapState({
+			collectivesFolder: (state) => state.settings.collectivesFolder,
+		}),
+
+		collectivesLink() {
+			return generateUrl('/apps/collectives' + this.path.replace(this.collectivesFolder, ''))
```
juliushaertl, line 56: Maybe check if the string starts with and slice it then to avoid issues when the collectivesFolder name is used somewhere in the page path

mejo-: Good catch, fixed now :+1:

src/helpers/files.js
mejo-: Ok, agreed. I moved the logic to get the `collectivesFolder` to `helpers/files.js`.

src/helpers/files.js
```diff
@@ -0,0 +1,56 @@
...
+		fileList.registerHeader({
+			id: 'collectives',
+			el: this.el,
+			render: this.render.bind(this),
+			// Higher order than Text rich workspaces
+			order: -1,
+		})
+	},
+
+	async render(fileList) {
```
mejo-, line 30: I'm a bit unsure whether it's acceptable to make `render()` a async function. While it works as expected, I have a bad gut feeling about it. What do you think @juliushaertl @max-nextcloud?

Maybe you have another idea how to implement the async `getCollectivesFolder()` function and still make sure that `render()` doesn't pass a promise in `propsData`?

juliushaertl: You could also pass the collective folder as initial state in the Listener for the files additional scripts. Then you could directly perform the check without waiting for another request.

juliushaertl (approved):
Looks good code wise, only comment would be to use initial state instead of the async request

src/helpers/files.js
mejo-: Oh, great idea, thanks for the hint! To be honest I didn't know about initial state before :see_no_evil:

Implemented now in https://github.com/nextcloud/collectives/pull/461/commits/6fd1f7304756e596be2272ee15f0ceb4637a0c4b

PR #471, opened 2022-12-28
Add API to register extra collective action for other apps

mejo-:
Other apps can listen to CollectivesLoadAdditionalScriptsEvent to load additional JS scripts when the Collectives app is rendered.

If they define an object OCA.Collectives.CollectiveExtraAction with string `title` and callback function `click`.

### 🏁 Checklist

### DIFF
--- a/lib/Controller/StartController.php
+++ b/lib/Controller/StartController.php
@@ -2,6 +2,7 @@
 
 namespace OCA\Collectives\Controller;
 
+use OCA\Collectives\Events\CollectivesLoadAdditionalScriptsEvent;
 use OCA\Viewer\Event\LoadViewer;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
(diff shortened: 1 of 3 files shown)


mejo-:
This is to be used by the [pandoc app](https://github.com/mejo-/pandoc).

juliushaertl (approved):
Looks good 👍 

Maybe as a general comment, for now the API is kind of single use, so that only one app can register and any further registrations would override. If there are other use cases you could consider providing a registration method for actions and keep them in an array so that multiple apps can provide actions.

Besides the page ids it might also be helpful to pass more metadata about the collective (like the name and collective id)

mejo-:
> Looks good +1
> 
> Maybe as a general comment, for now the API is kind of single use, so that only one app can register and any further registrations would override. If there are other use cases you could consider providing a registration method for actions and keep them in an array so that multiple apps can provide actions.
> 
> Besides the page ids it might also be helpful to pass more metadata about the collective (like the name and collective id)

I agree @juliushaertl. I decided to keep it simple for now for the concrete usecase I had in mind. But we could expand it as soon as more usecases pop up :relaxed:

PR #473, opened 2022-12-28
 Makefile: migrate release targets from Gitlab to Github

mejo-:


juliushaertl (approved):
Seems good, but we should also consider moving releases to https://github.com/nextcloud-releases

PR #481, opened 2023-01-02
ci: Use apcu and speed up basic auth for cypress/integration tests

juliushaertl:
### 📝 Summary

Speed up CI a bit. Let's see how the numbers are since collectives has quite a good coverage :)

Screenshot 2023-01-02 at 18 20 12

### 🏁 Checklist

### DIFF
--- a/.github/workflows/behat-sqlite.yml
+++ b/.github/workflows/behat-sqlite.yml
@@ -74,7 +74,9 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php-versions }}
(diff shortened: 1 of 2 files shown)


max-nextcloud (approved):
code looks good ( and familiar ;) )

max-nextcloud:
I think we need to split up the pages spec. It takes so long all other runners are already done and bored. It probably also only does auth once and then reuses it.

grafik

We could also use 2 less machines as cypress points out.
I'll take a look at the pages spec. I think this is still good to go.

julien-nc (approved):
I didn't know you could have so few things in config.php when installing :grin:.

Issue #468, opened 2022-12-23
Deleting page with attachment fails


mejo-:
**Describe the bug**
Delete a page with attachments is not possible. That's because the backend detects the attachments folder as additional content in the folder of the respective page and therefore denies to delete the page.

**To Reproduce**
1. Create a page with attachments
2. Try to delete that page
3. See the error message

**Expected behavior**
It should be possible to delete pages which attachments.

PR #487, opened 2023-01-05
Ignore attachments folder when deleting/renaming pages

mejo-:
Fixes: #468

### 🏁 Checklist

### DIFF
--- a/lib/Service/PageService.php
+++ b/lib/Service/PageService.php
@@ -292,7 +292,7 @@ class PageService {
 					$this->revertSubFolders($node);
 				} elseif ($node instanceof File) {
 					// Move index page without subpages into the parent folder (if's not the landing page)
-					if (self::isIndexPage($node) && !self::isLandingPage($node) && !$this->pageHasOtherContent($node)) {
+					if (self::isIndexPage($node) && !self::isLandingPage($node) && !$this->indexPageHasOtherContent($node)) {
 						$filename = NodeHelper::generateFilename($folder, $folder->getName(), PageInfo::SUFFIX);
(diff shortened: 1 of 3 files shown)


lib/Service/PageService.php
```diff
@@ -371,7 +371,9 @@ public function pageHasOtherContent(File $file): bool {
 					!self::isIndexPage($node)) {
 					return true;
 				}
-				if ($node->getName() !== PageInfo::INDEX_PAGE_TITLE . PageInfo::SUFFIX) {
+				// Ignore the page file itself and corresponding attachments folder
+				if ($node->getName() !== PageInfo::INDEX_PAGE_TITLE . PageInfo::SUFFIX
```
max-nextcloud, line 375: This could also use the `isIndexPage function, right?
```suggestion
				if (!self::isIndexPage($node))
```

max-nextcloud: I tried to understand this function on its own and noticed it does not work for normal pages - only for index pages because otherwise the `getParent()->getDirectoryListing()` in 368 will fetch all sibling pages.

I see that this is only used on index pages. Still maybe rename to `indexPageHasOtherContent`, document this limitation or add a check
```php
if (!self::isIndexPage($file)) {
  return false;
}
```
before the foreach.

max-nextcloud (approved):
Looks good to me.

Thanks for tackling this. Only minor comments.

lib/Service/PageService.php
mejo-: Thanks, both valid points. I renamed the function to `indexPageHasOtherContent` and used `self::isIndexPage()` inside it now.

PR #500, opened 2023-01-10
Make sure the collective setting modal is closed after deleting

mejo-:


juliushaertl (approved):
Code makes sense but haven't tested

PR #510, opened 2023-01-19
Refactor: Move TextEditor handling in its own component

mejo-:
Preparations for using the new editor API from Text, see https://github.com/nextcloud/text/pull/3615

### DIFF
--- a/src/components/Page/TextEditor.vue
+++ b/src/components/Page/TextEditor.vue
@@ -209,7 +209,6 @@ export default {
 		async getPageContent() {
 			this.pageContent = await this.fetchPageContent(this.currentPageDavUrl)
 			if (!this.pageContent) {
-				console.debug('emit start-edit in TextEditor.vue')
 				this.$emit('start-edit')
 			}
 		},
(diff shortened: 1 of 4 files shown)


src/components/Page/TextEditor.vue
```diff
@@ -0,0 +1,247 @@
...
+	watch: {
+		'editMode'(val) {
+			if (val === true) {
+				this.startEdit()
+			} else {
+				this.stopEdit()
+			}
+		},
+
+		/*
```
juliushaertl, line 95: Still relevant?

juliushaertl (approved):
Change looks good. I added two follow up requirements from reading the diff to the text pr checklist:

- Things that collectives triggers
  - Save through syncService.save()
  - focusEditor

src/components/Page/TextEditor.vue
mejo-: Thanks, removed now :blush:

max-nextcloud (approved):
Tested to confirm ui issue when pressing enter quickly after creating a new page is fixed.

PR #527, opened 2023-01-31
Add reference provider and basic reference widget

### LINKED ISSUE #509: Custom link preview and link picker implementation
When sharing a link to a collective page it would be nice to have a custom rendering for that. In addition having a link picker provider would be nice so you could easily add a link to pages in integrating apps like talk.

julien-nc:
closes #509 

Hey how. Here is a reference provider using the existing search provider.
It comes with a basic reference widget.

Not sure about the `PageReferenceProvider::resolveReference` implementation which might be a bit brutal but I didn't find a simpler way to get the page and the collective entities.

Also not sure about the widget content. Feel free to completely change it.

I took the liberty to display the emojis in the search results which I find nicer but that's just a suggestion. Feel free to get rid of this.

One more doubt: I didn't find an easy way to resolve the link without the file ID (when the fileId GET param is not there). Should we make it possible to resolve without the page file ID?

### The widget

image

### The link picker provider in action

image

### DIFF
--- a/lib/Reference/PageReferenceProvider.php
+++ b/lib/Reference/PageReferenceProvider.php
@@ -1,0 +1,222 @@
+<?php
+
+namespace OCA\Collectives\Reference;
+
+use DateTime;
+use OC\Collaboration\Reference\LinkReferenceProvider;
+use OCA\Collectives\Model\CollectiveInfo;
+use OCA\Collectives\Service\CollectiveService;
+use OCA\Collectives\Service\PageService;
+use OCP\Collaboration\Reference\ADiscoverableReferenceProvider;
+use OCP\Collaboration\Reference\ISearchableReferenceProvider;
+use OCP\Collaboration\Reference\Reference;
+use OC\Collaboration\Reference\ReferenceManager;
+use OCA\Collectives\AppInfo\Application;
+use OCP\Collaboration\Reference\IReference;
+use OCP\Files\File;
+use OCP\IConfig;
+use OCP\IDateTimeFormatter;
+use OCP\IL10N;
+
+use OCP\IURLGenerator;
+
+class PageReferenceProvider extends ADiscoverableReferenceProvider implements ISearchableReferenceProvider {
+	private const RICH_OBJECT_TYPE = Application::APP_NAME . '_page';
+
+	private ?string $userId;
+	private IConfig $config;
+	private ReferenceManager $referenceManager;
+	private IL10N $l10n;
+	private IURLGenerator $urlGenerator;
+	private LinkReferenceProvider $linkReferenceProvider;
+	private PageService $pageService;
+	private CollectiveService $collectiveService;
+	private IDateTimeFormatter $dateTimeFormatter;
+
+	public function __construct(IConfig $config,
+								CollectiveService $collectiveService,
+								PageService $pageService,
+								IL10N $l10n,
+								IURLGenerator $urlGenerator,
+								IDateTimeFormatter $dateTimeFormatter,
+								ReferenceManager $referenceManager,
+								LinkReferenceProvider $linkReferenceProvider,
+								?string $userId) {
+		$this->userId = $userId;
+		$this->config = $config;
+		$this->referenceManager = $referenceManager;
+		$this->l10n = $l10n;
+		$this->urlGenerator = $urlGenerator;
+		$this->linkReferenceProvider = $linkReferenceProvider;
+		$this->pageService = $pageService;
+		$this->collectiveService = $collectiveService;
+		$this->dateTimeFormatter = $dateTimeFormatter;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getId(): string {
+		return Application::APP_NAME . '-ref-pages';
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getTitle(): string {
+		return $this->l10n->t('Collective pages');
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getOrder(): int {
+		return 10;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getIconUrl(): string {
+		return $this->urlGenerator->getAbsoluteURL(
+			$this->urlGenerator->imagePath(Application::APP_NAME, 'collectives-dark.svg')
+		);
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getSupportedSearchProviderIds(): array {
+		return ['collectives-pages'];
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function matchReference(string $referenceText): bool {
+		if ($this->userId === null) {
+			return false;
+		}
+		$start = $this->urlGenerator->getAbsoluteURL('/apps/' . Application::APP_NAME);
+		$startIndex = $this->urlGenerator->getAbsoluteURL('/index.php/apps/' . Application::APP_NAME);
+
+		// link example:
+		// https://nextcloud.local/apps/collectives/supacollective/Tutos/Hacking/Spectre?fileId=14457
+		// https://nextcloud.local/apps/collectives/supacollective/Tutos/Hacking/Spectre
+		$noIndexMatch = preg_match('/^' . preg_quote($start, '/') . '\/[^\/]+\//i', $referenceText) === 1;
+		$indexMatch = preg_match('/^' . preg_quote($startIndex, '/') . '\/[^\/]+\//i', $referenceText) === 1;
+
+		return $noIndexMatch || $indexMatch;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function resolveReference(string $referenceText): ?IReference {
+		if ($this->matchReference($referenceText)) {
+			$pageInfo = $this->getPagePathFromDirectLink($referenceText);
+			if ($pageInfo !== null) {
+				// TODO should it work if the fileId GET param is not there?
+				if (isset($pageInfo['fileId'])) {
+					$pageFileId = $pageInfo['fileId'];
+					$collectiveName = $pageInfo['collectiveName'];
+					$collectives = $this->collectiveService->getCollectives($this->userId);
+					$collectives = array_filter($collectives, static function (CollectiveInfo $c) use ($collectiveName) {
+						return $c->getName() === $collectiveName;
+					});
+					if (!empty($collectives)) {
+						$collective = end($collectives);
+						$pageInfo['collective'] = $collective;
+						$collectiveFolder = $this->pageService->getCollectiveFolder($collective->getId(), $this->userId);
+						$pageFile = $collectiveFolder->getById($pageFileId);
+						if (!empty($pageFile) && isset($pageFile[0]) && $pageFile[0] instanceof File) {
+							$pageFile = $pageFile[0];
+							$pageInfo['page'] = $this->pageService->findByFile($collective->getId(), $pageFile, $this->userId);
+
+							$reference = new Reference($referenceText);
+
+							$reference->setTitle($pageInfo['page']->getEmoji() . ' ' . $pageInfo['page']->getTitle());
+
+							$description = $this->l10n->t('In collective %1$s %2$s', [$collective->getEmoji(), $collective->getName()])
+								. ' - ' . $pageInfo['page']->getFilePath();
+							$reference->setDescription($description);
+							$pageInfo['description'] = $description;
+
+							$date = new DateTime();
+							$date->setTimestamp($pageInfo['page']->getTimestamp());
+							$formattedRelativeDate = $this->dateTimeFormatter->formatTimeSpan($date);
+							$pageInfo['lastEdited'] = $this->l10n->t('Last edition %1$s', [$formattedRelativeDate]);
+
+							$imageUrl = $this->urlGenerator->getAbsoluteURL(
+								$this->urlGenerator->imagePath(Application::APP_NAME, 'page.svg')
+							);
+							$reference->setImageUrl($imageUrl);
+
+							$pageInfo['link'] = $referenceText;
+							$reference->setUrl($referenceText);
+
+							$reference->setRichObject(
+								self::RICH_OBJECT_TYPE,
+								$pageInfo,
+							);
+							return $reference;
+						}
+					}
+				}
+			}
+			// fallback to opengraph if it matches but somehow we can't resolve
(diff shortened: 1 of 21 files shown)


julien-nc:
Oh I just noticed when looking at the screenshots: in the search results subline, `in Collective ...`, you are using an uppercase C. I'm assuming it comes from the German in which nouns begin with an uppercase letter. Right? Would you agree that it's not the case in English?

src/reference.js
```diff
@@ -0,0 +1,42 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { registerWidget } from '@nextcloud/vue-richtext'
+import Vue from 'vue'
+import VTooltip from '@nextcloud/vue/dist/Directives/Tooltip.js'
+import PageReferenceWidget from './views/PageReferenceWidget.vue'
+
+Vue.directive('Tooltip', VTooltip)
```
juliushaertl, line 27: No need for custom tooltips anymore ;) there was the general decision to move to browser native ones for accessibility

src/views/PageReferenceWidget.vue
```diff
@@ -0,0 +1,123 @@
...
+	white-space: normal;
+	padding: 12px;
+	display: flex;
+
+	a {
+		padding: 0 !important;
+		color: var(--color-main-text) !important;
+		text-decoration: unset !important;
+		&:hover {
+			color: #58a6ff !important;
```
juliushaertl, line 103: Should probably not be a hardcoded color value

src/views/PageReferenceWidget.vue
```diff
@@ -0,0 +1,123 @@
...
+			type: Object,
+			default: null,
+		},
+		accessible: {
+			type: Boolean,
+			default: true,
+		},
+	},
+
+	data() {
```
juliushaertl, line 78: 🔥

src/views/PageReferenceWidget.vue
```diff
@@ -0,0 +1,123 @@
...
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<template>
+	<div class="collective-page">
```
juliushaertl, line 23: May want to add a v-if since richObject can be null

juliushaertl (commented):
🥳

src/views/PageReferenceWidget.vue
julien-nc: :+1:

src/views/PageReferenceWidget.vue
julien-nc: --color-primary-element-light-text?

src/reference.js
julien-nc: :+1: VTooltip is not even used in the widget...

juliushaertl:
- [ ] Check if this is backward compatible since collectives supports multiple versions

julien-nc:
Rebased on main

src/views/PageReferenceWidget.vue
juliushaertl: --color-primary-text-dark would be the server link hover/focus style

julien-nc: Hmmm ok let's forget about the color, fine with the server values. 
How about removing the text decoration (underline) when it is not hovered?

juliushaertl:
@julien-nc Could you check https://github.com/nextcloud/collectives/pull/527#issuecomment-1411808138 as well?

julien-nc:
@juliushaertl Right, this won't work with 25 as `ADiscoverableReferenceProvider` and `ISearchableReferenceProvider` were introduced in 26.

How do we proceed here? Should we duplicate the reference provider, one for 25 for resolving only and the full one for 26+? And a check on the NC version in Application.php? Not pretty but I don't see any alternative...except bumping the min NC version to 26...

juliushaertl:
> How do we proceed here? Should we duplicate the reference provider, one for 25 for resolving only and the full one for 26+? And a check on the NC version in Application.php? Not pretty but I don't see any alternative...except bumping the min NC version to 26...

Yep, that sounds reasonable. We can have the 26 provider extend the 25 one and just add the interface on top.

julien-nc:
> > How do we proceed here? Should we duplicate the reference provider, one for 25 for resolving only and the full one for 26+? And a check on the NC version in Application.php? Not pretty but I don't see any alternative...except bumping the min NC version to 26...
> 
> Yep, that sounds reasonable. We can have the 26 provider extend the 25 one and just add the interface on top.

As we have to extend ADiscoverableReferenceProvider, we can't also extend the other provider classe. Polymorphism can only be done with interfaces in Php, right?

I pushed them with code duplication.

julien-nc:
PhpUnit failure is caused by the new `OC\Files\Storage\Wrapper\Wrapper::filesize` method signature in 26 which is only compatible with Php >= 8.0 while Collectives main branch supports 25.

Anyway, not related.

mejo-:
> 

So far we used uppercase "Collectives" when we mean the app itself, but lowercase "collective(s)" when we're speaking about concrete collectives (e.g. "your collective abc"). I think it makes sense to treat app names as names and make them uppercase and I think that's done in a lot of places in Nextcloud, but not sure whether it's something we have an agreement about.

lib/Reference/PageReferenceProvider.php
```diff
@@ -0,0 +1,180 @@
...
+						$pageFile = $collectiveFolder->getById($pageFileId);
+						if (!empty($pageFile) && isset($pageFile[0]) && $pageFile[0] instanceof File) {
+							$pageFile = $pageFile[0];
+							$pageInfo['page'] = $this->pageService->findByFile($collective->getId(), $pageFile, $this->userId);
+
+							$reference = new Reference($referenceText);
+
+							$reference->setTitle($pageInfo['page']->getEmoji() . ' ' . $pageInfo['page']->getTitle());
+
+							$description = $this->l10n->t('In collective %1$s %2$s', [$collective->getEmoji(), $collective->getName()])
```
mejo-, line 98: Same: `getEmoji()` might return an empty string.

lib/Search/PageProvider.php
```diff
@@ -110,4 +115,28 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
 		);
 	}
+
+	/**
+	 * @param PageInfo $pageInfo
+	 * @return string
+	 */
+	private function getPageTitle(PageInfo $pageInfo): string {
+		$emoji = $pageInfo->getEmoji();
+		if ($emoji) {
```
mejo-, line 125:
```suggestion
		return $emoji
			? $emoji . ' ' . $pageInfo->getTitle()
			: $pageInfo->getTitle();
```

src/views/PageReferenceWidget.vue
```diff
@@ -0,0 +1,117 @@
...
+  -->
+
+<template>
+	<div v-if="richObject" class="collective-page">
+		<div class="collective-page--image">
+			<span v-if="emoji"
+				class="page-emoji">
+				{{ emoji }}
+			</span>
+			<CollectivesIcon v-else
```
mejo-, line 29: If this is the page emoji, it probably should display the PageIcon for pages without emojis, not the CollectivesIcon, right?

src/views/PageReferenceWidget.vue
```diff
@@ -0,0 +1,117 @@
...
+		},
+		accessible: {
+			type: Boolean,
+			default: true,
+		},
+	},
+
+	computed: {
+		emoji() {
+			return this.richObject.page.emoji ?? this.richObject.collective.emoji ?? null
```
mejo-, line 81: This is confusing in my eyes. We should only use the page emoji here in my opinion.

mejo- (changes requested):
Thanks a lot for working on this @julien-nc :heart: It's already looking great and I'm really looking forward to use this myself :heart_eyes:

I did a first code review and made some comments. Especially in the backend implementation I'd prefer a bit of refactoring. I can do it myself if you lack the time to do so.

In any case maybe we can have a quick call so you can guide me through the logic of `ADiscoverableReferenceProvider` and  `ADiscoverableReferenceProvider` a bit.

lib/Reference/PageReferenceProvider.php
```diff
@@ -0,0 +1,180 @@
...
+	public function resolveReference(string $referenceText): ?IReference {
+		if ($this->matchReference($referenceText)) {
+			$pageInfo = $this->getPagePathFromDirectLink($referenceText);
+			if ($pageInfo !== null) {
+				// TODO should it work if the fileId GET param is not there?
+				if (isset($pageInfo['fileId'])) {
+					$pageFileId = $pageInfo['fileId'];
+					$collectiveName = $pageInfo['collectiveName'];
+					$collectives = $this->collectiveService->getCollectives($this->userId);
+					$collectives = array_filter($collectives, static function (CollectiveInfo $c) use ($collectiveName) {
```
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Reference/PageReferenceProvider.php
```diff
@@ -0,0 +1,180 @@
...
+
+		return $noIndexMatch || $indexMatch;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function resolveReference(string $referenceText): ?IReference {
+		if ($this->matchReference($referenceText)) {
+			$pageInfo = $this->getPagePathFromDirectLink($referenceText);
```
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Reference/PageReferenceProvider.php
```diff
@@ -0,0 +1,180 @@
...
+			if ($pageInfo !== null) {
```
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Reference/PageReferenceProvider.php
```diff
@@ -0,0 +1,180 @@
...
+						return $c->getName() === $collectiveName;
+					});
+					if (!empty($collectives)) {
+						$collective = end($collectives);
+						$pageInfo['collective'] = $collective;
+						$collectiveFolder = $this->pageService->getCollectiveFolder($collective->getId(), $this->userId);
+						$pageFile = $collectiveFolder->getById($pageFileId);
+						if (!empty($pageFile) && isset($pageFile[0]) && $pageFile[0] instanceof File) {
+							$pageFile = $pageFile[0];
+							$pageInfo['page'] = $this->pageService->findByFile($collective->getId(), $pageFile, $this->userId);
```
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Reference/SearchablePageReferenceProvider.php
```diff
@@ -0,0 +1,218 @@
...
+use OC\Collaboration\Reference\ReferenceManager;
+use OCA\Collectives\AppInfo\Application;
+use OCP\Collaboration\Reference\IReference;
+use OCP\Files\File;
+use OCP\IDateTimeFormatter;
+use OCP\IL10N;
+
+use OCP\IURLGenerator;
+
+class SearchablePageReferenceProvider extends ADiscoverableReferenceProvider implements ISearchableReferenceProvider {
```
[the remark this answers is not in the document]
julien-nc: * `PageReferenceProvider` can only be used with Nextcloud 25. It only includes the "reference widget" stuff (to render link previews)
* `SearchablePageReferenceProvider` can be used with Nextcloud >= 26. It contains `PageReferenceProvider` plus the link picker stuff:
    * discoverable means it will be listed as a provider in the link picker
    * searchable means it has to declare a list of supported unified search providers (that will be used by the link picker once "Collectives" is selected)

lib/Search/PageProvider.php
```diff
@@ -110,4 +115,28 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
+			return $emoji . ' ' . $pageInfo->getTitle();
+		}
+		return $pageInfo->getTitle();
+	}
+
+	/**
+	 * @param CollectiveInfo $collectiveInfo
+	 * @return string
+	 */
+	private function getCollectiveName(CollectiveInfo $collectiveInfo): string {
```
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Search/PageProvider.php
```diff
@@ -110,4 +115,28 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
+		$emoji = $collectiveInfo->getEmoji();
+		if ($emoji) {
```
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Search/PageProvider.php
```diff
@@ -110,4 +115,28 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
+			return $emoji . ' ' . $collectiveInfo->getName();
+		}
+		return $collectiveInfo->getTitle();
```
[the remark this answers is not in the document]
julien-nc: Now you know all my test collectives have an emoji...

lib/Reference/PageReferenceProvider.php
```diff
@@ -0,0 +1,180 @@
...
+
+							$reference = new Reference($referenceText);
+
+							$reference->setTitle($pageInfo['page']->getEmoji() . ' ' . $pageInfo['page']->getTitle());
```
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Reference/PageReferenceProvider.php
julien-nc: :+1:

lib/Search/PageProvider.php
julien-nc: :+1:

src/views/PageReferenceWidget.vue
julien-nc: :+1:

src/views/PageReferenceWidget.vue
julien-nc: :+1:

mejo-:
Thanks a lot @julien-nc! I pushed unit tests for the CollectiveService functions and rebased.

mejo- (approved):

PR #540, opened 2023-02-08
fix: Adapt return type for Nextcloud 26

juliushaertl:
Changed in https://github.com/nextcloud/server/pull/36120

Should still be compatible with 25 as adding the return type on the child class is ok if none is there on the parent.

### DIFF
--- a/lib/ACL/ACLStorageWrapper.php
+++ b/lib/ACL/ACLStorageWrapper.php
@@ -177,7 +177,7 @@ class ACLStorageWrapper extends Wrapper {
 		return parent::filetype($path);
 	}
 
-	public function filesize($path) {
+	public function filesize($path): false|int|float {
 		if (!$this->checkPermissions(Constants::PERMISSION_READ)) {
 			return false;
 		}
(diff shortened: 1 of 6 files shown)


max-nextcloud (approved):

max-nextcloud (commented):
Good catch! But it turns out this is possible in stable25 - but not in php 7.4 which stable25 still supports. So i don't know what to do here.

juliushaertl:
cc @come-nc as you added the return type on the storage wrapper. Any idea how we could stay compatible with 7.4 still? Otherwise apps implementing a wrapper would be forced to release separate versions for 25/26.

juliushaertl:
Found a way by having two separate classes and let psalm ignore the error for php 7.4. Not really nice but at least save us from needing to branch off for Nextcloud 26. @mejo- What do you think?

max-nextcloud:
@juliushaertl sounds great! I was wondering if something like this would be possible.
I think the Lint / php-cs failure and the PHPUnit failures are still related to this though.

come-nc:
Way too many applications got hit by this change in OC.
It seems some of files classes should move to OCP, if so many applications depend on them anyway…

Issue #543, opened 2023-02-10
Printing: Collective title always on blank first page


susnux:
**Describe the bug**
When printing a collective the title is always printed as the only element on the first page.
So there is no other content, as the main collective content is printed starting from page 2.

This happens because of the `position: sticky` styling of the title.

**To Reproduce**
1. Go to collective
2. Press <kbd>CTRL</kbd> + <kbd>P</kbd>
4. See first page only contains the title

**Expected behavior**
Title and collective are on the same page / first page not wasted.

**Screenshots**
![title on first page, content on the second](https://user-images.githubusercontent.com/1855448/218117325-a8190e50-915e-47a6-bccc-df69f492af00.png)

**Client details:**
 - OS: Linux
 - Browser: Firefox
 - Version: 109
 - Device: desktop

<details>
<summary>Server details</summary>

**Collectives app version:** 2.2.1

**Operating system:** Docker image

**Web server:** apache

**Database:** mariadb

**PHP version:** 8.x

**Nextcloud version:** 25.0.3

</details>

Issue #542, opened 2023-02-10
NC25: Printing includes background image


susnux:
**Describe the bug**
The background image is not fully removed when printing a page.

**To Reproduce**
1. Go to collective
2. press print page
3. (select print to pdf)
4. see background image is included

**Expected behavior**
Only the collective content is printed not the background image.

**Screenshots**
Print view:
![background image is still visible on the borders](https://user-images.githubusercontent.com/1855448/218115643-4c145586-4b60-4a9f-b7f8-55a376f0427d.png)

Or print dialog:
![print dialog, border includes the background image](https://user-images.githubusercontent.com/1855448/218115926-3de4788e-eff3-44bd-9f59-80795549b43c.png)

**Client details:**
 - OS: Linux
 - Browser: Firefox
 - Version: 109
 - Device: desktop

<details>
<summary>Server details</summary>

**Collectives app version:** 2.2.1

**Operating system:** Docker (ubunut?)

**Web server:** apache

**Database:** mariadb

**PHP version:** 8.0

**Nextcloud version:** 25.0.3

</details>

mejo-:
@susnux I cannot reproduce this bug. Regardless whether I use the "Export of print" function to open the print view, or whether I directly print using <kbd>Ctrl</kbd>-<kbd>P</kbd>, no background is included in the printed document :thinking:

Are you still able to reproduce this issue?

susnux:
> Are you still able to reproduce this issue?

Now it is only a problem if you select "print backgrounds" in the browser printing dialog.

So this now could be intentionally, but I still think we should remove any non-text/collectives background when printing?

Issue #463, opened 2022-12-21
Improved export/dialog dialog


mejo-:
The idea is to transform the print view into a modal with different export options and a print option. The loading feedback while export is generated can be similar to the current splash screen of the print view.

mejo-:
Thanks @thopanX. Can you elaborate what you mean with header/footer? Do you mean the title of the page in question?

mejo-:
Related to #469

max-nextcloud:
@thopanX I think that header is added by the print dialogue of your browser and can be turned off. (See the blue checkmark below):

![grafik](https://user-images.githubusercontent.com/97337118/209465613-aaa76dc2-5590-413f-ace7-715022422309.png)
(`more settings` in print dialogue of firefox)

I'm not sure if we can make it configurable because it mainly consistes of the page number, date and url. So things that we cannot change easily.

max-nextcloud:
>> I'm not sure if we can make it configurable because it mainly consistes of the page number, date and url. So things that we cannot change easily.

> But what can be improved would be the name of the export, right? It currently doesn't mention the name of the collective, just „Kollektive – Nextcloud“.

Yes. I think that is the page title. That should be fairly simple actually.

max-nextcloud:
Related to #375

PR #544, opened 2023-02-10
Listen on printing shortcut to open print view for printing

susnux:
### 📝 Summary

* Resolves: #542 
* Resolves: #543 

Listen on <kbd>CTRL</kbd> + <kbd>P</kbd> (or <kbd>CMD</kbd> respectivly), then cancel printing the collective but open the print view to the correct styling get applied.
Moreover I changed the target name of the new tab / window, so the printing view gets recycled for all printing tasks to not spam the users tabs.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
 | image

### 🏁 Checklist

### DIFF
--- a/src/components/Collective/CollectiveActions.vue
+++ b/src/components/Collective/CollectiveActions.vue
@@ -50,7 +50,7 @@
 		<NcActionSeparator v-if="collectiveCanShare(collective) && !isPublic" />
 		<NcActionLink :close-after-click="true"
 			:href="printLink"
-			target="_blank">
+			target="ncCollectivesPrint">
 			{{ t('collectives', 'Export or print') }}
 			<template #icon>
 				<DownloadIcon :size="20" />
@@ -142,6 +142,7 @@ export default {
 			'isCollectiveAdmin',
 			'isPublic',
 			'loading',
+			'printLink',
 		]),
 
 		circleLink() {
@@ -165,6 +166,10 @@ export default {
 			return this.isPublic
 				? generateUrl(`/apps/collectives/p/${this.shareTokenParam}/print/${this.collective.name}`)
 				: generateUrl(`/apps/collectives/_/print/${this.collective.name}`)
+		},
+
+		unshareIcon() {
+			return this.loading('unshare') ? 'icon-loading-small' : 'icon-public'
 		},
 	},
 
--- a/src/router.js
+++ b/src/router.js
@@ -37,11 +37,13 @@ const routes = [
 		path: '/_/print/:collective',
 		component: CollectivePrintView,
 		props: (route) => route.params,
+		children: [{ path: ':page*' }],
 	},
 	{
 		path: '/p/:token/print/:collective',
 		component: CollectivePrintView,
 		props: (route) => route.params,
+		children: [{ path: ':page*' }],
 	},
 	{
 		path: '/p/:token/:collective',
--- a/src/store/pages.js
+++ b/src/store/pages.js
@@ -171,6 +171,14 @@ export default {
 			return state.pages.find(p => (p.parentId === parentId && p.title === TEMPLATE_PAGE))
 		},
 
+		pagePrintLink: (state, get) => (page) => {
+			const path = [page.collectivePath.split('/').at(-1), page.filePath.split('/'), page.fileName]
+			if (path.at(-1) === 'Readme.md') path.splice(-1, 1)
+			if (get.isPublic) path.splice(2, 0, 'print')
+			else path.splice(0, 0, '_', 'print')
+			return generateUrl(`/apps/collectives/${path.join('/')}`)
+		},
+
 		currentFileIdPage(state, _getters, rootState) {
 			const fileId = Number(rootState.route.query.fileId)
 			return state.pages.find(p => (p.id === fileId))
--- a/src/views/CollectiveView.vue
+++ b/src/views/CollectiveView.vue
@@ -19,7 +19,9 @@
 <script>
 
 import { mapGetters, mapMutations } from 'vuex'
+import { showWarning } from '@nextcloud/dialogs'
 import { NcAppContent, NcEmptyContent, NcLoadingIcon } from '@nextcloud/vue'
+
 import Collective from '../components/Collective.vue'
 import CollectiveNotFound from '../components/CollectiveNotFound.vue'
 import PageList from '../components/PageList.vue'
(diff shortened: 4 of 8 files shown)


src/views/CollectiveView.vue
```diff
@@ -39,12 +39,32 @@ export default {
...
+	},
+
 	methods: {
 		...mapMutations(['hide']),
+
+		/**
+		 * @param {KeyboardEvent} event the keydown event
+		 */
+		printKeyHandler(event) {
+			if ((event.metaKey || event.ctrlKey) && event.key.toLowerCase() === 'p') {
```
max-nextcloud, line 62: `ctrl+shift+p` opens a new private window in Firefox.
Let's make sure to not overwrite this as it's unrelated.

max-nextcloud (changes requested):
Looks good to me. Just one minor fix to prevent overwriting other keyboard shortcuts.

(Have not tried it - just from the sourcecode it seems ctrl+shift+P migth also be affected.)

src/components/Collective/CollectiveActions.vue
```diff
@@ -43,7 +43,7 @@
 		<NcActionSeparator v-if="collectiveCanShare(collective) && !isPublic" />
 		<NcActionLink :close-after-click="true"
 			:href="printLink"
-			target="_blank">
+			target="ncCollectivesPrint">
```
mejo-, line 46: Not sure what's the effect here? :thinking: Also, for me (with Firefox) this results in an error: `[Vue warn]: Invalid prop: custom validator check failed for prop "target".`

src/views/CollectiveView.vue
```diff
@@ -39,12 +41,38 @@ export default {
...
 	methods: {
 		...mapMutations(['hide']),
+
+		/**
+		 * @param {KeyboardEvent} event the keydown event
+		 */
+		printKeyHandler(event) {
+			// Handle `CTRL+P` or `CMD+P` but ensure ALT or SHIFT are NOT pressed (e.g. CTRL+SHIFT+P is new private tab on firefox)
+			if ((event.metaKey || event.ctrlKey) && event.key.toLowerCase() === 'p' && !(event.altKey || event.shiftKey)) {
+				const handle = window.open(this.printLink, 'ncCollectivesPrint').focus()
```
mejo-, line 66: Unfortunately, this doesn't work for me with Firefox: `Uncaught TypeError: window.open(...) is null`

mejo- (changes requested):
Thanks @susnux! This would make it impossible to print single pages though, no? Or am I missing something?

src/views/CollectiveView.vue
susnux: This happens if you have a popup blocker (see `if`-`else` below)...
The `.focus()` is required to go into the `else` part (only execute it if a handle is given) :see_no_evil:

Fixed this.

src/components/Collective/CollectiveActions.vue
susnux: The effect should be that the new tab / window is reused, `_blank` always opens a new window, but if you give a unique string, then that tab is used for all `target="myName"` links.
See MDN on this: [developer.mozilla.org#reuse_existing_windows_and_avoid_target_blank](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#reuse_existing_windows_and_avoid_target_blank)

But it looks like the `NcActionLink` has a custom validator only allowing `_...`, so this needs to be fixed in the vue components library. (That error is raised for development build).

mejo-: Ah, that makes a lot of sense. Thanks for the explanation :heart:

mejo-:
> Thanks @susnux! This would make it impossible to print single pages though, no? Or am I missing something?

@susnux what's your take on the question of printing a single page?

src/store/store.js
```diff
@@ -48,6 +49,10 @@ export default new Store({
 
 		isPublic: (_state, get) =>
 			!!get.shareTokenParam,
+
+		printLink: (_state, get) => get.isPublic
+			? generateUrl(`/apps/collectives/p/${get.shareTokenParam}/print/${get.currentCollective.name}`)
```
mejo-, line 54: `get.currentCollective` might be unset when you don't have a collective open. Also, this would always print the current collective. So better to pass the collective as parameter here and use it.

mejo- (changes requested):

susnux:
:facepalm: I did not noticed that you currently print the whole collective with subpages...

The idea was to not duplicate places with printing specific styling. So forward to the export view, I am currently thinking about resolving this by adding a property to the print view to only print the current page, do you think this is a acceptable solution?

mejo-:
> The idea was to not duplicate places with printing specific styling. So forward to the export view, I am currently thinking about resolving this by adding a property to the print view to only print the current page, do you think this is a acceptable solution?

Yeah, in the long run I would like to add a dedicated export view that allows to print or export pages to PDF. The default should be to export the whole collective, but it should be possible to (un)select specific pages. Unfortunately this is not very high on the priority list. See #463.

Your intermediate solution to allow printing only a specific page in the print view sounds good to me!

susnux:
I am currently try to work on this again, I think the best way to solve this is to add subpage support for the export view as well. But this will require quite some work on that view (as I am not that into the collectives app structure). I hope to get this working this month :)

susnux:
It is now possible to use the print view to only print subpages, this is now also used for `CTRL+P` and there is a print menu entry in the pages action menu.

susnux:
@mejo- Any idea why only one cypress ci was successfully?

susnux:
> @mejo- Any idea why only one cypress ci was successfully?

Never mind CI is green after rebasing :tada:

pjrobertson:
The images in the initial PR post look like this also solves the problem of the content being moved on to the 2nd page when it's longer than 1 page (no page break). I find that every collective I try to print longer than a specific amount has this issue - the title on the first page and content on the 2nd page. Hopefully this can be merged soon! 

Screen Region 2023-08-08 at 18 26 12 1
Screen Region 2023-08-08 at 18 26 16 1

src/components/Collective/CollectiveActions.vue
```diff
@@ -142,6 +142,7 @@ export default {
 			'isCollectiveAdmin',
 			'isPublic',
 			'loading',
+			'printLink',
```
mejo-, line 145: This seems like a leftover, `printLink()` is defined as computed property below and not available as getter from the store.

src/components/Collective/CollectiveActions.vue
```diff
@@ -166,6 +167,10 @@ export default {
 				? generateUrl(`/apps/collectives/p/${this.shareTokenParam}/print/${this.collective.name}`)
 				: generateUrl(`/apps/collectives/_/print/${this.collective.name}`)
 		},
+
+		unshareIcon() {
```
mejo-, line 171: Seems like a leftover or copy/paste error as well. `unshareIcon()` is not used here.

src/store/pages.js
```diff
@@ -171,6 +171,14 @@ export default {
 			return state.pages.find(p => (p.parentId === parentId && p.title === TEMPLATE_PAGE))
 		},
 
+		pagePrintLink: (state, get) => (page) => {
+			const path = [page.collectivePath.split('/').at(-1), page.filePath.split('/'), page.fileName]
+			if (path.at(-1) === 'Readme.md') path.splice(-1, 1)
```
mejo-, line 176: We don't use these one-line `if () ...` syntax so far, I'd prefer to stick to if () { \n ... \n }` :grimacing: 

Same for the following two lines.

src/views/CollectiveView.vue
```diff
@@ -39,13 +41,43 @@ export default {
...
 	methods: {
 		...mapMutations(['hide']),
+
+		/**
+		 * @param {KeyboardEvent} event the keydown event
+		 */
+		printKeyHandler(event) {
+			// Handle `CTRL+P` or `CMD+P` but ensure ALT or SHIFT are NOT pressed (e.g. CTRL+SHIFT+P is new private tab on firefox)
+			if ((event.metaKey || event.ctrlKey) && event.key.toLowerCase() === 'p' && !(event.altKey || event.shiftKey)) {
+				if (!this.currentPage) return
```
mejo-, line 68: Same here.

mejo- (changes requested):
I gave this a quick test. Unfortunately I ran into several issues:

* It doesn't work for first-level pages because they somehow get a double slash in the URL.
* It doesn't work for second-level pages (and first-level pages with the double slash removed). Console error is: `this.currentPage is undefined`.

src/router.js
```diff
@@ -37,11 +37,13 @@ const routes = [
 		path: '/_/print/:collective',
 		component: CollectivePrintView,
 		props: (route) => route.params,
+		children: [{ path: ':page*' }],
```
mejo-, line 40: Can you give me a quick hint what you want to achieve here? Maybe I oversaw something, but I don't see where this would be used :thinking:

mejo-:
Closing this PR now that we have fixes from #971.

PR #547, opened 2023-02-10
fix(cypress): selector for mention autocompletion

max-nextcloud:
* Fixes CI failures such as https://github.com/nextcloud/collectives/actions/runs/4103191125/jobs/7077067180
  (Please note that the retries fail for a different reason)

#### 🖼️ Screenshots

Page -- Editing a page -- Supports page content editing and switching to read mode (failed)

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [x] Documentation is not required as this is a fix in tests

### DIFF
--- a/cypress/e2e/pages.spec.js
+++ b/cypress/e2e/pages.spec.js
@@ -227,7 +227,7 @@ describe('Page', function() {
 			cy.getEditor()
 				.should('be.visible')
 				.type('@admi')
-			cy.get('.tippy-content > .items')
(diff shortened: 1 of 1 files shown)


susnux (approved):
Could also be changed to `.tippy-content > .suggestion-list` but I do not think we need this extra level of specificity.

mejo-:
Nice, thanks for looking into this @max-nextcloud :heart:

Issue #548, opened 2023-02-10
Weird borders on create collectives input element


susnux:
**Describe the bug**
Borders between submit and input of the `create collective` input are rounded which loos odd.
Also on errors only the submit part has red (error) border color.

**Expected behavior**
1. On errors the whole border should have the error color.
2. Borders between input and submit elements should not be rounded

**Screenshots**
![only the button part of the input element has red borders, the rest has default ones, borders are weirdly rounded](https://user-images.githubusercontent.com/1855448/218152091-5a72d17e-f797-494e-99c0-a686446efb80.png)

<details>
<summary>Server details</summary>

**Collectives app version:** 2.2.1

**Operating system:** Docker

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 8.x

**Nextcloud version:** 25.0.3

</details>

PR #551, opened 2023-02-10
Fix `new collective` name input elements

susnux:
### 📝 Summary

* Resolves: #548 

This resolves some styling issues, like hovering the submit input does not set the border on the text input or mismatching borders. Moreover the `NcMultiselect` is deprecated and should be replaced with the `NcSelect`.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
only submit button has error color, wrong border radius | both elements input and submit have correct borders
hovering only highlights the submit button | correct border color on whole element on hover
. | submit button with correct borders
circle select separated from submit button | submit button part of circle select
multselect no elements | NcSelect no results

Please note that the missing primary color border on hover will be fixed with the next vue components update (see https://github.com/nextcloud/nextcloud-vue/pull/3701 )

### 🏁 Checklist

### DIFF
--- a/src/components/Nav/NewCollective.vue
+++ b/src/components/Nav/NewCollective.vue
@@ -33,51 +33,65 @@
 			</NcActionButton>
 		</template>
 	</NcAppNavigationItem>
-	<div v-else class="collective-create">
-		<form v-show="editing" @submit.prevent.stop="createCollective">
-			<NcEmojiPicker :show-preview="true" @select="addEmoji">
-				<NcButton type="tertiary"
-					:aria-label="t('collectives', 'Select emoji for collective')"
-					:title="t('collectives', 'Select emoji')"
-					class="button-emoji"
-					@click.prevent>
-					{{ emoji }}
-				</NcButton>
-			</NcEmojiPicker>
+	<form v-else
+		v-show="editing"
+		class="collective-create"
+		@submit.prevent.stop="createCollective">
+		<NcEmojiPicker :show-preview="true" @select="addEmoji">
+			<NcButton type="tertiary"
+				:aria-label="t('collectives', 'Select emoji for collective')"
+				:title="t('collectives', 'Select emoji')"
+				class="button-emoji"
+				@click.prevent>
+				{{ emoji }}
+			</NcButton>
+		</NcEmojiPicker>
 
-			<input v-if="!pickCircle"
-				ref="nameField"
-				v-model="text"
-				:placeholder="t('collectives', 'New collective name')"
-				type="text"
-				required>
-			<NcMultiselect v-else
-				ref="circleSelector"
+		<NcTextField v-if="!pickCircle"
+			ref="nameField"
+			:value.sync="text"
+			:error="text === ''"
+			:label="t('collectives', 'New collective name')"
+			:show-trailing-button="!!text"
+			required
+			trailing-button-icon="arrowRight"
+			:trailing-button-label="t('collectives', 'Create new collective')"
+			@trailing-button-click="createCollective" />
+
+		<div v-else class="circle-selector--container">
+			<NcSelect ref="circleSelector"
 				v-model="circle"
+				:loading="loading"
 				:options="circles"
+				:searchable="!circle"
 				open-direction="below"
-				:placeholder="t('collectives', 'Select circle...')"
+				:placeholder="t('collectives', 'Select circle …')"
 				required />
-			<input type="submit"
-				value=""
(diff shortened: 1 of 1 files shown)


max-nextcloud:
I'd suggest to still merge this. It fixes ui bugs. If the ui is replaced even before the next release so be it. Maybe the modal takes longer than expected, maybe we want to use this input again later on. Even if not - i don't see any downside to merging this. But maybe I am missing something, @mejo-

src/components/Nav/NewCollective.vue
```diff
@@ -33,51 +33,60 @@
 			</NcActionButton>
 		</template>
 	</NcAppNavigationItem>
-	<div v-else class="collective-create">
-		<form v-show="editing" @submit.prevent.stop="createCollective">
```
mejo-, line 37: Removing the `<form>` element with its submit event means that the form cannot be submitted by pressing <kbd>Return</kbd> anymore.

src/components/Nav/NewCollective.vue
```diff
@@ -33,51 +33,60 @@
...
+				:title="t('collectives', 'Select emoji')"
+				class="button-emoji"
+				@click.prevent>
+				{{ emoji }}
+			</NcButton>
+		</NcEmojiPicker>
+
+		<NcTextField v-if="!pickCircle"
+			:value.sync="text"
+			:error="!text"
```
mejo-, line 49: Not sure whether I like that the input field is marked as erroneous from the very start. Visual feedback is nice, but it feels wrong to me to have the error border displayed before even starting to type something.

src/components/Nav/NewCollective.vue
```diff
@@ -33,51 +33,60 @@
...
+			:label="t('collectives', 'New collective name')"
+			:show-trailing-button="!!text"
+			required
+			trailing-button-icon="arrowRight"
+			:trailing-button-label="t('collectives', 'Create new collective')"
+			@trailing-button-click="createCollective" />
+
+		<div v-else class="circle-selector--container">
+			<NcSelect ref="circleSelector"
```
mejo-, line 58: Not sure whether it's the switch from `NcMultiSelect` to `NcSelect`, but when selecting a circle with a longer name, the select field size grows larger than it should. Also the cancel and expand/collapse buttons inside the select field are not properly aligned.

mejo- (changes requested):
After some first testing, I identified the following issues.

src/components/Nav/NewCollective.vue
susnux: Reverted, does work now :)

src/components/Nav/NewCollective.vue
susnux: Me neither... decided to restore the previous behavior and only show the error state if the user started typing

src/components/Nav/NewCollective.vue
susnux: I can not confirm this, for me (Firefox 109, collectives on current NC master) it works as expected and looks like this:
image

mejo-: Ah, and it only grows bigger when closing the dropdown. With the dropdown open it keeps its desired size, like in your screenshot.

mejo-: Also, for some reason the selection is dark even when I use the light theme :thinking:

mejo-: > Also, for some reason the selection is dark even when I use the light theme thinking

Nevermind this. It seems to be a bug in nextcloud-vue. At least all other NcSelect fields are dark as well for me when my browser prefers dark theme but Nextcloud is set to light theme.

susnux: Yes this is known and already fixed, waiting for release: https://github.com/nextcloud/nextcloud-vue/pull/3707

susnux: > Ah, and it only grows bigger when closing the dropdown. With the dropdown open it keeps its desired size, like in your screenshot.

I pushed a commit which fixes parts of this issue. The other  parts are fixed with: https://github.com/nextcloud/nextcloud-vue/pull/3759

> Nevermind this. It seems to be a bug in nextcloud-vue. At least all other NcSelect fields are dark as well for me when my browser prefers dark theme but Nextcloud is set to light theme.

Fixed if we upgrade `@nextcloud/vue` to the new released `7.6.0`.

mejo- (approved):

mejo-:
Thanks a lot @susnux :heart:

Issue #559, opened 2023-02-13
Page related three-dot menu missing without edit permissions


mejo-:
**Describe the bug**
When a page is opened without edit permissions, the three-dot menu for page-context actions is missing. That's why it's impossible to toggle the outline view in this case. See https://github.com/nextcloud/collectives/issues/410#issuecomment-1428094105

**Collectives app version:** 2.2.1

PR #565, opened 2023-02-15
Readonly view fixes in page titlebar

mejo-:
### 📝 Summary

* Don't load emoji picker if collective is readonly
* Show stripped-down page actions menu in readonly mode
* Resolves: #559

### DIFF
--- a/cypress/e2e/collective-readonly.spec.js
+++ b/cypress/e2e/collective-readonly.spec.js
@@ -32,9 +32,10 @@ describe('Read-only collective', function() {
 	})
 
 	describe('in read-only collective', function() {
-		before(function() {
+		beforeEach(function() {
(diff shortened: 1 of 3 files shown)


max-nextcloud (approved):
Had a look at the code. Did not try it locally.
Might be good to check for the page menu in https://github.com/nextcloud/collectives/blob/main/cypress/e2e/collective-readonly.spec.js to avoid a regression as we don't look at read only collectives so often.

mejo-:
> Might be good to check for the page menu in https://github.com/nextcloud/collectives/blob/main/cypress/e2e/collective-readonly.spec.js to avoid a regression as we don't look at read only collectives so often.

Good point, added a test now.

PR #586, opened 2023-03-01
Cypress: Refactor link handling tests

mejo-:
Make sure that the search part of URLs matches the expected value.


mejo-:
This failure should go away once https://github.com/nextcloud/text/issues/3857 is fixes: https://github.com/nextcloud/collectives/actions/runs/4306118932/jobs/7509507368#step:19:179

Issue #135, opened 2021-03-20
Attachment management (handle non-markdown files)


mejo-:
We have to decide on how our app handles non-markdown files.

Issue #322, opened 2022-04-26
Attachments design review


mejo-:
In GitLab by @jancborchardt on Apr 26, 2022, 15:08

* [ ] Sidebar tab
  * [x] Lists media in the document, more like the Talk media tab and not interactive with the ability to upload attachments which are not referenced anywhere
  * [ ] Wording: "Media"? Does not include files necessarily. Or "Files & media", or just "Files" is also fine
  * [x] Sorted chronologically, most recently added on top
  * [x] Hint at the bottom "Add media to the document using drag & drop or via 'upload media'"
  * [x] Have "View in document" action in the 3-dot menu, first place
  * [x] Document titles are ellipsized very early
  * [ ] Future: Look at how Talk does it
* [x] Case of deleting media
  * [x] Deleting via Text document is undo-able, until you end the Text session via "Done"
  * [x] Remove the "Delete attachment" from the action menu
  * [x] "Recently deleted" section in media, for quickly restoring recently deleted
  * [x] When restoring, put the file back into the document at cursor position and scroll to it
* [x] "Upload image" action rename to "Upload media" or "Upload files & media" depending on how we word the sidebar tab
* [x] Adding already existing media: Don’t prepend timestamp. If there is a duplicate, append a counter like in files, starting with " (2)" like "New document (2).txt"
* [x] Move drag & drop style to a Vue component – compare Files & Mail for example
* [x] Possibly add a subline to title with "3 backlinks · 4 files" which on click opens the sidebar (ref Notion: <https://www.notion.so/help/create-links-and-backlinks> )

mejo-:
All implemented in #603

PR #603, opened 2023-03-16
Add attachments list to page sidebar

mejo-:
### 📝 Summary

Add attachments list to page sidebar. Clicking on the attachment will open it in the viewer. Additional actions allow to scroll to the attachments in the document and to open it in the Files app.

* Resolves: #135, #322

#### 🖼️ Screenshots

grafik

#### Screencasts

[recording.webm](https://user-images.githubusercontent.com/3582805/226637302-2a6207c7-0e35-4ad5-864f-90d7ac2f7c7b.webm)

---

[recording2.webm](https://user-images.githubusercontent.com/3582805/226972777-2f494f87-36f6-41b7-b720-d172b1be1591.webm)

### 🚧 TODO

- [x] Find out why no `notify_files` event is triggered when uploading a file to Text
- [x] Emit event from Text when deleting an attachment from the document, listen to it and move the attachment to a "Recently deleted" section. Add option to restore the attachment in the document. (Probably moved to a follow-up)
- [x] When restoring, put the file back into the document at cursor position and scroll to it (Probably should be  to a follow-up)

### 🏁 Checklist


nimishavijay (changes requested):
Super nice work! The flow already looks great :) Great placement of the attachments and love the scrolling! Some feedback from my side:

- [x] Is there any reason we need to show the heading "Attachments on this page" specifically? Unless there are other types of attachments being shown, I would say we can remove that heading
- [x] "Add attachments to the document using drag & drop or via "Insert attachment"" --> "Add attachments using drag and drop or via "Insert attachment" in the formatting bar" (remove "to the document", use "and" instead of `&`, indicate where the "Insert attachment" button is present)
- [x] In view mode the "Add attachments to the document using..." info box can be removed, since attachments can't be added unless you are in edit mode (I assume)
- [x] When you click on "View in document", if possible we could also highlight that row for a few seconds in `--color-primary-light`? Slightly similar to the behaviour when you click on a link to a message in Talk. (nice to have)
- [x] in the future it may also be nice to have a "Download" feature for each attachment (nice to have)


- [ ] for the design of each attachment item, a good reference could be the "Versions" tab in the sidebar in Files
image
The design points that we can adopt from the versions style:
  - [ ] More spacing between each item
  - [ ] More space between thumbnail and info
  - [ ] More space between the title and subline
  - [ ] Rounded corners on hover

The main points that are different from the versions style would be:
- [ ] Subline info order would be reversed: it would be `size` · `date` since size would be the more relevant info here as opposed to date for versions
- [ ] Title does not have to be bold, it looked nice without bolding in the screenshot you have posted 

Questions that I have:
- [ ] How are the attachments sorted? I would expected them to be sorted by their position in the document
- [x] What happens if you click on an attachment in the sidebar? I would expect that if it is a supported filetype it opens in the viewer, else it is downloaded


This is really great work! cc @jancborchardt for any more feedback :)

mejo-:
Thanks a lot for the thorough review @nimishavijay, much appreciated :heart: 

Concerning your two questions:

>  How are the attachments sorted? I would expected them to be sorted by their position in the document

Well, sorting them according to their position in the document would be hard, would need to parse the document in order to do so. For now they're ordered chronologically, newest first.

>  What happens if you click on an attachment in the sidebar? I would expect that if it is a supported filetype it opens in the viewer, else it is downloaded.

For now, it's opened in the viewer if possible and in Files app otherwise. I'll change it to your suggestion once I implemented the download feature.

mejo-:
I think I adressed all your remarks @nimishavijay (apart from sorting by occurance).

nimishavijay (approved):
Looks perfect now! Love this feature! :)

lib/Service/AttachmentService.php
```diff
@@ -0,0 +1,93 @@
...
+	 */
+	private function fileToInfo(File $file, string $userId): array {
+		try {
+			return [
+				'id' => $file->getId(),
+				'name' => $file->getName(),
+				'filesize' => $file->getSize(),
+				'mimetype' => $file->getMimeType(),
+				'timestamp' => $file->getMTime(),
+				'path' => substr($file->getPath(), strlen('/' . $userId . '/files')),
```
juliushaertl, line 36: Not fully sure if this works for all cases, another option would be to use the user folder and call getRelativePath here, but fine with me if this approach also works.

src/components/Page/RichText.vue
```diff
@@ -85,6 +92,7 @@ export default {
 			'currentPageDirectory',
 			'currentPageFilePath',
 			'isPublic',
+			'loading',
```
juliushaertl, line 95: Seems to be not introduced here but would be nice to have the naming a bit more clear to tell what loading state this is. Not blocking.

src/components/PageSidebar.vue
```diff
@@ -2,27 +2,29 @@
 	<NcAppSidebar ref="sidebar"
 		:title="title"
 		@close="close">
-		<NcAppSidebarTab id="backlinks"
+		<NcAppSidebarTab id="attachments"
```
juliushaertl, line 5: Is the attachment sidebar supported on public shares? From the component below I assumed not, then we may want to hide it.

src/components/PageSidebar/SidebarTabAttachments.vue
```diff
@@ -0,0 +1,331 @@
...
+				return '%' + c.charCodeAt(0).toString(16).toUpperCase()
+			})
+		},
+
+		filesUrl() {
+			return (fileId) => generateUrl(`/f/${fileId}`)
+		},
+
+		davUrl() {
+			return (attachment) => generateRemoteUrl(`dav/files/${getCurrentUser().uid}/${this.currentPage.collectivePath}/${encodeURI(attachment.internalPath)}`)
```
juliushaertl, line 173: This would not be compatible with public shares of collectives I guess

src/components/PageSidebar/SidebarTabAttachments.vue
```diff
@@ -0,0 +1,331 @@
...
+		previewUrl() {
+			return (attachment) => {
+				return attachment.hasPreview
+					? this.attachmentPreview(attachment)
+					: OC.MimeType.getIconUrl(attachment.mimetype ?? 'undefined')
+			}
+		},
+
+		attachmentPreview() {
+			return (attachment) => (attachment.id ? generateUrl(`/core/preview?fileId=${attachment.id}&x=64&y=64&a=true`) : null)
```
juliushaertl, line 185: Would also not work for public shares.

juliushaertl (commented):
Code looks good, with some questions regarding public shares and smaller non-blocking comments.

juliushaertl:
😍 Looks very nice from the screenshots, but haven't tested yet.

src/components/Page/RichText.vue
mejo-: It's a generic object holding the loading state for anything in Collectives, see https://github.com/nextcloud/collectives/blob/c391ee9766cc0f7b67381722f1e49e8c5fba995e/src/store/store.js#L30 and https://github.com/nextcloud/collectives/blob/c391ee9766cc0f7b67381722f1e49e8c5fba995e/src/store/store.js#L56-L57

juliushaertl: All right, thanks for the insight :)

juliushaertl (approved):

Issue #613, opened 2023-03-24
WARNING: Data vanishes when circle is deleted


ludenticus:
Using `Collectives` now in a production environment is extremely risky. It should be warned that **the app will wipe out all your data** whenever you delete its associated circle. If you have a backup of your Documents, you will recover the `.md` files, but you will inevitably lose your Collective.

It's a mayor design flaw. At least the owner should be able to _recreate_ the Collective.

mejo-:
Dear @ludenticus, thanks for your report. I agree with your assessment. There's also #314 to track this problem. I hope to find time to work on this in the upcoming weeks.

Issue #314, opened 2022-04-02
Prevent circles of collectives from being deleted in Contacts


mejo-:
Right now, it's easy to accidently destroy a collective by deleting its underlying circle from the Contacts app.

Fortunately, Circles recently [added a new feature to flag a circle as managed by an app](https://github.com/nextcloud/circles/pull/974) which disallows it to remove the circle from the Contacts app. We should set this flag for circles of collectives.

Once new Nextcloud versions with this feature have been released, we should probably also add a migration step to set this flag for all circles of existing collectives.

PR #617, opened 2023-03-27
Flag circles for collectives as managed by app

mejo-:
Fixes: #314
Fixes: #613


ArtificialOwl (approved):
looks good to me

juliushaertl:
How would those circles then be removed finally? Or would they stick around forever?

mejo-:
> When removing the circle along with the collective in Collectives, the circle will be gone. But good point. if you remove the collective without the circle, it indeed would no longer be removable through UI. I'll change this and unflag the circle as app maintained when removing a collective. Then it can be removed through Contacts app later on.

This is done now. But for some reason there's sometimes issues with the `(un)flagAsAppManaged()` function in `CircleManager`. In CI it seems like it doesn't persist, and in the repair step it neither does on my local dev instance. Will have to do some further investigation.

Issue #466, opened 2022-12-21
Use skeleton loading screens instead of loading spinner for page list and content


mejo-:
Instead of showing the loading spinner, we should use sekleton loading screens like in Talk and Mail:

* For list of collectives
* For page list
* For page content

max-nextcloud:
In Talk (spreed) the component used is [`LoadingPlaceholder`](https://github.com/nextcloud/spreed/blob/bae297e41d15fa864d0c453b310af9eb1fce18f5/src/components/LoadingPlaceholder.vue).
It looks slightly different then our lists because it has two lines on the right and is a bit bigger to mimic the conversations in talk.
![Bildschirmfoto vom 2023-01-05 17-45-15](https://user-images.githubusercontent.com/97337118/210835073-ed5acc86-04e2-49a6-aa68-bddefce51725.png)

Still i think it could be a good starting point.

PR #649, opened 2023-04-18
feat: add component SkeletonLoading

luka-nextcloud:
### 📝 Summary

* Resolves: #466

#### 🖼️ Screenshots
image

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/PageList.vue
+++ b/src/components/PageList.vue
@@ -1,6 +1,5 @@
 <template>
-	<NcAppContentList :class="{loading: loading('collective') || loading('pagelist')}"
-		:show-details="showing('details')">
+	<NcAppContentList :show-details="showing('details')">
 		<div class="page-list-headerbar">
 			<NcTextField name="pageFilter"
 				:label="t('collectives', 'Search pages')"
@@ -55,7 +54,10 @@
 				</NcActionButton>
 			</NcActions>
 		</div>
-		<div v-if="currentCollective && collectivePage" class="page-list">
+		<div v-if="loading('collective') || loading('pagelist')" class="page-list">
+			<SkeletonLoading type="items" :count="3" />
+		</div>
+		<div v-if="!loading('collective') && currentCollective && collectivePage" class="page-list">
 			<Item key="Readme"
 				:to="currentCollectivePath"
 				:page-id="collectivePage.id"
@@ -127,11 +129,13 @@ import PagesTemplateIcon from './Icon/PagesTemplateIcon.vue'
 import { SET_COLLECTIVE_USER_SETTING_PAGE_ORDER } from '../store/actions.js'
 import { scrollToPage } from '../util/scrollToElement.js'
 import { pageOrders } from '../util/sortOrders.js'
+import SkeletonLoading from './SkeletonLoading.vue'
 
 export default {
 	name: 'PageList',
 
 	components: {
+		SkeletonLoading,
 		NcActions,
 		NcActionButton,
 		NcAppContentList,
--- a/src/components/PageList.vue
+++ b/src/components/PageList.vue
@@ -54,10 +54,10 @@
 				</NcActionButton>
 			</NcActions>
 		</div>
-		<div v-if="loading('collective') || loading('pagelist')" class="page-list">
+		<div v-if="!currentCollective || !collectivePage || loading('collective')" class="page-list">
 			<SkeletonLoading type="items" :count="3" />
 		</div>
-		<div v-if="!loading('collective') && currentCollective && collectivePage" class="page-list">
+		<div v-else class="page-list">
 			<Item key="Readme"
 				:to="currentCollectivePath"
 				:page-id="collectivePage.id"
--- a/src/components/SkeletonLoading.vue
+++ b/src/components/SkeletonLoading.vue
@@ -32,10 +32,10 @@
 						class="text-placeholder"
 						xmlns="http://www.w3.org/2000/svg"
 						:fill="'url(#placeholder-gradient' + suffix + ')'">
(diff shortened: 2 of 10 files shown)


juliushaertl:
I think we should make sure that the skeleton actually matches the icon size and text lines that are there when the actual content is shown. 

- left sidebar: small icon with one text line next to it (matching the exact position of the icon/text)
- page list: same as left sidebar
- page content: small box for icon, one line text for the heading, one line skeleton for the metadata, 3 lines of text only blocks for the page content

For some visual presentation, orange boxes would roughly show which skeleton placeholders we'd want to have:

Screenshot 2023-04-24 at 16 04 46

max-nextcloud:
@luka-nextcloud Could you fix the lint errors and upload a new screenshot please? That would make it easier to review the change. Thanks!

juliushaertl:
@mejo- had some concerns if it wouldn't be better to reuse the existing component stylesheets for building the placeholder which I think make sense. @luka-nextcloud Please pair up with @mejo- on this one to align on how to implement it?

luka-nextcloud:
@mejo- @juliushaertl 
I have checked spreed/talk and found a component for placehoder/loading skeleton.
https://github.com/nextcloud/spreed/blob/master/src/components/LoadingPlaceholder.vue
I have made a new component that can be reusable (same as spreed).

mejo- (changes requested):
Thanks @luka-nextcloud, looks pretty nice already! Additionally to the code comments, some things I would improve:

* [x] In Collectives navigation list, always show the caption "Select a collective".
* [x] In Collectives navigation list, make sure that the placeholders for icon and text match exactly the places of actual content that is loaded afterwards. Currently, they jump a bit.
* [ ] In page header, the placeholder doesn't align exactly. Text placeholder bar is a bit too much to the left and subheading needs to move further down.
* [ ] In page content, the placeholder bars don't align exactly with the start of text. They need to be moved a tiny bit more to the right.

In order to test the pixel perfectness, the Vue browser extension is really helpful. It allows you to toggle the `loading` booleans in Vuex store on and off and compare the view with placeholder and loaded content easily. E.g. `loading.collectives` for the collectives navigation list, `loading.collective` for page list and page header, and `loading.pageContent` for the page content.

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,232 @@
...
+						:fill="'url(#placeholder-gradient' + suffix + ')'">
+						<circle class="page-heading-placeholder-icon" />
+						<rect class="page-heading-placeholder-line-one" :style="width" />
+						<rect class="page-heading-placeholder-line-two" />
+					</svg>
+					<svg v-if="type === 'text'"
+						class="text-placeholder"
+						xmlns="http://www.w3.org/2000/svg"
+						:fill="'url(#placeholder-gradient' + suffix + ')'">
+						<rect class="text-placeholder-line-one" />
```
mejo-, line 35: Is there a reason why you don't set a variable width in lines 1-3 for `type === 'text'`?

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,232 @@
+<template>
+	<div class="placeholder-main" :class="'placeholder-main-' + type">
+		<!-- Placeholder animation -->
+		<template v-for="(suffix, gradientIndex) in ['-regular', '-reverse']">
+			<svg :key="'gradient' + suffix" :class="'placeholder-gradient placeholder-gradient' + suffix">
```
mejo-, line 5: This whole `placeholder-gradient` `<svg>` object seems useless to me. Maybe a leftover` it has `position: fixed` and I don't see where it actually results in something being displayed. So maybe it can be removed altogether?

luka-nextcloud: It is the gradient animation which is needed.

mejo-: Makes sense, thanks for pointing it out.

src/components/SkeletonLoading.vue
```diff
@@ -0,0 +1,232 @@
...
+						:fill="'url(#placeholder-gradient' + suffix + ')'">
+						<circle class="page-heading-placeholder-icon" />
+						<rect class="page-heading-placeholder-line-one" :style="width" />
+						<rect class="page-heading-placeholder-line-two" />
+					</svg>
+					<svg v-if="type === 'text'"
+						class="text-placeholder"
+						xmlns="http://www.w3.org/2000/svg"
+						:fill="'url(#placeholder-gradient' + suffix + ')'">
+						<rect class="text-placeholder-line-one" style="width: 95%" />
```
mejo-, line 35: How about using real random computed values instead, like the `width` property of `placeholderData`?

src/components/PageList.vue
```diff
@@ -54,6 +53,7 @@
 				</NcActionButton>
 			</NcActions>
 		</div>
+		<SkeletonLoading v-if="loading('collective') || loading('pagelist')" type="items" :count="3" />
```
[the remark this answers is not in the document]
luka-nextcloud: I have checked, there is 1 function getting pages in background and we already pass false:

https://github.com/nextcloud/collectives/blob/7914cc6c34d61ae66f23e412c3d79aa38c948e47/src/components/Collective.vue#L188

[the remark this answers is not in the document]
mejo-: Exactly, and we need to make sure that `false` is passed to `GET_PAGES` (or respectively `dispatchGetPages()` in more places. I'll push a commit with some refactoring of the loading states in vuex store to address this.

mejo- (approved):

Issue #655, opened 2023-04-24
Error: Call to a member function getDisplayName() on null


crocodeev:
**Description**
After updating Nextcloud to 26 version with PHP to 8.1 version, Collectives throw error "**Could not fetch pages
Internal Server Error**" when user trying to get collective. 
Server logs provide next error "**Error: Call to a member function getDisplayName() on null**"

<details>
<summary>Server details</summary>

**Collectives app version: 2.4.0** 

**Operating system: Ubuntu 18.04 LTS**

**Web server: Apache/2.4.29**

**Database: MariaDB-1:10.3.16**

**PHP version: 8.1.18**

**Nextcloud version: 26.0.1**

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
 0. /var/www/html/nextcloud/apps/collectives/lib/Service/PageService.php line 493
    OCA\Collectives\Service\PageService->getPageByFile()
 1. /var/www/html/nextcloud/apps/collectives/lib/Service/PageService.php line 526
    OCA\Collectives\Service\PageService->recurseFolder()
 2. /var/www/html/nextcloud/apps/collectives/lib/Controller/PageController.php line 51
    OCA\Collectives\Service\PageService->findAll()
 3. /var/www/html/nextcloud/apps/collectives/lib/Controller/ErrorHelper.php line 25
    OCA\Collectives\Controller\PageController->OCA\Collectives\Controller\{closure}("*** sensitive parameters replaced ***")
 4. /var/www/html/nextcloud/apps/collectives/lib/Controller/PageController.php line 55
    OCA\Collectives\Controller\PageController->handleErrorResponse()
 5. /var/www/html/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 230
    OCA\Collectives\Controller\PageController->index()
 6. /var/www/html/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 137
    OC\AppFramework\Http\Dispatcher->executeController()
 7. /var/www/html/nextcloud/lib/private/AppFramework/App.php line 183
    OC\AppFramework\Http\Dispatcher->dispatch()
 8. /var/www/html/nextcloud/lib/private/Route/Router.php line 315
    OC\AppFramework\App::main()
 9. /var/www/html/nextcloud/lib/base.php line 1056
    OC\Route\Router->match()
10. /var/www/html/nextcloud/index.php line 36
    OC::handleRequest()
```

#### Browser log
```
[...]

PR #658, opened 2023-04-26
fix: Use getDisplayName to avoid calling on null

juliushaertl:
### 📝 Summary

* Resolves: #655

Avoid errors when the user cannot be found and additionally makes use of the display name cache which might save some queries

### 🏁 Checklist


juliushaertl:
Remaining cypress failures are known to also happen on main.

mejo- (approved):
Thanks a lot for looking into this @juliushaertl. Code changes look sensible to me.

mejo-:
Rebased against main.

PR #672, opened 2023-05-11
Refactor PHP `empty()` checks

mejo-:
See https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/ for why using `empty()` might be a bad idea.


lib/Service/PageService.php
```diff
@@ -942,7 +942,7 @@ public function matchBacklinks(PageInfo $pageInfo, string $content): bool {
 
 		$protocol = 'https?:\/\/';
 		$trustedDomainConfig = (array)$this->config->getSystemValue('trusted_domains', []);
-		$trustedDomains = !empty($trustedDomainConfig) ? '(' . implode('|', $trustedDomainConfig) . ')' : 'localhost';
+		$trustedDomains = !($trustedDomainConfig === []) ? '(' . implode('|', $trustedDomainConfig) . ')' : 'localhost';
```
max-nextcloud, line 945: You could avoid the `!` in the condition here - either by using the non-identity operator or by changing the order of the ternary operator.
```suggestion
		$trustedDomains = ($trustedDomainConfig !== []) ? '(' . implode('|', $trustedDomainConfig) . ')' : 'localhost';
```

max-nextcloud (approved):
Just a little bit of nitpicking.

juliushaertl (approved):

lib/Service/PageService.php
mejo-: Uh, nice catch :ok_hand:

Issue #47, opened 2020-08-02
Allow restoring a deleted page


mejo-:
After deleting a page we should allow restoring it.

marcotrevisan:
Yesterday a colleague deleted the whole contents of a collective (leaving only the root Collective existing, but empty), because of a misteke in using his webdav Files client.
I was luckily able to spot it by checking the web server logs, where I found the recursive DELETEs. Otherwise I would have said it "misteriously disappeared".
As you know Collectives are available via Webdav for client side editing too (no Collectives UI is protecting them).

For this reason in my humble opinion a trashbin is an expected feature. Mistakes in deletion can happen without the user even being aware of it.
Alternatively, an option to decide wether or not Collectives may show up in Files should do.... not sure how many people really use them in Files or via webdav client. Even the page in Files is suggesting the user not to use it.

Hope this helps in your decisions. 
Thanks and Regards

mejo-:
Thanks for your feedback @marinkaltani, much appreciated! We're currently working on a collective-wide trashbin for deleted pages.

PR #709, opened 2023-05-30
Add page trash frontend

mejo-:
### 📝 Summary

This PR depends on #700.

* Resolves: #47

#### 🖼️ Screenshots

Delete from page list | Restore from page trash | Mobile
---|---|---
2023-05-30T15:37:30,359120578+01:00 | 2023-05-30T15:38:19,318669963+01:00 | 2023-05-30T15:39:14,046148359+01:00

---

| Screencast |
|---|
| [recording2.webm](https://github.com/nextcloud/collectives/assets/3582805/53ca6f0b-7747-439e-b339-bd9f1604f176) |

### 🚧 TODO

- [ ] Feedback from design team
- [ ] Somehow highlight trash when a page gets deleted

### 🏁 Checklist


nimishavijay (changes requested):
Very cool! General idea looks great! :) Some points from my end:

- [ ] It would be useful to show the path to the pages as well, as I can see a use case where 2 pages with the same name have been deleted and then it's difficult to tell which one to restore. I also have concerns about the height of the list when there are more pages. How about a modal like this (adapted from the calendar trash bin)? There is a subline with the page "location" shown
image
- [x] Agreed that we can highlight the deleted pages when the page is deleted! Opening the trash is a bit much indeed
- [x] Similarly when a page it restored it can also be highlighted for a couple of seconds
- [x] if the page being restored is a subpage of a page that is collapsed in the list then it could automatically be expanded upon restoring

What do you think? :) also ccing @jancborchardt and @szaimen for more design feedback

jancborchardt:
Totally agree with @nimishavijay's proposal, modal is more proper here. :)

mejo-:
Thanks a lot for your feedback @nimishavijay and @jancborchardt. I implemented most of your suggestions. See the updated sceenshots/casts.

Only remaining is how to represent the path of the page in trashbin. In the files trashbin, original path is not displayed either. And it can become quite long and cluttered soon if we display it. That's why I went without first, but would be curious about your ideas.

jancborchardt (changes requested):
Nice! Some further feedback:
- [ ] It is a bit strange that "Delete page" is not the last entry of the action menu. The "last modified" info entry could be moved to first position instead?
- [ ] Pages with subpages also seems to remove all subpages. Maybe in that case it is ok to have a confirmation modal?
- [ ] The feedback with highlighting the "Deleted pages" is not super obvious. Generally we should always also use showNotification, in this case it could say "[Pagename] deleted", ideally even with an " Undo" button.
- [ ] In the modal:
   - [ ] The table headings don't need to be bold
   - [ ] "Restore" button could be outside of the menu so it's quick to access, in secondary button style.
   - [ ] The last horizontal line below the last list item is not necessary

mejo-:
Thanks @jancborchardt! Some quick follow-ups:

>  Pages with subpages also seems to remove all subpages. Maybe in that case it is ok to have a confirmation modal?

Well, it was forbidden before, but now that it's easy to restore I thought it would be ok to just allow it. But if you prefer, we can add a confirmation modal for sure.

>  It is a bit strange that "Delete page" is not the last entry of the action menu. The "last modified" info entry could be moved to first position instead?

I'm not sure whether I like that. I will give it a try, but in general the "last modified" info is really more something like an info and not something we should emphasize by putting it as the first entry, no?

jancborchardt:
> Thanks @jancborchardt! Some quick follow-ups:
> 
> > Pages with subpages also seems to remove all subpages. Maybe in that case it is ok to have a confirmation modal?
> 
> Well, it was forbidden before, but now that it's easy to restore I thought it would be ok to just allow it. But if you prefer, we can add a confirmation modal for sure.

Yeah, it's just that it's not super clear that this is what happens, at least before you do it. So yes I would say a confirmation is in order here.


> > It is a bit strange that "Delete page" is not the last entry of the action menu. The "last modified" info entry could be moved to first position instead?
> 
> I'm not sure whether I like that. I will give it a try, but in general the "last modified" info is really more something like an info and not something we should emphasize by putting it as the first entry, no?

Right, yeah and actually the "Convert with pandoc" is also a bit of a nerdy option (at least wording-wise)? What do you think about this order:

- Select emoji
- Move page
- Convert with pandoc
- Add template for subpages
- (edited by info entry)
- Delete page

mejo-:
> Right, yeah and actually the "Convert with pandoc" is also a bit of a nerdy option

Haha, yeah sorry this is from a custom extension (https://github.com/mejo-/pandoc) and not part of Collectives.

jancborchardt:
> Haha, yeah sorry this is from a custom extension (https://github.com/mejo-/pandoc) and not part of Collectives.

Nice! :) May I suggest "Convert or export" (or "Convert & export")?

jancborchardt (approved):
Very nice! I would still say the "modified" info line should move one up in the menu so delete is the last entry, but that can also be a follow-up. :)

mejo-:
> I would still say the "modified" info line should move one up in the menu so delete is the last entry, but that can also be a follow-up. :)

I would prefer to discuss this in one of the next design review calls. Because having an information box in between actionable menu items doesn't seem very logical to me. I see the point that "delete" should be the last itam in an action menu, but then the information box is not an "action" and would not fit in between other actions in my eyes.

src/components/PageList/Item.vue
```diff
@@ -449,4 +463,10 @@ export default {
 	right: 0;
 	margin: 0;
 }
+
+@keyframes highlight-animation {
```
juliushaertl, line 467: Maybe this could be moved out into a helper scss file as it is duplicate with src/components/PageList/PageTrash.vue

juliushaertl (commented):
Only a minor comment, non blocking, otherwise the code looks good to me. Haven't tested yet though.

juliushaertl:
Tested and found a few issues that might be worth to look into before merging:

- [ ] Page not removed from trashbin UI list after deleting it permanently
- [ ] Delete permanently breaks collective, steps to reproduce:
   - Create a collective
   - Add 1 subpage
   - Delete the sub page
   - Permanently delete the subpage
   - Reload the collective -> See Page not found: {page}
- [ ] When having a page open while deleting the main view should navigate to the parent page, otherwise you may end up with a broken text editor as the file is no longer present or can continue edit a deleted page
- [ ] Deleting a page does not immediately remove it from the list 
- [ ] Restoring a page failes if the page has been reloaded (might also happen if the previous would be the case):
  Screenshot 2023-06-20 at 21 13 24

mejo-:
Thanks for testing @juliushaertl. I was unable to reproduce most of the issues you face. The problem with broken collective was a severe bug though that I fixed in 3abce813f6e4842500b97048dcee1baa393f3ad5 now. Maybe you can test again?

>  Page not removed from trashbin UI list after deleting it permanently

For me, pages always get removed from the trashbin UI after deleting permanently

>  When having a page open while deleting the main view should navigate to the parent page, otherwise you may end up with a broken text editor as the file is no longer present or can continue edit a deleted page

That's what happens in my tests.

>  Deleting a page does not immediately remove it from the list
>  Restoring a page failes if the page has been reloaded (might also happen if the previous would be the case):

Mh, unable to reproduce these two as well.

juliushaertl:
Strange, now it works all fine. I'm wondering if I just have been on an outdated branch, though I'm quite sure I pulled before testing.

Anyways, works like a charm now ❤️

juliushaertl (approved):

Issue #727, opened 2023-06-15
missing slash at the end of the menu link


quenenni:
**Describe the bug**
The menu link to Collectives is the only one that does not finish with a /
Because of this, it gets a wrong address from the server (the port switches from 443 to 4431) and tries to enter the Collective page on the port 4431.

If I put the same address but with a / at the end, I enter the collectives page without problem.

I recently had to check my nginx configuratin to fix things with nextcloud v26, so I know I have the last version.

My setup is using Haproxy in front of Nginx and the port 4431 that the collective app gets to forge the address comes from somewhere there as Haproxy send requests to Nginx on port 4431.

I tried to force the port 443 in nginx configuration (with a `fastcgi_param SERVER_PORT 443;`), but no change.

Can't the solution be simply to add the slash at the end like all other menu points?

**To Reproduce**
1. Click on Collectives menu
2. Timeout error and we cn see in the addressbar that the port switched to 4431 (`https://cloud.xxx.yyy:4431/apps/collectives/`)

**Client details:**
Several browsers, several Os

<details>
<summary>Server details</summary>

**Collectives app version: v2.5.0** (see Nextcloud apps page)

**Operating system: Debian v11.7**

**Web server: Nginx v1.18.0-6.1+deb11u3**

**Database: Mariadb-server v1:10.5.19-0+deb11u2**

**PHP version: v8.1.8**

**Nextcloud version: v26.0.2** (see Nextcloud admin page)

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
nothing in nextcloud.log in debug mode.
nothing in nginx log.
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The network log
When I click on the menu, I can see a fist request to https://cloud...../apps/collective  (without the slash)
just after, a second line that does a GET request to https://cloud....:4431/apps/collectives/
Then nothing (except the notifications requests) until the error.

```

</details>

mejo-:
Thanks @quenenni for your report. Seems like the same problem reported in #411. I developed a fix at #733.

quenenni:
Hello,

Thanks a lot for your quick reaction.
I tested your patch and the problem is no more.

Much appreciated.

nunoleite:
Sorry to open this, but i think this created a new bug.
Now in search the link has 2 backslashes and it doesn't open
https://xxx.xxx.com/apps/collectives//something?fileId=32324

mejo-:
[...]

Issue #411, opened 2022-11-03
Collectives behind SSL Terminating reverse proxy not working (redirects https -> http)


mejo-:
In GitLab by @t0mcat1337 on Nov 3, 2022, 21:31

A Nextcloud (Version 24.0.6) is running behind an SSL terminating reverse proxy (Traefik), so internal traffic (Traefik -> Nextcloud) is running with HTTP. To the outside Nextcloud is just reachable via https through the mentioned reverse proxy at a subdomain (https://cloud.my.domain). 
Nextcloud and each and every installed app (News, Deck, Office, Contacts, Circles, Cookbook...) are working fine; when clicking at the menu icons the apps open. 
But when clicking on the menu icon of the freshly installed Collectives App, it redirects to http... and so is not reachable:

![grafik](https://gitlab.com/collectivecloud/collectives/uploads/4cae3a025456123bae31a510e7db271d/grafik.png)
(Notice the URL of the menu icon is correct https://....)
![grafik](https://gitlab.com/collectivecloud/collectives/uploads/6a6f991771747eda3889d9df668bfd33/grafik.png)
(Notice the 301 redirect to http://...)

When I manually correct the URL to https://... Collectives works normal.
![grafik](https://gitlab.com/collectivecloud/collectives/uploads/cbe651ad511e590d4e06ed48fb1e72cc/grafik.png)

While doing the screenshot I noticed, that the menu link goes to:

https://cloud.my.domain/apps/collectives

The redirect URL is:
http://cloud.my.domain/apps/collectives/

Notice the trailing slash.... could this have sth. to do with this behaviour? But why doesn't it correctly redirect to https at all?

BTW: For testing I even made a "overwriteprotocol=>'https'" entry in nextclouds config.php but this didn't help.

Any hint? 
Thx in advance

PR #733, opened 2023-06-19
Add extra default route without path parameter

mejo-:
This is required to have a trailing slash in our default route (`collectives.start.index`).

Fixes: #727
Fixes: #411

### 🏁 Checklist

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -99,14 +99,17 @@ return [
 		['name' => 'publicPageTrash#restore', 'url' => '/_api/p/{token}/_pages/trash/{id}', 'verb' => 'PATCH',
 			'requirements' => ['id' => '\d+']],
 
-		// default public route (Vue.js frontend)
+		// default Vue.js router route (Vue.js frontend)
+		['name' => 'start#index', 'url' => '/', 'verb' => 'GET'],
+
+		// Vue.js router public route (Vue.js frontend)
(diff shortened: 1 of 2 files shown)


appinfo/routes.php
```diff
@@ -99,14 +102,14 @@
...
 
-		// default public route (Vue.js frontend)
+		// Vue.js router public route (Vue.js frontend)
 		['name' => 'publicStart#publicIndex', 'url' => '/p/{token}/{path}', 'verb' => 'GET',
 			'requirements' => ['path' => '.*'],	'defaults' => ['path' => '']],
 
-		// default route (Vue.js frontend)
-		['name' => 'start#index', 'url' => '/{path}', 'verb' => 'GET',
+		// Vue.js router route (Vue.js frontend)
+		['name' => 'start#indexPath', 'url' => '/{path}', 'verb' => 'GET',
```
juliushaertl, line 110: Maybe move this up to the other route of StartController to have them all in one place?

juliushaertl (approved):
Minor comment, but looks good to me 👍

appinfo/routes.php
mejo-: Since these are basically catch-all routes, they need to come last. We could move the route without path param down to have all start controller routes together. Probably a good idea :+1:

juliushaertl: Ah I see, wasn't aware of that.

PR #740, opened 2023-06-26
Update Github workflow files from templates

mejo-:


mejo-:
CI failures are unrelated. I'll merge already as it improves the situation with CI a lot. But reviewers don't hesitate to comment and I'll address it in a follow-up.

PR #743, opened 2023-06-28
fix: Switch to ObjectStoreScanner if NoopScanner is not available

juliushaertl:
Adapt to the rename of the internal scanner class for object storage. I kept the NoopScanner to not break compatibility with old releases. 

This originates from https://github.com/nextcloud/server/pull/37691 so it would be good to include it in the next bugfix release for 27 as well.

### 🏁 Checklist


mejo- (approved):

mejo-:
Thanks for the quick fix @juliushaertl!

Issue #308, opened 2022-04-01
Membership management


mejo-:
Managing members of the collective should be possible from the Collectives app. Currently you have to navigate to the corresponding Circle in the Contacts app to do so.

mejo-:
@maximelehericy yes, good news: I'm currently working on that feature :blush: It will most likely be part of the next Collectives release.

Issue #244, opened 2021-09-17
Improve user management with a growing number of collectives


mejo-:
In GitLab by @larsbecker on Sep 17, 2021, 18:48

Since we begun evaluating collectives with more users it occurred to us, that the user management is quite uncomfortable with a growing number of collectives. This is because there's a hard relation between collectives and circles. I think this dependency doesn't make much sense and it would be much better to allow to configure the circles (and possibly groups) which have access to the collective, so that there is effectively a n:m relationship. 

This would ease user management and help to reduce the circles and the need to enter many users again and again.

mejo-:
@larsbecker we started to implement user management inside the Collectives app. For now, you will be able to select members in the dialog for a new collective, but we plan to have a dialog for adding and removing members from existing collectives later on.

Given that this hides the complexity of "collective <-> circle" from users, would this solve this issue for you? Admins could just add circles to collectives without having to care about the fact that the collective always has an "owner circle" itself.

lars-becker:
> Given that this hides the complexity of "collective <-> circle" from users, would this solve this issue for you? Admins could just add circles to collectives without having to care about the fact that the collective always has an "owner circle" itself.

That certainly would help with user management. I still don't like the fact, that there needs to be an owning circle because this seems somewhat redundant in some use cases (such as having three different collectives which share the same user group, e.g. an ngo board, that needs to access it), but you surely could argue that this is a subjective aesthetic preference for not populating the circle index with unnecessary circles.  I guess in praxi this won't hurt. Thanks!

PR #804, opened 2023-07-24
Manage members in app

### LINKED ISSUE #779: Remove the notification "The contacts app is required to manage members"

I provide access to the App Contacts through a group because it's an unnecessary module for most users. Only admins of the "Collective" have access to it. As a result, all users who don't have access to the App Contacts constantly encounter the error message "The contacts app is required to manage members," which is disruptive.

Screenshot 2023-07-11 095305

Remove the existing notification. Add the text "You don't have access to the Contacts App. Please contact the Administrator" to the Collective card.

Screenshot 2023-07-11 100551

**Additional context**

If the proposed solution is suitable, I can implement it.

mejo-:
### 📝 Summary

This PR adds functionality to manage members for a collective inside the Collectives app. This removes our dependency on the Contacts app to do so.

It's possible to add and remove members and promote/demote their membership level.

* Resolves: #308
* Resolves: #244

#### 🖼️ Screenshots

Searching for members | Changing membership level
---|---
grafik | grafik

#### 🖼️ Screencast

| 🏡 Walkthrough creating a collective and managing its members |
|---|
| [recording1.webm](https://github.com/nextcloud/collectives/assets/3582805/4ef9d177-205e-4937-98d0-c272afd35e49) |

### 🚧 TODO

- [x] Cypress tests
- [x] Don't allow to delete circle/group if current admin is member through it
- [x] Always show option to leave collective (also to admins)
- [x] Remove link to manage members in contacts
- [x] Remove dependency checks on contacts app (and close #779)
- [x] Filtering out member groups doesn't work (groups have type 16 in circle response)

### 🏁 Checklist


mejo-:
@nextcloud/designers I'd appreciate feedback from you :blush: In particular on the icons used for admin/moderator/member levels and on the way the loading indicator icon is used.

marcoambrosini (approved):
From a design perspective, I'm sorry but I don't have any constructive feedback to share. It looks very neat and easy to use, congrats!! :)

cypress/e2e/collective-members.spec.js
```diff
@@ -0,0 +1,109 @@
...
+			.find('.action-item__menutoggle')
+			.click({ force: true })
+		cy.get('button.action-button')
+			.contains('Manage members')
+			.click()
+	})
+
+	describe('Manage members', function() {
+		it('Allows to add members', function() {
+			// const addMembers = ['alice', 'Bobs Group']
```
juliushaertl, line 48: Any reason to leave this around? Shall we have a separate test for the groups?

juliushaertl (commented):
Looks very nice, code also looks good, just a few commments from testing:

- [x] The search results are not keyboard accessible
- [ ] The scroll in the modal seems a bit odd with two scroll areas:
   
Screenshot 2023-07-25 at 09 30 04

- [ ] Maybe we can hide the "Add users, groups  or circles ..." heading if the search is not triggered, the placeholder hint to search to add groups seems enough and the members also don't have a heading in the default state

Screenshot 2023-07-25 at 09 32 39

cypress/e2e/collective-members.spec.js
mejo-: Yes, I want to switch back to test with a group as well as soon as I resolved the following problem: for now, groups that are part of members already are not filtered out of the search results. The problem is that groups have type `16` (`TYPE_GROUP`) in circle response.

mejo-:
>     Maybe we can hide the "Add users, groups or circles ..." heading if the search is not triggered, the placeholder hint to search to add groups seems enough and the members also don't have a heading in the default state

@juliushaertl I would prefer to keep it. Talk behaves the same in the participants tab and I think its good to emphasize that you can search for new members to add.

juliushaertl (approved):

max-nextcloud:
Looking at the screencast it seems that people, circles and groups that are already in the collective are still listed in the sections to add members. I tried in talk and it seems to filter them out. I think that would be good.

mejo-:
> Looking at the screencast it seems that people, circles and groups that are already in the collective are still listed in the sections to add members. I tried in talk and it seems to filter them out. I think that would be good.

This is already fixed in the PR, but I didn't update the screencast :relaxed:

nimishavijay:
Looks super nice! Only some minor points :)

- We generally don't use the "light" font weight very much, instead we use the regular font weight in `--color-text-maxcontrast`
- We could switch the icons for admin and moderator since the admin role has the most permissions
- Since there are 2 levels of permissions, there could be an simple explanation in a subline for all the roles to avoid confusion between admin and moderators
  - Admin: Manage collective members, view, edit, share and delete collectives
  - Moderators: View collectives, edit and share collectives if permitted
  - Members: View collectives, edit and share collectives if permitted

mejo-:
Thanks @nimishavijay. I already changed the font color to `--color-text-maxcontrast` and switched the icons. For the level explanations this is a bigger thing and I'd like to postpone it, as I either would have to reimplement or extend NcActionButton.

Issue #808, opened 2023-07-27
Navigation pills are not clickable inside the upper or lower paddings


arnowelzel:
**Describe the bug**
On the left side there are "navigation pills" when hovering a link to a document. However not the whole pill is clickable, even then the shape of the mous pointer gives the impression that you can click. You have to move the mouse pointer exactly over the document title. Clicking slightly above or below it does not work.

**To Reproduce**
1. Create a collective with some documents.
2. Hover document in the tree but move the mouse cursor right or above or below the the document title with the navigation pill still focused.
3. Try to click the pill to open the document - nothing happens.

**Expected behavior**
The document will open, regardless where inside the navigation pill you click, except when clicking "..." or "+". See the Nextcloud files app for reference.

**Screenshots**
![image](https://github.com/nextcloud/collectives/assets/6613614/39225b69-5156-4765-9a08-c63527673acc)

**Client details:**
 - OS: Windows 10
 - Browser: Vivaldi, Firefox, Chrome
 - Version: current Versions (e.g. Firefox 115.0.2), but the version does not matter
 - Device: Desktop

<details>
<summary>Server details</summary>

**Collectives app version:** 2.6.1

**Operating system:** Ubuntu 22.04

**Web server:** Apache 2.4

**Database:** MariaDB

**PHP version:** PHP 8.1

**Nextcloud version:** 27.0.1

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

arnowelzel:
Using [CustomCSS](https://apps.nextcloud.com/apps/theming_customcss) I created a workaround using CSS - enlarge the clickable area with an additional padding:

```
.app-collectives .app-content-list-item-link {
  padding-top: 10px !important;
  padding-bottom: 10px !important;
  margin-top: -10px !important;
  margin-bottom: -10px !important;
}
```

mejo-:
Thanks a lot for the report @arnowelzel. I just pushed a fix.

PR #810, opened 2023-07-28
fix(pagelist): Grow link area in page list items to full height

mejo-:
Fixes: #808


max-nextcloud:
Looks good. Should we maybe also upstream this in nextcloud-vue or is it a collectives specific issue?

max-nextcloud (approved):

Issue #311, opened 2022-04-01
Improved landing page


mejo-:
The idea is to display more information about the collective on the landing page and turn it into a dashboard step by step.

* [ ] List five most recently changed pages
* [ ] List collective members avatars, sorted by most recent activity
* [ ] List of personal favourite pages (once #300 got implemented)

The editable page would be displayed below these information.

szaimen:
As for the favorites, the same layout could be used like the one of recent pages :)

mejo-:
Notes from discussing the mockup with @szaimen:

* Only one line (scrollable horizontally) with avatars for collective members. Tooltip shows display name, click optionally opens member modal.
* Last edited user and time in the recent pages tiles. Recent pages are scrollable horizontally as well.
* Editable page content moves blow recent pages and members and keeps full height.

szaimen:
Hey, so I tried to incoroprate the changes (and also some that I discussed with Jan). This is my new mockup. Let me know what you think:
![image](https://github.com/nextcloud/collectives/assets/42591237/8152a774-eafe-42e4-8f9c-ea358eff0611)

Unfortunately I will be away starting form monday so if you need additional changes, probably someone else from the Design team needs to take over.

The penpot file is here: https://design.penpot.app/#/workspace/09f8f1e0-a390-11eb-bb49-81aaa3a367f9/c65f3168-697c-808c-8002-dd48bd82cca0?page-id=583600af-26f0-8056-8002-dd3a532f090e&layout=layers

max-nextcloud:
Looks good to me. I just have some minor comments on the page list:
* The page emojis look somewhat lost with all the whitespace around. I would double their size to reduce the whitespace to the text below.
* The user bubbles should not be primary color
* It would be good to add a timestamp so one can tell wether these pages were changed yesterday or a long time ago.

nimishavijay:
Super nice mockup and agreed with the comments! Some more minor feedback from my side:
- "Collective Members" --> "Collective members" (note the lowercase "m")
- Instead of a timestamp separately, how about changing the wording to "Edited 10 minutes ago by Alex Doe" and "Edited last week by Sam Smith", etc? :)

jancborchardt:
> Instead of a timestamp separately, how about changing the wording to "Edited 10 minutes ago by Alex Doe" and "Edited last week by Sam Smith", etc? :)

[...]

PR #812, opened 2023-07-31
List recent pages and members on landing page

mejo-:
### 📝 Summary

* Show 10 most recent pages on landing page.
* Show 3 last active members on landing page if it's not a public share.
* Resolves: #311

#### 🖼️ Screenshots

| Overview |
|---|
| grafik |

| Screencast (slightly outdated) |
|---|
| [recording1.webm](https://github.com/nextcloud/collectives/assets/3582805/7da61f0e-0954-4300-99fe-e9a0447f8dea) |

### 🚧 TODO

- [x] Animate scroll to left/right in recent pages slider
- [x] Update scroll buttons on scrolling (without using the buttons)
- [x] Don't display scroll to right button initially when not necessary
- [ ] Show status in members avatars
- [x] Cypress tests

### 🏁 Checklist

### DIFF
--- a/src/components/Member/CurrentMembers.vue
+++ b/src/components/Member/CurrentMembers.vue
@@ -5,6 +5,7 @@
 		<Member v-for="item in searchedMembers"
 			:key="item.singleId"
 			:circle-id="circleId"
+			:current-user-is-admin="currentUserIsAdmin"
 			:member-id="item.id"
 			:user-id="item.userId"
 			:display-name="item.displayName"
@@ -45,6 +46,10 @@ export default {
 		searchQuery: {
 			type: String,
 			default: '',
+		},
+		currentUserIsAdmin: {
+			type: Boolean,
+			default: true,
 		},
 	},
 
--- a/src/components/Member/CurrentMembers.vue
+++ b/src/components/Member/CurrentMembers.vue
@@ -60,7 +60,7 @@ export default {
 		sortedMembers() {
 			return this.currentMembers
 				.slice()
-				.sort(this.sortMembers)
+				.sort(this.sortCurrentUserFirst)
 		},
 
 		searchedMembers() {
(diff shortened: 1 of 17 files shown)


jancborchardt (commented):
Very nice! Would only say 2 things:

> Show 3 last active members on landing page if it's not a public share.

Why not all until ellipsis is needed? It is nice to see faces. :)

And also a bit more vertical whitespace between the sections would be cool.

src/components/Member/CurrentMembers.vue
```diff
@@ -86,37 +91,17 @@ export default {
...
-		 * @param {number} m2.level Second member level
-		 * @param {number} m2.userType Second member user type
 		 */
-		sortMembers(m1, m2) {
+		sortCurrentUserFirst(m1, m2) {
 			if (m1.userId === this.currentUser) {
 				return -1
 			} else if (m2.userId === this.currentUser) {
 				return 1
 			}
```
max-nextcloud, line 104: What if neither of the two is the currentUser?
Right now this will return nothing / undefined, right?
Is that what you want?

max-nextcloud (approved):
Had a brief look at the code. So far everything seems fine. Just have one question.

src/components/Member/CurrentMembers.vue
mejo-: Well, strictly speaking we could also `return 0`. But to my knowledge that's the default - not to sort the two. Given that I moved sorting by user type and level into `src/store/circles.js`, this sort functions sole left purpose is to put `currentUser` first.

max-nextcloud: Ahh... now i get it. `currentMembers` is already passed in presorted.
Let's still return 0 to follow the requirements for the sort function as described here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters

> The return value should be a number whose sign indicates the relative order of the two elements: negative if a is less than b, positive if a is greater than b, and zero if they are equal. NaN is treated as 0.

max-nextcloud: Oh... and I know this is accademic and probably has no real effect :monocle_face: 
Another requirement for the function to return stable results is for it to be reflexive:
```js
compareFn(a, a) === 0.
```

mejo-: Agreed, thanks for being thorough here :heart: Fixed now.

juliushaertl:
Looks really great already 😍 

Just a minor UI nitpick:

- The contrast of icons for pages without emojis seems a bit too low
- I think the relative timestamp might need more space especially for languages that are a bit more verbose or if the name is longer

Screenshot 2023-08-01 at 16 44 44

mejo-:
> I think the relative timestamp might need more space especially for languages that are a bit more verbose or if the name is longer

Unfortunately I don't have a good idea here. I'd rather avoid to make the tiles even wider, which would result in less tiles being displayed on narrow screens. And I configured it to be ellipsised on purpose to prevent randomly growing tile width. One option would be to add a line break, like "admin\nseconds ago". What do designers think @jancborchardt @nimishavijay?

max-nextcloud:
Ideas that come to my mind:
* as you said... use two rows: one for the user bubble one for the timestamp. Maybe even move the page title into the tile above?
* Use the avatar only instead of the user bubble. Maybe with a tooltip with the user display name.

jancborchardt:
@max-nextcloud yep, avatar only seems fine. :)

nimishavijay:
@mejo- @max-nextcloud I would actually vote for the line break, for the following reasons:
- since the avatar is so small it will most likely not be understood easily
- we generally don't use the avatar in such a small size in other places
- even if there is hover feedback, the touch target is only around 16-20px so possibly not accessible either. Especially on mobile I can imagine trying to click on the avatar and accidentally opening the page.

We could instead show the full user bubble on one line and the timestamp on the next line. That way the entire item could be clickable without any issues :)

juliushaertl (approved):

jancborchardt (commented):
The headings (main page heading and also "Recent pages", "Collective members" and "Landing page") being grey instead of color-main-text looks a bit off, otherwise all good. :)

szaimen:
🎉🎉🎉

PR #817, opened 2023-08-06
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


mejo-:
The app upgrade failure is because the Collectives version in the app store is no longer compatible with Nextcloud 28. It will resolve itself as soon as we release a new version.

mejo- (approved):

PR #819, opened 2023-08-07
fix(webpack): set `__webpack_public_path__`

mejo-:
Required for loading webpack chunks if the app is in a non-default apps folder (like `apps-extra/`).

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -25,9 +25,12 @@ import Collectives from './Collectives.vue'
 import router from './router.js'
 import store from './store/store.js'
 import { sync } from 'vuex-router-sync'
+import { generateFilePath } from '@nextcloud/router'
 
 /** Global directives */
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -29,6 +29,8 @@ import { sync } from 'vuex-router-sync'
 /** Global directives */
 import VTooltip from '@nextcloud/vue/dist/Directives/Tooltip.js'
 
+__webpack_public_path__ = OC.linkTo('collectives', 'js/') // eslint-disable-line
```
juliushaertl, line 32: Sorry I copied the outdated one, we should rather use the npm package API:

```js
import { generateFilePath } from '@nextcloud/router'

__webpack_public_path__ = generateFilePath('collectives', '', 'js/')
```

mejo-: Oh, thanks for the pointer. Fixed now!

juliushaertl (approved):

PR #822, opened 2023-08-09
fix: Make sure collectives own scroll area is used for menubar stickyness

juliushaertl:
As https://github.com/nextcloud/text/pull/2832 introduced a max-height to properly handle `.text-editor` being the scroll container for any viewer rendered editor, we will need to adapt the styling here slightly as collectives comes with its own scroll container.

Note that the API for rendering the editor externally would solve that as this one doesn't bring its own scroll container https://github.com/nextcloud/text#%EF%B8%8F-integrate-text-in-your-app but needs more work to have feature parity for collectives.

- [x] Test with all compatible nextcloud releases
  - [x] 27.1
  - [x] 27.0
  - [x] 26
  - [x] 25
### 🏁 Checklist

### DIFF
--- a/src/Collectives.vue
+++ b/src/Collectives.vue
@@ -131,6 +131,10 @@ export default {
 	padding-bottom: 200px;
 }
 
+#titleform {
+	z-index: 10022;
+}
+
 #version-title, #titleform input[type='text'] {
 	font-size: 35px;
 	border: none;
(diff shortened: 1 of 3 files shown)


juliushaertl:
I think the remaining cypress failures when loading a page could stabilize with https://github.com/nextcloud/text/pull/4657

mejo-:
Nice, thanks a lot @juliushaertl! Tested with Nextcloud 25, 26, 27 and master and all work well. Pushed a comment to fix stickyness of the titlebar when viewing old versions.

mejo- (approved):

Issue #828, opened 2023-08-11
IEventDispatcher issue on 27


juliushaertl:
Using the latest release on my 27 instance caused the following log entry:

We cannot just pass the IEventDispatcher here for Nextcloud <28, so we may need a fallback to pass the SymfonyDispatcher on older releases.

I can check this next week.

```
{
  "reqId": "YiBHLTQHe1Pao8XpFaFQ",
  "level": 3,
  "time": "2023-08-11T14:15:39+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "core",
  "method": "",
  "url": "--",
  "message": "Error while running background job (class: OCA\\Collectives\\BackgroundJob\\ExpirePageVersions, arguments: )",
  "userAgent": "--",
  "version": "27.1.0.0",
  "exception": {
    "Exception": "TypeError",
    "Message": "OC\\User\\User::__construct(): Argument #3 ($dispatcher) must be of type Symfony\\Component\\EventDispatcher\\EventDispatcherInterface, OC\\EventDispatcher\\EventDispatcher given, called in /var/www/nextcloud/apps-extra/collectives/lib/Versions/CollectiveVersionsExpireManager.php on line 121",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/nextcloud/apps-extra/collectives/lib/Versions/CollectiveVersionsExpireManager.php",
        "line": 121,
        "function": "__construct",
        "class": "OC\\User\\User",
        "type": "->"
      },
      {
        "file": "/var/www/nextcloud/apps-extra/collectives/lib/Versions/CollectiveVersionsExpireManager.php",
        "line": 96,
        "function": "expireFolder",
        "class": "OCA\\Collectives\\Versions\\CollectiveVersionsExpireManager",
        "type": "->"
      },
      {
        "file": "/var/www/nextcloud/apps-extra/collectives/lib/BackgroundJob/ExpirePageVersions.php",
        "line": 35,
        "function": "expireAll",
        "class": "OCA\\Collectives\\Versions\\CollectiveVersionsExpireManager",
        "type": "->"
      },
      {
        "file": "/var/www/nextcloud/lib/public/BackgroundJob/Job.php",
        "line": 81,
        "function": "run",
        "class": "OCA\\Collectives\\BackgroundJob\\ExpirePageVersions",
        "type": "->"
      },
      {
        "file": "/var/www/nextcloud/lib/public/BackgroundJob/TimedJob.php",
        "line": 103,
[...]

PR #833, opened 2023-08-14
fix: Fallback to old eventdispatcher on older releases

juliushaertl:
Fix https://github.com/nextcloud/collectives/issues/828

### 🏁 Checklist

### DIFF
--- a/.github/workflows/phpunit-sqlite.yml
+++ b/.github/workflows/phpunit-sqlite.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       matrix:
         php-versions: ['8.0']
-        server-versions: ['master']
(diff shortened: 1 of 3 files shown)


juliushaertl:
Let me just push a unit test to assert the relevant bits, a full test for the background job is more tricky

juliushaertl:
Pushed, though the test is not very fancy

max-nextcloud (approved):
Thanks for the test! :tada:

theroch (approved):
I've have running this patch for several days now without any issues.

level420:
manually applied the PR and it works as exptected.
please release a new version of circes compatible with nextcloud 27 (and 26) with this PR!

PR #834, opened 2023-08-14
fix: Use shared init file to properly set nonce for async component loading

juliushaertl:
Otherwise mermaid diagrams are not loaded in read only view as the async chunks cannot be fetched due to CSP errors

### 🏁 Checklist


max-nextcloud (approved):
:elephant: trust you on this one.

There's a lint error still though.

PR #851, opened 2023-08-28
fix(cy): use flexible selector for file list in Nextcloud 25-28

max-nextcloud:
### 📝 Summary

* Resolves: https://github.com/nextcloud/collectives/actions/runs/5981758520/job/16230201814

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
The apps -- Disabled circles app does not break files view -- Renders the default files list (failed)
 | A

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/apps.spec.js
+++ b/cypress/e2e/apps.spec.js
@@ -51,7 +51,7 @@ describe('The apps', function() {
 
 		it('Renders the default files list', function() {
 			cy.login('jane', { route: 'apps/files' })
-			cy.get('.files-fileList tr').should('contain', 'welcome.txt')
+			cy.get('.files-fileList tr, .files-list__tbody tr').should('contain', 'welcome.txt')
 		})
 
 	})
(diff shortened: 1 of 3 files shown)


cypress/e2e/apps.spec.js
```diff
@@ -51,7 +51,7 @@ describe('The apps', function() {
 
 		it('Renders the default files list', function() {
 			cy.login('jane', { route: 'apps/files' })
-			cy.get('.files-fileList tr').should('contain', 'welcome.txt')
+			cy.get('.files-fileList tr, files-list__tbody tr').should('contain', 'welcome.txt')
```
juliushaertl, line 54: Shouldn't it be a class selector? 

```suggestion
			cy.get('.files-fileList tr, .files-list__tbody tr').should('contain', 'welcome.txt')
```

max-nextcloud: Good point. I wonder why the test in question passed never the less but failed on last run in main.

mejo- (approved):

PR #864, opened 2023-09-06
fix(CollectiveFolderManager): Use our own LazyFolder implementation

mejo-:
Required because constructor parameters changed for the private LazyFolder class with Nextcloud 28.

### DIFF
--- a/lib/Trash/PageTrashBackend.php
+++ b/lib/Trash/PageTrashBackend.php
@@ -539,7 +539,8 @@ class PageTrashBackend implements ITrashBackend {
 			$trashNode = $this->getTrashNodeById($user, $fileId);
 			// Get parent folder for index pages
 			if ($trashNode instanceof File && NodeHelper::isIndexPage($trashNode)) {
-				$trashNode = $trashNode->getParent();
+				// The extra `get()` is required to resolve the lazy folder from getParent()
+				$trashNode = $trashNode->getParent()->get('');
 			}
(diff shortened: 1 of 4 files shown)


max-nextcloud:
Thanks a lot for taking care of this.

lib/Trash/PageTrashBackend.php
```diff
@@ -539,7 +539,8 @@ public function getTrashItemByCollectiveAndId(IUser $user, int $collectiveId, in
 			$trashNode = $this->getTrashNodeById($user, $fileId);
 			// Get parent folder for index pages
 			if ($trashNode instanceof File && NodeHelper::isIndexPage($trashNode)) {
-				$trashNode = $trashNode->getParent();
+				// The extra `get()` is required to resolve the lazy folder from getParent()
+				$trashNode = $trashNode->getParent()->get('');
```
mejo-, line 543: While this seems to work, I'm a bit unsure whether it's the best approach. Problem i that `getParent()` returns a `LazyFolder` and `method_exists($trashNode, 'getFileInfo')` three lines below returns `false`. Adding `getFileInfo()` to our own `LazyFolder` implementation (see above) doesn't change this - probably because the `LazyFolder` returned by `getParent()` is the implementation from server.

@juliushaertl and @blizzz maybe you have ideas/opinions?

juliushaertl (approved):

Issue #242, opened 2021-09-08
Width of the Collectives app (white space between page content and page list)


mejo-:
On large screens, there's a lot of unusued white space between page content and page list.

We got the feedback a few times that this looks "wrong" and ugly.

KaeTuuN:
Today, on a modern PC o Laptop with at least a 16:9 screen, this App is rather useless as a Wiki. Looking at a table that is larger than 2x2 is just a pain in the ***.

My suggestions are:
1. Increase the standard width at least by 25%!
2. Implement a Button to switch between "Full Width" an "Standard width" (like @theron29 wrote)
3. Think about a dynamic width based on the available Space (maybe width a min and max value)
All of these points could be developed and implemented separatly but I think this will increase the usability of Collectives a lot!

Greetings,
Kae

PR #874, opened 2023-09-12
feat(page): Add toggle for full width page view

mejo-:
The full width mode is a per-page user setting and is persisted in browser storage. It will not persist across devices.

Fixes: #242

### Screenshots

| Nomal | Full width |
| --- | --- |
| grafik | grafik |

### 🏁 Checklist


nimishavijay (changes requested):
Super nice! Notes from the design review:

- [x] Move "Last changed by" item to the top
- [x] Add divider before and after "Full width" and "Show outline"
```
Jane Doe an hour ago
---
Full width
Show outline 
---
Add template for subpages
Show in Files
Delete page
```

- [ ] Bug: if full-width is toggled, the action menu does not close, but 3 dot menu moves --> to be fixed in component
  - [x] Temporary solution: Always show the group of 3 buttons (Edit, 3dot menu, sidebar toggle) at the right end of the page 
  - [x] Group of buttons should have equal top and right margin
  - [ ] Nice-to-have: buttons are aligned with top right header buttons (avatar, notifications, contact search, universal search)

mejo-:
@nimishavijay I addressed all the topics we discussed in the call, please see the updated screenshots :relaxed:

nimishavijay (approved):
Super! :)

jancborchardt:
@KaeTuuN @theron29 since you commented in the issue this fixes, would this "per-page per-device" setting work for you?

@mejo- I’m leaning a bit towards that the layout should be a per-device setting across pages, since per the issue it seems it’s mostly people with large screens, and that doesn’t change across pages.

juliushaertl:
> @mejo- I’m leaning a bit towards that the layout should be a per-device setting across pages, since per the issue it seems it’s mostly people with large screens, and that doesn’t change across pages.

From my perspective it is more the actual page content that would influence if you want to use the full width or regular width mode, e.g. for pages with long text you might prefer a limited width for readability while for pages with larger tables or other content a full width is more suitable. I think mid term we should also adapt this in text itself, so having it per document could also be stored in the file rather than per user.

cypress/e2e/pages.spec.js
```diff
@@ -269,6 +269,36 @@ describe('Page', function() {
 		})
 	})
 
+	describe.only('Full width view', function() {
```
max-nextcloud, line 272:
```suggestion
	describe('Full width view', function() {
```

max-nextcloud (changes requested):
Had a quick look at the code. Looks good all in all.

Just one minor thing that needs to be addressed before the merge.

theron29:
@jancborchardt Yes, this actually looks very good and promising :+1:

cypress/e2e/pages.spec.js
mejo-: Good catch, fixed now :relaxed:

mejo-:
Failing Cyress tests are related, will look into it next week.

marcoambrosini (approved):

KaeTuuN:
@jancborchardt Thanks for mentioning me!

First: Thanks for developing! (@mejo- )
Second:

1. Will this Setting also affect the Edit Mode? (I hope so :-) )
2. Won't it be possible to move the Option directly to the top? Reason: Most complains were about table views and now you have to do more clicks for switching if you just want to view a table on full width and then go back to "normal" mode.

Greetings,
Kae

KaeTuuN:
I'm on my Smartphone atm.
I just want to put it as an Icon next to the edit Button.
Expand Button might look like this: https://cdn4.iconfinder.com/data/icons/ui-controls-editor/32/interface-controls-horizontal-expand-512.png
And the shrink Button might look like this: https://cdn.onlinewebfonts.com/svg/img_278259.png
The Icon changes in case of switching width.

Greetings,
Kae

jancborchardt (approved):
Thanks for the additional explanation @juliushaertl, then looks good design-wise. :)

And the setting is good in the menu, moving it out would make it too present compared to how important it is.

mejo-:
>     1. Will this Setting also affect the Edit Mode? (I hope so :-) )

Sure :blush: 

>     2. Won't it be possible to move the Option directly to the top? Reason: Most complains were about table views and now you have to do more clicks for switching if you just want to view a table on full width and then go back to "normal" mode.

Well, as @jancborchardt already said, this would be a bit too prominent. We try to not clutter the user interface with too many buttons and switches to not overwhelm users.

src/store/pages.js
```diff
@@ -763,5 +778,37 @@ export default {
...
+		 * @param {object} store.getters getters of the store
+		 * @param {boolean} fullWidthView Whether full width view is enabled or not
+		 */
+		[SET_FULL_WIDTH_VIEW]({ commit, getters }, fullWidthView) {
+			const pageId = getters.currentPage.id
+			const fullWidthPageIds = JSON.parse(persistentStorage.getItem('fullWidthPageIds') ?? '[]')
+			if (fullWidthView && !fullWidthPageIds.includes(pageId)) {
+				fullWidthPageIds.push(pageId)
+				commit(SET_FULL_WIDTH_PAGEIDS, fullWidthPageIds)
+				persistentStorage.setItem('fullWidthPageIds', JSON.stringify(fullWidthPageIds))
```
juliushaertl, line 806: Would be nice to prefix this with the app id, considering we may want to share this state with text at some point maybe we can already use something like `text-fullWidthPageIds`?

PR #878, opened 2023-09-13
test(ci): upload all the nextcloud logs in cypress tests

max-nextcloud:
Use different names for the artifacts so they all become available.

Also rename the screenshot artifacts
to clearify which version they are for.


max-nextcloud:
I confirmed in the uploaded attachments that this does the trick... So moving forward with this to ease debugging other issues.

PR #880, opened 2023-09-13
test(ci): always upload nextcloud log for debugging

max-nextcloud:
Right now i need this to see if exceptions in failed runs also occure in successful runs.


mejo-:
Mh, is this something you want to merge into `main` or merely a debugging PR? :thinking:

max-nextcloud:
I think uploading the logs and config might be generally useful. Not sure about the circles test. Maxence says it might actually block the run. But we're investigating circle errors in the logs.

mejo- (approved):

PR #883, opened 2023-09-14
test(cypress): Make page list trash test retriable

max-nextcloud:
Only insert the attachment once.

When opening the page for the first time
it is in edit mode and the attachment can be inserted.

On test retries the attachment was already there
and the page opened in view mode.
This led to test failures when attempting to switch to view mode.

### DIFF
--- a/cypress/e2e/page-list.spec.js
+++ b/cypress/e2e/page-list.spec.js
@@ -148,15 +148,18 @@ describe('Page list', function() {
 	})
 
 	describe('Page trash', function() {
-		it('allows to trash and restore page with subpage and attachment', function() {
-			cy.visit('/apps/collectives/Our%20Garden/Day%201')
 
-			// Insert attachment
+		// Insert attachment once
+		before(function() {
+			cy.visit('/apps/collectives/Our%20Garden/Day%201')
 			cy.intercept({ method: 'POST', url: '**/text/attachment/upload*' }).as('attachmentUpload')
 			cy.get('input[data-text-el="attachment-file-input"]')
 				.selectFile('cypress/fixtures/test.png', { force: true })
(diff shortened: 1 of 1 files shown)


max-nextcloud:
I'm not sure this will actually make the test more robust. If the test already fails in the new `before` block it will not be retried either. But then at least the actual failure will be logged rather than the failed retry attempts.

cypress/support/commands.js
```diff
@@ -360,3 +360,11 @@ Cypress.Commands.add('seedCircleMember', (name, userId, type = 1, level) => {
 			}
 		})
 })
+
+/**
+ * Fail the test on the initial run to check if retries work
+ */
+Cypress.Commands.add('testRetry', () => {
```
mejo-, line 367: Is this actually used anywhere? Or do you want to add it for using it later on?

mejo- (approved):
Looks good to me, only one comment on the added Cypress command that doesn't seem to get used anywhere.

cypress/support/commands.js
max-nextcloud: The latter. you can put `cy.testRetry()` at the end of a test and it will fail on the first two runs.

This way you can check if retries can actually pass or will always fail. I thought this might be useful to keep around. but i can als always add it back if i need it.

mejo-: Ah, great. I'm fine with keeping it :blush:

juliushaertl: Nice one

mejo- (changes requested):
Apparently the page list trash test reliably fails now. So seems like it got worse :laughing:

max-nextcloud:
I can just open a new one based on the current cypress tests.

Issue #860, opened 2023-09-05
Port filelist collective wrapper to current server dev version


max-nextcloud:
**Is your feature request related to a problem? Please describe.**
In the current server development version there is no message inside the collective folders because the since the files 2 vue rewrite the API for extending the files ui changed.
You can see this in the failing cypress tests on the server master branch.

**Describe the solution you'd like**
Register the message using the new hooks.

**Describe alternatives you've considered**
skip the cypress tests - but that's not easy and not what we want in the end.

**Additional context**
text already uses the new hook system for registering the rich workspaces:
https://github.com/nextcloud/text/pull/4666/files#diff-6653fe3a8ac35dad2e28ef2134d921749c4334d4b81a593f80ae312cde95df4e

PR #895, opened 2023-09-21
Bring back file list header in Nextcloud 28

max-nextcloud:
### 📝 Summary

* Resolves: #860

#### 🖼️ Screenshots

grafik

### :heavy_check_mark: TODO

- [x] figure out if we can continue to use the `import Vue` as we did in the past
- [x] determine if we want to load the `init.js` only on 28.

### 🏁 Checklist


max-nextcloud:
I checked and it looks like despite including vue the file is not unreasonably large:
grafik

Would rather invest time to reduce the size of that text include.

mejo- (approved):
Code changes look good to me, thanks for tackling this!

PR #897, opened 2023-09-21
ci: Update workflows from templates

max-nextcloud:
### 📝 Summary

In particular this will avoid downloading cypress when not needed and make the affected runs more stable against network timeouts and failures.

### 🏁 Checklist

- Tests are improved by this

### DIFF
--- a/.github/workflows/lint-eslint.yml
+++ b/.github/workflows/lint-eslint.yml
@@ -56,6 +56,8 @@ jobs:
         run: npm i -g npm@"${{ steps.versions.outputs.npmVersion }}"
 
       - name: Install dependencies
+        env:
+          CYPRESS_INSTALL_BINARY: 0
         run: npm ci
 
       - name: Lint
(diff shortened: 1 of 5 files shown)


.github/workflows/lint-eslint.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
```
mejo-, line 41: Any reason to downgrade the action? Dependabot updates them regularly for our workflows, so I'd rather stick to the latest version if there's no clear reason not to do so.

mejo- (changes requested):
Thanks for looking into this! Apart from downgrading the referenced actions it looks good to me :relaxed:

.github/workflows/lint-eslint.yml
max-nextcloud: uh wow... We're more up to date than the nextcloud one. Maybe we should install dependabot there. But I guess that's another can of worms :worm: 
I'll revert the downgrade and then merge ahead.

mejo- (approved):

Issue #113, opened 2020-10-07
[Feature Request] Dashboard Widget


mejo-:
## Summary

The idea is to have a dashboard widget for Collectives with a list of recently changed pages from the collectives you're part of.

### Backend

We would need a way to retrieve recently changed pages from the backend. Something like a `findRecent()` function in [`PageService.php`](https://github.com/nextcloud/collectives/blob/main/lib/Service/PageService.php) that gets exposed as API endpoint in [`PageController.php`](https://github.com/nextcloud/collectives/blob/main/lib/Controller/PageController.php).

### Frontend 

A dashboard widget that displays the recently changed pages. There's [documentation for writing a dashboard widget](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/dashboard.html)

Not sure whether the pages should be sorted by collective or not. Probably just sorting them by date is the best. The graphical representation could be similar to how the link previews reference collective pages, reusing [`PageReferenceWidget.vue`](https://github.com/nextcloud/collectives/blob/main/src/views/PageReferenceWidget.vue):

![2023-05-15T10:58:13,799910884+01:00](https://github.com/nextcloud/collectives/assets/3582805/59dfcced-5c17-4b73-9410-910edadd578c)

PR #898, opened 2023-09-21
enh(dashboard): add recent pages dashboard widget

blizzz:
Fixes: #113 

Screenshot_20230929_115738

The implementation is not necessarily the most performant one… but using the services of Collectives that already are at hand.

### DIFF
--- a/lib/Dashboard/RecentPagesWidget.php
+++ b/lib/Dashboard/RecentPagesWidget.php
@@ -16,14 +16,28 @@ class RecentPagesWidget implements IReloadableWidget {
 	public const REFRESH_INTERVAL_IN_SECS = 33;
 	public const MAX_ITEMS = 10;
 
+	protected IL10N $l10n;
+	protected IURLGenerator $urlGenerator;
+	protected IUserSession $userSession;
+	protected PageService $pageService;
+	protected CollectiveService $collectiveService;
+	protected RecentPagesService $recentPagesService;
+
 	public function __construct(
-		protected IL10N $l10n,
-		protected IURLGenerator $urlGenerator,
-		protected IUserSession $userSession,
-		protected PageService $pageService,
-		protected CollectiveService $collectiveService,
-		protected RecentPagesService $recentPagesService,
-	) {}
+		IL10N $l10n,
+		IURLGenerator $urlGenerator,
+		IUserSession $userSession,
+		PageService $pageService,
+		CollectiveService $collectiveService,
+		RecentPagesService $recentPagesService
+	) {
+		$this->recentPagesService = $recentPagesService;
+		$this->collectiveService = $collectiveService;
+		$this->pageService = $pageService;
+		$this->userSession = $userSession;
+		$this->urlGenerator = $urlGenerator;
+		$this->l10n = $l10n;
+	}
 
 	public function getItemsV2(string $userId, ?string $since = null, int $limit = 7): WidgetItems {
 		if (!($user = $this->userSession->getUser())) {
--- a/lib/Dashboard/RecentPagesWidget.php
+++ b/lib/Dashboard/RecentPagesWidget.php
@@ -2,10 +2,9 @@
 
 namespace OCA\Collectives\Dashboard;
 
-use OCA\Collectives\Model\CollectiveInfo;
-use OCA\Collectives\Model\PageInfo;
 use OCA\Collectives\Service\CollectiveService;
 use OCA\Collectives\Service\PageService;
+use OCA\Collectives\Service\RecentPagesService;
 use OCP\Dashboard\IReloadableWidget;
 use OCP\Dashboard\Model\WidgetItem;
 use OCP\Dashboard\Model\WidgetItems;
@@ -23,6 +22,7 @@ class RecentPagesWidget implements IReloadableWidget {
 		protected IUserSession $userSession,
 		protected PageService $pageService,
 		protected CollectiveService $collectiveService,
+		protected RecentPagesService $recentPagesService,
 	) {}
 
 	public function getItemsV2(string $userId, ?string $since = null, int $limit = 7): WidgetItems {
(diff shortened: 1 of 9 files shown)


mejo-:
Super nice, thanks @blizzz!

We probably should think about more performant solutions to fetch the list of recent pages for all collectives. What do you think about a dedicated `getRecentPagesForCollectives(array $collectiveIds)` function that gets the most recent pages for a list of collectives from the `oc_filecache` table along with a `JOIN` on `oc_collectives_pages`?

Something along the lines of:

```sql
SELECT p.*, fc.mtime FROM oc_filecache fc LEFT JOIN oc_collectives_pages p ON fc.fileId = p.file_id WHERE (fc.path LIKE 'appdata_123/collectives/13/%' OR fc.path LIKE 'appdata_123/collectives/14/%') AND mimetype=14;
```

mejo-:
By the way, that's the corresponding issue that will be closed with this PR: https://github.com/nextcloud/collectives/issues/113

blizzz:
@mejo- please see the latest commit that targets your suggestions and our conversation in talk.

lib/Service/RecentPagesService.php
```diff
@@ -0,0 +1,118 @@
...
+	) {	}
+
+	/**
+	 * @return RecentPage[]
+	 * @throws MissingDependencyException
+	 * @throws Exception
+	 */
+	public function forUser(IUser $user, int $limit = 10): array {
+		try {
+			$collectives = $this->collectiveService->getCollectives($user->getUID());
```
mejo-, line 34: I wonder whether it would improve readability to rename `$collectives` to something else (e.g. `$userCollectives`) here, given that `$collectives` is used again for a different purpose down below.

lib/Service/RecentPagesService.php
```diff
@@ -0,0 +1,118 @@
...
+		$r = $qb->executeQuery();
+
+		$pages = [];
+		$collectives = [];
+		while ($row = $r->fetch()) {
+			$collectiveId = (int)explode('/', $row['path'], 4)[2];
+			if (!isset($collectives[$collectiveId])) {
+				try {
+					// collectives are not cached, but always read from DB, so keep them
+					$collectives[$collectiveId] = $this->collectiveService->getCollectiveInfo($collectiveId, $user->getUID());
```
mejo-, line 66: If `$collectives` from above wouldn't be overwritten, we could reuse the `CollectiveInfo` object from that array and wouldn't have to do an extra `getCollectiveInfo()` here, no?

lib/Service/RecentPagesService.php
```diff
@@ -0,0 +1,118 @@
...
+			if ($internalPath !== '' && $internalPath !== '.') {
+				$pathParts = array_merge($pathParts, explode('/', $internalPath));
+			}
+			if ($row['filename'] !== 'Readme.md') {
+				$pathParts[] = basename($row['filename'], PageInfo::SUFFIX);
+				$title = basename($row['filename'], PageInfo::SUFFIX);
+			} else {
+				$title = basename($internalPath);
+			}
+			$url = $this->urlGenerator->linkToRoute('collectives.start.indexPath', ['path' => implode('/', $pathParts)]);
```
mejo-, line 89: Probably the `$pathParts` elements have to be URL encoded - they might contain any kind of special characters.

lib/Model/RecentPage.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace OCA\Collectives\Model;
+
+class RecentPage {
```
mejo-, line 5: How about adding a constructor that allows to pass all the values to `new RecentPage()`? We'll probably only fill the `RecentPage` objects once in most cases anyway, no?

lib/Dashboard/RecentPagesWidget.php
```diff
@@ -0,0 +1,158 @@
...
+			if ($result['page']->getFilePath() !== '') {
+				$pathParts = array_merge($pathParts, explode('/', $result['page']->getFilePath()));
+			}
+			if ($result['page']->getFileName() !== 'Readme.md') {
+				$pathParts[] = $result['page']->getTitle();
+			}
+
+			$iconData = $result['page']->getEmoji()
+				? $this->getEmojiAvatar($result['page']->getEmoji())
+				: $this->getEmojiAvatar('🗒');
```
mejo-: It would be more consistent to use the page icon here instead of a close (but different) looking emoji. I see that you already tried to do that, so I would trust your assessment for now that it's hard to get this right.

lib/Model/RecentPage.php
```diff
@@ -0,0 +1,56 @@
...
+	protected string $collectiveName = '';
+	protected string $pageUrl = '';
+	protected string $title = '';
+	protected string $emoji = '🗒';
```
mejo-, line 9: I don't think we should set a default for `$emoji` - in order to allow special-handling of empty emojis in the dashboard widget code.

lib/Service/RecentPagesService.php
```diff
@@ -0,0 +1,118 @@
...
+use OCP\Files\IMimeTypeLoader;
+use OCP\IConfig;
+use OCP\IDBConnection;
+use OCP\IURLGenerator;
+use OCP\IUser;
+
+class RecentPagesService {
+
+	public function __construct(
+		protected CollectiveService $collectiveService,
```
mejo-, line 20: Given that we still target PHP 7.4, constructor property promotion is not possible in Collectives yet.

mejo- (changes requested):
Thanks a lot @blizzz, looks really good overall. I have some minor comments though.

lib/Service/RecentPagesService.php
blizzz: It would end up in double encode actually. I tried with building the URL with PageService::getPageLink (or how it is called) and faced that issue. I did not dig where it is happening, but should be somewhere in the upstream dashboard logic.

lib/Model/RecentPage.php
blizzz: The idea crossed my mind too, but I dislike huge constructors. They make it also a bit less flexible.

blizzz: Note to myself: chain the methods :)

lib/Model/RecentPage.php
blizzz: TODO: remove the default (logic from above is sufficient)

blizzz:
not sure why the cypress tests are failing, this is rather a frontend thing. Similar failures reported at https://github.com/nextcloud/collectives/pull/936

lib/Service/CollectiveHelper.php
```diff
@@ -59,7 +62,8 @@ public function getCollectivesForUser(string $userId, bool $getLevel = true, boo
 				$userPageOrder = ($settings ? $settings->getSetting('page_order') : null) ?? Collective::defaultPageOrder;
 				$userShowRecentPages = ($settings ? $settings->getSetting('show_recent_pages') : null) ?? Collective::defaultShowRecentPages;
 			}
-			$collectiveInfos[] = new CollectiveInfo($c,
+			$collectiveInfos[$c->getId()] = new CollectiveInfo(
```
juliushaertl, line 65: Quickly checked and can reproduce the cypress failure locally when loading the collectives app.

It seems that this changed the API response from an array to an object, so it either needs to revert this and apply the id index elsewhere just for the widgets or make use of an array_values on the controller to avoid that the response format is changed.

blizzz: I'll revert and deal with it on the only consumer using that mapping.

juliushaertl (approved):

mejo- (commented):
I think a small Cypress smoke test that tests whether the recent pages widget displays and lists a page would be good. @blizzz would you like to look into this yourself?

Given that I'd like to release Collectives latest tomorrow we can either postpone this or I'll see whether I find time to do it myself later.

blizzz:
> I think a small Cypress smoke test that tests whether the recent pages widget displays and lists a page would be good. @blizzz would you like to look into this yourself?
> 
> Given that I'd like to release Collectives latest tomorrow we can either postpone this or I'll see whether I find time to do it myself later.

I cannot promise to accomplish this today.

juliushaertl:
Let's follow up on that then I'd say

lib/Service/RecentPagesService.php
```diff
@@ -0,0 +1,127 @@
...
+
+			// prepare link and title
+			$pathParts = [$collectivesMap[$collectiveId]->getName()];
+			if ($internalPath !== '' && $internalPath !== '.') {
+				$pathParts = array_merge($pathParts, explode('/', $internalPath));
+			}
+			if ($row['filename'] !== 'Readme.md') {
+				$pathParts[] = basename($row['filename'], PageInfo::SUFFIX);
+				$title = basename($row['filename'], PageInfo::SUFFIX);
+			} else {
```
mejo-, line 94:
```suggestion
			} elseif ($internalPath === '' || $internalPath === '.') {
				$title = $this->l10n->t('Landing page');
			} else {
```

We have to special-treat landing pages here.

mejo- (changes requested):
Super nice, thanks again for the PR. Only two comments. I have both suggested code changes ready locally and could commit+push them to this PR to get the PR in quickly.

But also fine if you want to do the changes yourself.

mejo-:
I pushed a simple Cypress test.

mejo- (approved):

PR #901, opened 2023-09-24
test(e2e): allow more time for editor to load

max-nextcloud:
Let's see if waiting longer helps.


max-nextcloud:
It's not a matter of too little time. Test also fails locally for me when running `npx cypress run` on master.

Issue #422, opened 2022-11-08
Allow to remove the emoji of a page


mejo-:
In GitLab by @narF_narF on Nov 8, 2022, 15:13

Currently, it's impossible to remove an emoji after it has been set. It would be nice to allow the user to undo its change.

Steps to reproduce:
1. create a page in a Collective
2. assign an emoji to the page

Current result: There are no way to remove the emoji
Expected result: The user can remove the emoji

Note: While it's possible to assign a different emoji, it's not possible to get rid of it entirely

Suggestions:
1. Add a button in the emoji picker to remove the emoji
2. While the emoji picker is open, pressing Delete or Backspace on your keyboard will also get rid of the emoji

narFnarF:
In the meantime, you could detect if the user presses Esc or Del or Backspace on the keyboard while the emoji picker is open, and remove the emoji. It's not ideal because it's hidden and not touch-devices friendly, but it's at least better than the status quo.

Another option would be to display a delete button only when mouse over the emoji. You could even use the ⛔ emoji for that button! 😋 
[](https://i.imgur.com/m9bJ6vl.png)

The note app _Joplin_ moved the emoji picker inside the popup when editing the name of a notebook:
![](https://i.imgur.com/MrjnrC4.png)
Maybe Collectives could display the ⛔ button on the emoji picker when the cursor is inside the title textbox? This way, it would let users on touch devices be able to remove the emoji.

I have more ideas if you're interested.

narFnarF:
For anyone reading this and looking for a solution on how to remove an emoji, a workaround is to go in the File app and locate the file of the Collective page, and then duplicate that file. The copy will not have the emoji. (The downside is that it will break links to that page, if any)

Snafu:
Hello,
I'm facing the same issue as @narFnarF and would like to request some kind of fix as well.

> Sure. We use an external javascript library as emoji picker. Unfortunately, until now, this picker doesn't support to select an "empty emoji". I now went ahead and opened a feature request against the upstream project: [serebrov/emoji-mart-vue#253](https://github.com/serebrov/emoji-mart-vue/issues/253)
> 
> Adding an own action "Remove emoji" would be a bit too heavy and clutter the actions menu in my eyes.
[...]

PR #907, opened 2023-09-26
fix(EmojiPicker): allow to unselect collective and page emoji

mejo-:
Fixes: #422

Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/4575

### 🏁 Checklist

### DIFF
--- a/src/components/Page.vue
+++ b/src/components/Page.vue
@@ -15,7 +15,10 @@
 				<NcEmojiPicker v-else-if="currentCollectiveCanEdit"
 					ref="page-emoji-picker"
 					:show-preview="true"
-					@select="setPageEmoji">
+					:allow-unselect="true"
+					:selected-emoji="currentPage.emoji"
+					@select="setPageEmoji"
+					@unselect="unselectPageEmoji">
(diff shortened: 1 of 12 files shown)


src/components/Page.vue
```diff
@@ -279,6 +282,10 @@ export default {
 			await this.setEmoji(this.currentPage.parentId, this.currentPage.id, emoji)
 		},
 
+		async unselectPageEmoji() {
+			await this.setPageEmoji('')
```
max-nextcloud, line 286: I don't think there's a reason to use async await here. How about this:
```suggestion
		unselectPageEmoji() {
			return this.setPageEmoji('')
```

If `this.setPageEmoji` is async i.e. returns a promise so will `unselectPageEmoji`.

max-nextcloud (approved):
Looks good... even tests included and covering both pages and collectives.
:100: :grinning: 

Just one small comment.

src/components/Page.vue
mejo-: Thanks, sounds good to me :ok_hand:

PR #911, opened 2023-09-28
Remove double negatives in docs

pjrobertson:
Very small edit. I was just reading these docs changes, and got confused by the double negative about "not disabling". I tried to update this to be clearer.


max-nextcloud (approved):
My understanding is that enabled is the default and that's where the double negative comes from. If you didn't actively disable it everything is fine.

However I agree that readability is improved and hardly anyone remembers exactly what they changed in the settings - so i approve this.

Thanks for taking care of the details!

mejo- (changes requested):
Thanks a lot for the pull request, @pjrobertson!

> My understanding is that enabled is the default and that's where the double negative comes from. If you didn't actively disable it everything is fine.

Exactly, that's why I chose the initial wording. That's also, why I would prefer to not merge it as is. The way it is read with the proposed change, people probably think that they have to change something per default to make public shares work, which is not the case. @pjrobertson if you have another suggestion that improves readability while maintaining that information, I'm open to it for sure :blush:

pjrobertson:
>  @pjrobertson if you have another suggestion that improves readability while maintaining that information, I'm open to it for sure 😊

What you said makes sense! I agree with Max that we should expect people to remember what settings they have in the admin, but also I understand the double negative. I've provided a different wording which hopefully clarifies both points.

mejo- (approved):

PR #918, opened 2023-10-03
test(ci): do not use group parameter when run without cypress cloud

max-nextcloud:
Cypress is currently failing on the main branch with:
```
You passed the --ci-build-id, --group, --tag, --parallel, or --auto-cancel-after-failures flag without also passing the --record flag.

The --ci-build-id flag you passed was: Cypress - 2f6588277fb008bc1e918726f469df2e01e50d1d

These flags can only be used when recording to Cypress Cloud.
```

Frankly... I have no idea why as i thought all these options were blank when not using cypress cloud.
So this is an attempt at fixings this.

The `ci-build-id` is not in our yml file and will only be passed if either `parallel` or `group` are given.
https://github.com/cypress-io/github-action/blob/bd9dda317ed2d4fbffc808ba6cdcd27823b2a13b/index.js#L667-L680

So I suspect we're still passing group.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -215,11 +215,11 @@ jobs:
         with:
           record: '${{ !!matrix.use-cypress-cloud }}'
           parallel: '${{ !!matrix.use-cypress-cloud }}'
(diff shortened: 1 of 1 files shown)


max-nextcloud:
I'll actually propose to not run cypress tests on integration branches anymore - or maybe once a week.

PR #923, opened 2023-10-04
test(update): run app update test with mysql 8.0

max-nextcloud:
Let's see if we can reproduce the update failure to 2.8.1


max-nextcloud:
app upgrade fails as expected and the logs contain the error message reported by users.

mejo- (approved):

Issue #917, opened 2023-10-02
Update to v2.8.1 crashed


koebi80:
**Describe the bug**
The update newest release v2.8.1 of the collective app crashed.

The following error message apears: 
Exception: Database error when running migration 020800Date20230911000000 for app collectives An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1101 BLOB, TEXT, GEOMETRY or JSON column 'settings' can't have a default value

**To Reproduce**
- Start update from v2.7.1 to v2.8.1 within the admin app panel.

**Expected behavior**

**Screenshots**
![Bildschirmfoto 2023-10-02 um 21 15 01](https://github.com/nextcloud/collectives/assets/106907779/7d1ec042-3689-41ee-8226-94cca6b83821)

**Client details:**

<details>
<summary>Server details</summary>

**Collectives app version:** (see Nextcloud apps page)
Update from v2.7.1 to v2.8.1

**Operating system:**

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** (see Nextcloud admin page)
26.0.2 or 27.1.1

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

nolens:
I have exactly the same issue.

nolens:
> I have exactly the same issue.
@max-nextcloud 
mysql 8.0.33-25 in my case. 

PS: 
I have downgraded to an earlier version as a workaround for now.

Thanks for taking care of this!

afgeissler:
i have the same problem with NC 27.1.1 and mysql 8.0.34

koebi80:
On System with NC 27.1.1 is MySQL 5.7.42

On System with NC 26.0.2 (managed system, no access to console or server infos)

max-nextcloud:
The 2.8.0 and 2.8.1 releases have been removed from the appstore to prevent others from running into this issue.

We also have a failing test for the upgrade with mysql now #923 here.

max-nextcloud:
> On System with NC 27.1.1 is MySQL 5.7.42

Independent of this issue - Please check if you can update your mysql version. The last nextcloud version to support mysql 5.7 is Nextcloud 20. So it's been a while since we don't test mysql 5.7 compatibility anymore.

[...]

PR #924, opened 2023-10-04
fix(db): make settings field from collectives_user_settings type string

mejo-:
JSON as DB field type causes too much headache, let's do the JSON parsing in PHP instead.

Fixes: #917

### DIFF
--- a/lib/Migration/Version010200Date20220616000000.php
+++ b/lib/Migration/Version010200Date20220616000000.php
@@ -5,8 +5,8 @@ declare(strict_types=1);
 namespace OCA\Collectives\Migration;
 
 use Closure;
-use Doctrine\DBAL\Types\Types;
 use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
 use OCP\Migration\IOutput;
 use OCP\Migration\SimpleMigrationStep;
 
--- a/lib/Migration/Version010200Date20220616000000.php
+++ b/lib/Migration/Version010200Date20220616000000.php
@@ -37,6 +37,11 @@ class Version010200Date20220616000000 extends SimpleMigrationStep {
 			]);
 			$table->addColumn('page_order', Types::INTEGER, [
 				'notnull' => true,
+				'default' => 0,
+			]);
+			$table->addColumn('settings', Types::STRING, [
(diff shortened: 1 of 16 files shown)


max-nextcloud (approved):

psalm.xml
```diff
@@ -28,7 +28,7 @@
     <issueHandlers>
         <UndefinedClass>
             <errorLevel type="suppress">
-                <referencedClass name="Doctrine\DBAL\Types\Types"/>
+                <referencedClass name="Doctrine\DBAL\Types\Type" />
             </errorLevel>
         </UndefinedClass>
```
nickvergessen, line 33: you removed the usages, so remove this whole block?

lib/Migration/Version010200Date20220616000000.php
```diff
@@ -38,6 +38,10 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$table->addColumn('page_order', Types::INTEGER, [
 				'notnull' => true,
 			]);
+			$table->addColumn('settings', Types::STRING, [
```
nickvergessen, line 41: should not be needed, the later migration will take care

lib/Migration/Version020802Date20231004000000.php
nickvergessen, line 1: It is not recommended to remove migrations once released, as it can give you future headaches if a DB is inconsistent with your expectation and there is a migration mentioned which you don't find in the source code :)

nickvergessen (approved):
But fine in general

psalm.xml
mejo-: We use `Doctrine\DBAL\Types\Type` now, so it's still needed.

lib/Migration/Version010200Date20220616000000.php
mejo-: I prefer to not have the final schema split all over the migrations, but probably just a matter of taste.

lib/Migration/Version020802Date20231004000000.php
mejo-: I see the point, will keep it in mind for the future :grimacing:

Issue #330, opened 2022-05-11
Log error: preg_match(): Compilation failed: quantifier does not follow a repeatable item at offset 18


mejo-:
In GitLab by @juliushaertl on May 11, 2022, 14:33

Found in the server logs

```
{
  "reqId": "9EWw2EoRZHbAqQCRvan0",
  "level": 3,
  "time": "2022-05-05T08:57:11+00:00",
  "remoteAddr": "REMOVED",
  "user": "REMOVED",
  "app": "PHP",
  "method": "GET",
  "url": "/ocs/v2.php/search/providers/collectives-page-content/search?term=*copy+fpp&from=[Version handling](https://github.com/mejo-/collectives/milestone/2)Fapps%2Fcollectives%2FNextcloud%2520Test%2FTest2%2FOnboarding%3FfileId%3D5709100",
  "message": "preg_match(): Compilation failed: quantifier does not follow a repeatable item at offset 18 at /var/www/html/apps/collectives/lib/Search/PageContentProvider.php#115",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/101.0.4951.41 Safari/537.36",
  "version": "24.0.0.11",
  "exception": {
    "Exception": "Error",
    "Message": "preg_match(): Compilation failed: quantifier does not follow a repeatable item at offset 18 at /var/www/html/apps/collectives/lib/Search/PageContentProvider.php#115",
    "Code": 0,
    "Trace": [
      {
        "function": "onError",
        "class": "OC\\Log\\ErrorHandler",
        "type": "::"
      },
      {
        "file": "/var/www/html/apps/collectives/lib/Search/PageContentProvider.php",
        "line": 115,
        "function": "preg_match"
      },
      {
        "file": "/var/www/html/lib/private/Search/SearchComposer.php",
        "line": 160,
        "function": "search",
        "class": "OCA\\Collectives\\Search\\PageContentProvider",
        "type": "->"
      },
      {
        "file": "/var/www/html/core/Controller/UnifiedSearchController.php",
        "line": 115,
        "function": "search",
        "class": "OC\\Search\\SearchComposer",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 225,
        "function": "search",
        "class": "OC\\Core\\Controller\\UnifiedSearchController",
        "type": "->"
      },
      {
[...]

PR #928, opened 2023-10-05
fix(PageService): preg_quote `trusted_domains` before using in regexp

mejo-:
### 📝 Summary

Fixes: #330

### DIFF
--- a/lib/Service/PageService.php
+++ b/lib/Service/PageService.php
@@ -971,8 +971,10 @@ class PageService {
 		$suffix = '\)/';
 
 		$protocol = 'https?:\/\/';
-		$trustedDomainConfig = (array)$this->config->getSystemValue('trusted_domains', []);
-		$trustedDomains = $trustedDomainConfig !== [] ? '(' . implode('|', $trustedDomainConfig) . ')' : 'localhost';
+		$trustedDomainArray = array_map(static function (string $domain) {
+			return preg_quote($domain, '/');
+		}, (array)$this->config->getSystemValue('trusted_domains', []));
+		$trustedDomains = $trustedDomainArray !== [] ? '(' . implode('|', $trustedDomainArray) . ')' : 'localhost';
 
 		$basePath = str_replace('/', '/+', str_replace('/', '/+', preg_quote(trim(\OC::$WEBROOT, '/'), '/'))) . '(\/+index\.php)?';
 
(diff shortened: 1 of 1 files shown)


lib/Service/PageService.php
```diff
@@ -971,8 +971,10 @@ public function matchBacklinks(PageInfo $pageInfo, string $content): bool {
 		$suffix = '\)/';
 
 		$protocol = 'https?:\/\/';
-		$trustedDomainConfig = (array)$this->config->getSystemValue('trusted_domains', []);
-		$trustedDomains = $trustedDomainConfig !== [] ? '(' . implode('|', $trustedDomainConfig) . ')' : 'localhost';
+		$trustedDomainArray = array_map(static function (string $domain) {
+			return preg_quote($domain, '/');
+		}, (array)$this->config->getSystemValue('trusted_domains', []));
```
max-nextcloud, line 976: `preg_quote` will quote `*` and thus remove the special meaning it has inside `trusted_domains`. So for example with a trusted domains of `*.my.tld` the resulting regexp would fail to find a absolute link using `www.my.tld` as the domain name.

I'm pretty sure this did not work before either - but still, while we're at it...

I wonder if replacing `\*` with `\w*` after the preg replace would do the trick.

mejo-: Edit: ah, just saw that you did this :blush:

lib/Service/PageService.php
```diff
@@ -971,8 +971,10 @@ public function matchBacklinks(PageInfo $pageInfo, string $content): bool {
 		$suffix = '\)/';
 
 		$protocol = 'https?:\/\/';
-		$trustedDomainConfig = (array)$this->config->getSystemValue('trusted_domains', []);
-		$trustedDomains = $trustedDomainConfig !== [] ? '(' . implode('|', $trustedDomainConfig) . ')' : 'localhost';
+		$trustedDomainArray = array_map(static function (string $domain) {
+			return preg_quote($domain, '/');
```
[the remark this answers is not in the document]
mejo-: Implemented this now.

lib/Service/PageService.php
max-nextcloud: > I'm not sure if wildcards in trusted domains would even be supported

docs say so:

grafik

max-nextcloud: https://docs.nextcloud.com/server/27/admin_manual/configuration_server/config_sample_php_parameters.html?highlight=trusted_domains

max-nextcloud (approved):

juliushaertl (approved):

PR #961, opened 2023-10-23
chore(php): Add PHP 8.3 to test matrix

juliushaertl:
To be prepared for the bump with Nextcloud 28

### DIFF
--- a/.github/workflows/behat-sqlite.yml
+++ b/.github/workflows/behat-sqlite.yml
@@ -43,6 +43,8 @@ jobs:
         include:
           - php-versions: '7.4'
             server-versions: 'stable25'
+          - php-versions: '8.3'
+            server-versions: 'master'
 
     steps:
       - name: Set app env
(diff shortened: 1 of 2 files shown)


.github/workflows/cypress.yml
```diff
@@ -133,7 +133,7 @@ jobs:
       matrix:
         node-version: [20]
         containers: [1, 2, 3]
-        php-versions: ['8.0', '8.2']
+        php-versions: ['8.0', '8.2', '8.3']
```
juliushaertl, line 136: @mejo- I think we currently don't conver the minimum supported php version anywhere. Any thought about running stable25 against 7.4 rather than 8.0? We could also run phpunit against all but i would pick the one that you feel has the better coverage for lines of code.

mejo-: No objections at all. We run the behat tests against PHP 7.4 on stable25 though, and my gut feeling is that we test the API endpoints pretty well with the behat tests. Not sure if Cypress covers more here.

mejo-: Thinking more about it, I'd also suggest to add PHP8.3 to the `behat-sqlite` matrix, given that these tests most likely have the highest coverage of PHP code.

PR #963, opened 2023-10-24
Copy/move pages in/between collectives

### LINKED ISSUE #488: Allow to copy pages to another Collective

The "move page" modal should allow to copy the page into another collective.

From design review call:
* Additionally would allow copying pages to another collective ("My collectives" -> "List of Collectives" in breadcrumbs?)

Follow-up from #478

mejo-:
### 📝 Summary

* Move pages to other collectives via the move page modal
* Copy pages in/between collectives via the move page modal
* Resolves: #488

Also includes quite some refactoring.

#### 🖼️ Screenshots

Page list action | Copy/move modal default | Copy/move modal collectives list
---|--- | ---
grafik | grafik |grafik

### Screencast

[recording1.webm](https://github.com/nextcloud/collectives/assets/3582805/4a96f86d-768b-4f74-b97e-1524545e58de)

### 🏁 Checklist


jancborchardt (dismissed):
Very nice, just small points:
- [x] Order should be "Move or copy", like we do in Files
- [x] "Move page here" could be the primary button, like in Files

mejo-:
>  "Move page here" could be the primary button, like in Files

How about when copying/moving to another collective? Do we still want a primary action there? As moving the page is potentially destructive to users who are part of source collective and not of the target collective.

nimishavijay:
> How about when copying/moving to another collective? Do we still want a primary action there? As moving the page is potentially destructive to users who are part of source collective and not of the target collective.

In that case we could change the text to "Move page to [collective name]" when you navigate into another collective, it will catch the eye when the text changes, that could act as an error prevention measure

Additionally, right now pages and collectives look exactly the same, which also may result in accidentally moving a page out of a collective. So we could add some info in `--color-text-maxcontrast` on the subline of a collective (eg, "70 members")

juliusknorr (approved):
Tested and works 👍 

Also no further objections on the code 🎉

mejo-:
> > How about when copying/moving to another collective? Do we still want a primary action there? As moving the page is potentially destructive to users who are part of source collective and not of the target collective.
> 
> In that case we could change the text to "Move page to [collective name]" when you navigate into another collective, it will catch the eye when the text changes, that could act as an error prevention measure
> 
> Additionally, right now pages and collectives look exactly the same, which also may result in accidentally moving a page out of a collective. So we could add some info in `--color-text-maxcontrast` on the subline of a collective (eg, "70 members")

Sounds like a good plan. Should the "Copy page here" button also change to "Copy page to [collective name]" in this case? For Collectives with long names, the name most likely would be cut off then.

| Only move button with collective name | Both buttons with collective name |
| --- | --- |
| grafik | grafik |

@nimishavijay

mejo-:
I really wonder whether we should make move and copy both secondary actions when navigating inside another collective. What do you think @nimishavijay and @jancborchardt?

mejo-:
> Additionally, right now pages and collectives look exactly the same, which also may result in accidentally moving a page out of a collective. So we could add some info in --color-text-maxcontrast on the subline of a collective (eg, "70 members")

I agree that a visual distinction would be good. We don't have the information how many members the collective has, though. So this would come with extra cost. Maybe we find another solution for that?

jancborchardt:
@mejo- feedback:
- I think including the Collective name is good when moving to another collective :) Yes it has to be ellipsized of course when there is not enough space but that is fine
- "Move" should stay on the right, as primary actions do for us
- I would say it should stay primary even when moving to a different collective.

nimishavijay:
Agreed with Jan! :)

> I agree that a visual distinction would be good. We don't have the information how many members the collective has, though. So this would come with extra cost. Maybe we find another solution for that?

Ah alright, We could maybe then just make the text bold and add some text "All collectives" between the breadcrumbs bar and collectives list in `--color-primary-element-text`. Also checked out what Confluence did for this action and they have used a dropdown component, so that's another option to consider. Quick mockups:

image


Also, another point of feedback:
- [ ] The page name in the breadcrumbs bar is ellipsizing very early even if there is enough space, we could definitely show more of long page names in the breadcrumbs bar :)
- [ ] Regarding the primary action, it could be "Move", we should change it in Files also :)

mejo-:
> Ah alright, We could maybe then just make the text bold and add some text "All collectives" between the breadcrumbs bar and collectives list in `--color-primary-element-text`.

Mh, not sure whether this will work. `--color-primary-element-text` is white by the way, so probably you meant something else, no? Is that how you imagined it?

grafik

nimishavijay:
> Mh, not sure whether this will work. --color-primary-element-text is white by the way, so probably you meant something else, no? Is that how you imagined it?

Ah you are right, my bad. I meant in the primary color itself. I played around with it more and how about showing the "All collectives" right next to the collectives icon, something like this: 

image

If that is not possible we could just stick to making it bold :)

mejo-:
> Ah you are right, my bad. I meant in the primary color itself. I played around with it more and how about showing the "All collectives" right next to the collectives icon, something like this:

Ah, great idea. Implemented this now. I left the opacity to 0.5 though because effectively it's a disabled button:

2023-10-26T16:12:44,509155487+02:00

nimishavijay (approved):
Super! Looks great now :)

jancborchardt (commented):
- [ ] "All collectives" is a bit unclear. It looks more like a heading, maybe it needs a "Back" arrow to left icon?
- [ ] If a button is disabled, it should be done in markup. We can't play around with opacity manually as that is not accessible. (If it's not a button, text that is not active needs to be color-text-maxcontrast)

mejo-:
> * "All collectives" is a bit unclear. It looks more like a heading, maybe it needs a "Back" arrow to left icon?
> * If a button is disabled, it should be done in markup. We can't play around with opacity manually as that is not accessible. (If it's not a button, text that is not active needs to be color-text-maxcontrast)

Well, the Collectives icon along with "All collectives" is the disabled button when no collective is selected. In this case, the button has no effect because the collectives overview is already displayed. When a collective is selected, the button title "All collectives" is set to empty and only the Collectives icon is the button.

I don't manually change the opacity. It's set by the button component.

Does this make sense for you @jancborchardt?

mr-gosh:
Why can this not be used for copying a whole collective?
Can't find a propper way of doing that...

max-nextcloud:
@mr-gosh Good point. Would you mind opening a new issue for it?

Workarounds i can think of:
* Copying files in the collectives folder in the files app... does not include emojis
* Copying all pages one after the other... cumbersome
* moving all pages into one page, copy that, moving them back... also cumbersome.

PR #985, opened 2023-11-06
chore(ci): Update Github workflows from templates

mejo-:


AndyScherzinger (approved):

.github/workflows/cypress.yml
```diff
@@ -1,46 +1,48 @@
...
+            src:
+              - '.github/workflows/**'
+              - 'src/**'
+              - 'appinfo/info.xml'
+              - 'package.json'
+              - 'package-lock.json'
+              - 'tsconfig.json'
+              - '**.js'
+              - '**.ts'
+              - '**.vue'
```
max-nextcloud, line 38: I think we should also run cypress on changes to the backen - i.e. `lib`.

max-nextcloud: While you're at it you could make the workflow expression more precise and only run on changes to the cypress workflow.

max-nextcloud (approved):
Two minor comments. Other than that looks good.

Thanks for taking care of this :heart:

.github/workflows/cypress.yml
mejo-: Well, the workflow regex is the default from all upstream templates, I'd rather keep it that way to not diverge more. But agreed, `lib` should be added. Good catch :ok_hand:

PR #986, opened 2023-11-06
tests(cy): use @nextcloud/cypress login and logout

max-nextcloud:
They authenticate using the API rather than the UI. This is much faster.

Visiting the desired url then happens separately.

### 🏁 Checklist


mejo- (approved):
Super nice, thanks so much for this :heart:

PR #987, opened 2023-11-06
fix(composer): Add dependency on `php-pdo_sqlite`

mejo-:
Our dependency tntsearch requires it.

Fixes psalm and php lint CI workflows with PHP 8.2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
         "php": ">=7.4",
         "ext-json": "*",
         "ext-pdo": "*",
+        "ext-pdo_sqlite": "*",
         "teamtnt/tntsearch": "^3.0"
     },
     "require-dev": {


blizzz (approved):
Might then also should have an entry as `lib` in the `dependencies` section in info.xml https://nextcloudappstore.readthedocs.io/en/latest/developer.html?highlight=php#info-xml (although i do not think it is enforced anywhere)

mejo-:
> Might then also should have an entry as `lib` in the `dependencies` section in info.xml https://nextcloudappstore.readthedocs.io/en/latest/developer.html?highlight=php#info-xml (although i do not think it is enforced anywhere)

Thanks @blizzz, I was not aware of this `info.xml` property. A quick grep revealed that it's rarely used. But nevertheless good to have it defined if it exists. I opened a follow-up PR for this: https://github.com/nextcloud/collectives/pull/991

PR #988, opened 2023-11-06
Javascript wrapper for collectives API

max-nextcloud:
Wrap all usage of the collectives api in a simple functional interface.
This will allow us to change the layout of the api
and only update it in a single place.

## Refactor client (browser) app

Separate the API handling from the store.
The api is concerned with axios and generateUrl
while the store is only concerned with the current state of the app.

### Url generation from parts

Accept multiple arguments to `collectivesUrl` and `pagesUrl`
and join them with `/` to build the url:

Example:
```js
api.pagesUrl({collectiveId: 10}, 'trash', 12)
```
results in `/apps/collectives/_api/10/_pages/trash/12`

## Tests (Cypress)

Use the api directly in the cypress commands.
That way we do not require the page to be loaded.

Also extract `collective-edit-mode.spec.js` from `pages.spec.js`.
It requires a different beforeEach function
to seed the edit mode before visiting the collective.

Signed-off-by: Max <max@nextcloud.com>

## 🏁 Checklist

### DIFF
--- a/cypress/e2e/circle-with-group.spec.js
+++ b/cypress/e2e/circle-with-group.spec.js
@@ -32,7 +32,6 @@
 describe('Pages are accessible via group membership to circle', function() {
 	before(function() {
 		cy.loginAs('jane')
-		cy.visit('apps/collectives')
 		cy.deleteAndSeedCollective('Group Collective')
 		cy.circleFind('Group Collective')
 			.circleAddMember('Bobs Group', 2)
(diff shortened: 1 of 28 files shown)


max-nextcloud:
@mejo- I still have some more ideas here but i feel like this could already go in and I can open a new PR for the next steps.

I'd be happy about a review if you can spare some time. If you're busy with other things that's also fine. :slightly_smiling_face:

mejo-:
@max-nextcloud after rebase quite a few cypress tests fail now on all NC versions. Could you take a look?

mejo- (approved):
I skimmed through the changes and they look good to me. Thanks a lot for all the refactoring work, the whole requests logic in frontend is so much cleaner now :heart:

PR #991, opened 2023-11-07
chore(info.xml): Add `ext-pdo` and `ext-pdo_sqlite` as lib dependencies

mejo-:


appinfo/info.xml
```diff
@@ -44,6 +44,8 @@ In your Nextcloud instance, simply navigate to **»Apps«**, find the
 	</screenshot>
 	<dependencies>
 		<nextcloud min-version="25" max-version="27" />
+		<lib>ext-pdo</lib>
+		<lib>ext-pdo_sqlite</lib>
```
blizzz, line 48:
```suggestion
		<lib>pdo</lib>
		<lib>pdo_sqlite</lib>
```

Never seen the ext- prefix used in that context

blizzz (changes requested):

blizzz (approved):

appinfo/info.xml
mejo-: Ok, I was not sure which one is correct. In `package.json` it's `ext-pdo`. Changed now.

PR #995, opened 2023-11-07
fix(LazyFolder): FileInfo interface added `getMedata` function

mejo-:


mejo-:
App upgrade checks will fail now on master until we publish a new Collectives release with this change.

blizzz (approved):

PR #1003, opened 2023-11-09
fix(NewCollectiveModal): Mention that own user is part of collective

mejo-:
People got confused by searching their own user.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
screenshot | recording1

### 🏁 Checklist

### DIFF
--- a/src/components/Nav/NewCollectiveModal.vue
+++ b/src/components/Nav/NewCollectiveModal.vue
@@ -86,6 +86,7 @@
 				<div class="modal-collective-members">
 					<MemberPicker :show-selection="true"
 						:selected-members="selectedMembers"
+						:no-delete-members="noDeleteMembers"
 						:on-click-searched="onClickSearched"
 						@delete-from-selection="deleteMember" />
 				</div>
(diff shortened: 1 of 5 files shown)


src/components/Nav/NewCollectiveModal.vue
```diff
@@ -87,6 +87,7 @@
 					<MemberPicker :show-selection="true"
 						:selected-members="selectedMembers"
 						:on-click-searched="onClickSearched"
+						:empty-content-hint="t('collectives', 'Your user is added automatically.')"
```
mejo-, line 90: Not sure about the wording. Maybe "As creator, you're already added." is better? :thinking:

jancborchardt (commented):
Instead of as a message, we could show the owner's account directly below in the list like added accounts do? That would make it more obvious than a line of text.

Slightly related: We want to move from the wording "user" to either "account" or "person", could be replaced in the search field.

mejo-:
Thanks @jancborchardt, much better. Implemented now, see the screencast above. I also changed "users" to "accounts" where appropriate.

mejo-:
I wonder whether it would further help to change the "Create" button to something like "Create without adding people" in case no other members were selected.

juliushaertl:
Small addition, we could preserve the previously entered collective name in the dialog title to make it more obvious what this is about, e.g. "Add people to fox" instead of just "Add people".

Overall i find the different usages of account/people/members a bit confusing. 

Screenshot 2023-11-09 at 19 56 28


Screenshot 2023-11-09 at 19 55 35

Screenshot 2023-11-09 at 19 55 54

Maybe we can somehow unify that? Pick one high level word (members) and a fixed set for the actual membership entities (account, group, circle)?

mejo-:
> Overall i find the different usages of account/people/members a bit confusing.
> Maybe we can somehow unify that? Pick one high level word (members) and a fixed set for the actual membership entities (account, group, circle)?


I agree, thanks for bringing this up. Your suggestion makes sense: account, group, circle for the different entities and "members" for all of them. So I'd replace "Add people" with "Add members" and search for more occurrences of "users" to be replaced by "accounts". Agreed?

mejo-:
> Small addition, we could preserve the previously entered collective name in the dialog title to make it more obvious what this is about, e.g. "Add people to fox" instead of just "Add people".

Great idea. Implemented now (using "Add members to <name>"). See updated screencast above.

> I agree, thanks for bringing this up. Your suggestion makes sense: account, group, circle for the different entities and "members" for all of them. So I'd replace "Add people" with "Add members" and search for more occurrences of "users" to be replaced by "accounts". Agreed?

Also implemented. See updated screencast above.

juliushaertl (approved):

Issue #1007, opened 2023-11-11
Backlinks not showing


Jeefongithub:
**Describe the bug**

When creating links to pages using the smart picker the page containing the link is not shown in the page where the link connects to

**To Reproduce**
1. Open page in collectives
2. use / or click link & smart picker
3. select collective page
4. Find page to link to and select
5. Once link is created follow link to selected page
6. Once in linked page open sidebar and select tab for backlinks
7. verify Backlinks section shows "No backlinks available"

**Expected behavior**

Expected behavior is backlinks section should show the pages that contain links to said page

**Screenshots**

First screen shot shows a page that contains a link to a different page
![Screenshot from 2023-11-11 16-09-10](https://github.com/nextcloud/collectives/assets/53583661/1e8cd05b-5029-410d-af8f-8e5b421f6407)

Second screenshot shows the linked page with the sidebar showing no backlinks
![Screenshot from 2023-11-11 16-11-49](https://github.com/nextcloud/collectives/assets/53583661/6f3c0f69-9463-436e-ad6e-a91ca9e8f630)

**Server details:**
- Collectives app version: 2.9.0
- Nextcloud version: 27.1.3
- PHP Version: 8.2.11
- Database: mariadb 10.5.21

**Client details:**
 - OS: Debian 12
 - Browser: Mercury (Firefox Fork)
 - Browser version: 115.4.0esr
 - Device: desktop

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```
[nextcloud.log](https://github.com/nextcloud/collectives/files/13326527/nextcloud.log)

#### Browser log
```
Insert your browser log here, this could for example include:
[console-export-2023-11-11_16-21-35.txt](https://github.com/nextcloud/collectives/files/13326532/console-export-2023-11-11_16-21-35.txt)


a) The javascript console log
b) The network log
c) ...
```

</details>

mejo-:
Thanks a lot for the bugreport @Jeefongithub. I was able to reproduce. A fix will follow soon.

PR #1012, opened 2023-11-13
fix(backlinks): Accept `<link>` syntax as valid in parsed markdown

mejo-:
For example, the smart picker inserts links in this format.

Fixes: #1007


datenangebot (approved):
looks good

Issue #1010, opened 2023-11-13
Show parent pages on the dashboard widget


provokateurin:
**Is your feature request related to a problem? Please describe.**
The new dashboard widget is awesome, but it only shows the collective and page name. Often times this might not be enough context about which page it actually is, because there might be two pages that have the same name or the names are just not well chosen.

**Describe the solution you'd like**
The dashboard widgets should also display the parent pages in some way to indicate the exact location of the page. This could either be in the title or in the subtitle depending on how visible it should be.

PR #1013, opened 2023-11-14
fix(dashboard): Display parent page path in recent pages widget

mejo-:
Fixes: #1010

#### 🖼️ Screenshots

:derelict_house:  Before | 🏡 After
|---|---
2023-11-14T11:36:03,256783199+01:00 | 2023-11-14T11:35:32,870722255+01:00

### DIFF
--- a/lib/Dashboard/RecentPagesWidget.php
+++ b/lib/Dashboard/RecentPagesWidget.php
@@ -40,9 +40,12 @@ class RecentPagesWidget implements IReloadableWidget {
 
 		$items = [];
 		foreach ($recentPages as $recentPage) {
+			$subtitle = $recentPage->getPagePath()
+				? $recentPage->getCollectiveName() . ': ' . $recentPage->getPagePath()
+				: $recentPage->getCollectiveName();
 			$items[] = new WidgetItem(
 				$recentPage->getTitle(),
-				$recentPage->getCollectiveName(),
+				$subtitle,
 				$recentPage->getPageUrl(),
 				'data:image/svg+xml;base64,' . base64_encode($this->getEmojiAvatar($recentPage->getEmoji() ?: '🗒')),
 				(string)$recentPage->getTimestamp()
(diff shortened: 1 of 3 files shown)


provokateurin (approved):
Maybe a `/` without any space would be better? Then less of the parent page names are cut off

mejo-:
> Maybe a / without any space would be better? Then less of the parent page names are cut off if the subtitle is too long

I tried this before, but I think it's important to separate the collective name from the path to make it easy to grasp which collective this is all about.

blizzz:
The slash is part of the path. Maybe instead of truncating in at the end, the ellipsis should be in the middle, or even beginning? But that would be work on the dashboard side of things.

We could save a bit of space using a colon and sparing the first whitespace, so `: ` instead of ` > ` though.

provokateurin:
Ah sorry I thought the pages were also separate by the `>`. Then this already does what I suggested :)

mejo-:
> The slash is part of the path. Maybe instead of truncating in at the end, the ellipsis should be in the middle, or even beginning? But that would be work on the dashboard side of things.

Yeah, I'd say it's out of scope here.

> We could save a bit of space using a colon and sparing the first whitespace, so : instead of > though.

Well, one character less. But yes, let's do that :blush:

mejo-:
Screenshots updated

blizzz:
> > We could save a bit of space using a colon and sparing the first whitespace, so : instead of > though.
> 
> Well, one character less. But yes, let's do that 😊

It's not monospace, so a few pixels more :superhero:

blizzz (approved):

juliushaertl:
In the page title we actually use "Page - Parent page - collective", maybe that would be an option to be consistent with separators?

blizzz:
> In the page title we actually use "Page - Parent page - collective", maybe that would be an option to be consistent with separators?

There we don't have an issue with screen estate though, and typically you know on which collective you are, when you open that page. On a concise overview, you do not have this context, making it more important to provide the collective info, I'd say.

Issue #968, opened 2023-10-26
Collectives App cause nextcloud's cron.php to fail with exception


andwei:
**Describe the bug**
Running the nextcloud cron job causes an exception from Collectives Code
```
ArgumentCountError: Too few arguments to function OCA\Collectives\BackgroundJob\ExpirePageTrash::__construct(), 0 passed in /var/www/html/lib/private/BackgroundJob/JobList.php on line 322 and exactly 3 expected in /var/www/html/custom_apps/collectives/lib/BackgroundJob/ExpirePageTrash.php:18
Stack trace:
#0 /var/www/html/lib/private/BackgroundJob/JobList.php(322): OCA\Collectives\BackgroundJob\ExpirePageTrash->__construct()
#1 /var/www/html/lib/private/BackgroundJob/JobList.php(227): OC\BackgroundJob\JobList->buildJob(Array)
#2 /var/www/html/cron.php(145): OC\BackgroundJob\JobList->getNext(false)
#3 {main}
```

**To Reproduce**
`/usr/local/bin/php -f /var/www/html/cron.php`

**Expected behavior**
The cronjob does not throw an exception. Deactivating the Collectives App using the admin menu makes cron.php to run wihtout errors.

<details>

**Collectives app version:** 2.8.2

**Operating system:** Debian bullseye

**Web server:** nginx version: nginx/1.25.3

**Database:** /usr/bin/mariadb from 11.1.2-MariaDB

**PHP version:** PHP 8.2.11 (cli) (built: Oct 21 2023 04:44:42) (NTS)

**Nextcloud version:** Nextcloud Hub 6 (27.1.2)  (Docker official nextcloud Image tagged nextcloud:stable-fpm-alpine)

</details>

mejo-:
Apparently this happens if the `files_trashbin` app is disabled.

PR #1016, opened 2023-11-14
fix(BackgroundJob): Don't make `ExpirePageTrash` choke without trash

mejo-:
Fixes: #968

### 📝 Summary

* Resolves: # 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist


lib/BackgroundJob/ExpirePageTrash.php
```diff
@@ -14,18 +14,20 @@
 class ExpirePageTrash extends TimedJob {
 	private Expiration $expiration;
 	private PageTrashBackend $trashBackend;
+	private bool $hasTrashBackend = false;
 
 	public function __construct(
-		ITimeFactory $time,
-		Expiration $expiration,
-		PageTrashBackend $trashBackend) {
+		ITimeFactory $time) {
```
juliushaertl, line 20: Starting with nextcloud 28 you could actually make them nullable, but good as it is as we cover more versions ;)

https://github.com/nextcloud/server/pull/41265

lib/BackgroundJob/ExpirePageTrash.php
```diff
@@ -14,18 +14,20 @@
...
 
 		// Run once per hour
 		$this->setInterval(60 * 60);
 
-		$this->expiration = $expiration;
-		$this->trashBackend = $trashBackend;
+		if (class_exists(Expiration::class)) {
+			$this->hasTrashBackend = true;
+			$this->expiration = \OC::$server->get(Expiration::class);
+			$this->trashBackend = \OC::$server->get(PageTrashBackend::class);
```
juliushaertl, line 29:
```suggestion
			$this->expiration = \OCP\Server::get(Expiration::class);
			$this->trashBackend = \OCP\Server::get(PageTrashBackend::class);
```

juliushaertl (approved):
Small comment, otherwise fine 👍

lib/BackgroundJob/ExpirePageTrash.php
mejo-: Thanks, changed now.

lib/BackgroundJob/ExpirePageTrash.php
mejo-: Uh, that's great to learn. Not for now, but one day we'll drop support for everything below 28 :blush:

PR #1037, opened 2023-12-05
feat(search): Always search for page content, not only in app realm

mejo-:


lib/Search/PageContentProvider.php
```diff
@@ -95,13 +95,7 @@ public function getOrder(string $route, array $routeParameters): int {
...
 	 */
 	public function search(IUser $user, ISearchQuery $query): SearchResult {
-		// Only search for page content if the app is active
-		if ($this->appManager->isEnabledForUser('circles', $user) &&
-			strpos($query->getRoute(), 'collectives.') === 0) {
-			$collectiveInfos = $this->collectiveHelper->getCollectivesForUser($user->getUID(), false, false);
-		} else {
-			$collectiveInfos = [];
-		}
+		$collectiveInfos = $this->collectiveHelper->getCollectivesForUser($user->getUID(), false, false);
```
juliushaertl, line 98: Do we still need the circles check or is that gracefully handled in getCollectivesForUser if circles is disabled?

juliushaertl (commented):
Small question inline

We could actually also think about merging the two providers to only have one which displays matches on the page title first? But might be a bigger topic to work properly with pagination, so just consider this a future suggestion.

mejo-:
> We could actually also think about merging the two providers to only have one which displays matches on the page title first? But might be a bigger topic to work properly with pagination, so just consider this a future suggestion.

Yeah, I'd say let's discuss this separately.

lib/Search/PageContentProvider.php
mejo-: Nope, good catch. The check for circles is still required.

juliushaertl (approved):

Issue #633, opened 2023-04-06
Allow multiple share links with different edit rights


hansimausisteffiwurstbrot:
**Is your feature request related to a problem? Please describe.**
It might be useful to have two public links. One with the rights to edit, the other without.

**Describe the solution you'd like**
Allow the creation of multiple links.

**Describe alternatives you've considered**
I don't see any relevant.

Thank you very much for the good app!

Issue #515, opened 2023-01-23
Add support for public shares of pages


42ske:
**Is your feature request related to a problem? Please describe.**
If you use collectives as cooperative knowledge base of your group/organization it’s a matter of time when the first member wants to share contents with non-members. This is easily possible at the level of collectives. For pages you can manage this a bit complicated via the files-app. 

**Describe the solution you'd like**
I want to be able to:
1. create a public link for pages according to the function of public links for entire collectives.
2. There should be an option if this link should also include all supages of a page or not.
3. The already existing sharing setting in the collectives’ settings should have an effect on the page links of course.

**Describe alternatives you've considered**
A detour is to share page via the files app, which is not very intuitive.

**Additional context**

max-nextcloud:
Is there really a need for being able to share a page without its subpages?
I think it would be fine to only have a way to share both together. One could still fallback to sharing the single page file via files.

> The already existing sharing setting in the collectives’ settings should have an effect on the page links of course.
The only option i can think of right now is read-only vs. editable shares. I think they should be independent so i can share the entire collective as read only and a page with its subpages as editable - or the other way around. These links may go to different audiences anyway.

42ske:
> Is there really a need for being able to share a page without its subpages? I think it would be fine to only have a way to share both together. One could still fallback to sharing the single page file via files.

Being able to share a page with its subpages would be a very useful feature of course. There would be lot of scenarios where this would be enough. But I can imagine, there are also other scenarios in which a seperate possibility would be great. 

And yes you are right that you can use the files app for that, but I know for lot of users this ist not very intuitive (I handle it this way).

> > 3. The already existing sharing setting in the collectives’ settings should have an effect on the page links of course.
> 
> The only option i can think of right now is read-only vs. editable shares. I think they should be independent so i can share the entire collective as read only and a page with its subpages as editable - or the other way around. These links may go to different audiences anyway.

Yes, you are absolutely correct with this.

hansimausisteffiwurstbrot:
to whom it may concern: If we can support this feature with some financial means let us know. :)

max-nextcloud:
Ideas for implementation:
* provide a share tab in the page sidebar for sharing that particular page and its subpages.
* extend the `collectives_shares` database.

PR #1041, opened 2023-12-06
Public page shares

mejo-:
### 📝 Summary

* Allow to share pages as links (along with their subpages)
* Add page sidebar "share" tab to manage link shares
* Move management of collective share links into page sidebar of landingpage
* Resolves: #515
* Resolves: #633

#### 🖼️ Screenshots

🏚️ Collective actions before | 🏡 Collective actions after
---|---
grafik | grafik

'Share' sidebar tab | Configure share permissions
--- | ---
grafik | grafik

| Screencast (slightly outdated) |
| --- |
| recording1 |

### 🚧 TODO

* [x] backend:
  * [x] share parent folder if pageId is index page
  * [x] implement `checkPagePermissions()` in PublicPageController
  * [x] implement `findChildren() in PageService
  * [x] add support for single-page share
  * [x] check consequences of adding subpage in single-page shares
  * [x] allow to create several shares
  * [x] make sure move/copy page cannot break out of share realm
* [x] frontend:
  * [x] add share tab in page sidebar
  * [x] add options to edit share permissions
  * [x] move share collective to landingpage page sidebar
  * [x] add link pointing to sidebar to collective actions
* [x] UI adjustments
  * [x] don't show trash
  * [x] hide 'delete page' options in UI
  * [x] adjust move/copy page picker
* [x] make sure mountpoint works
* [x] behat tests:
  * [x] add/get/delete single-page share
  * [x] fix attachment check
  * [x] add/get/delete page folder share
  * [x] fail to trash page in page share
  * [x] fail to get/create/touch/move/setEmoji/setSubpageOrder/getAttachments/getBacklinks page outside page share
* [x] cypress tests:
  * [x] create read-only collective share, open (read-only)
  * [x] update collective share to editable, open, edit
  * [x] delete collective share
  * [x] create read-only page share, open (read-only)
  * [x] update page share to editable
  * [x] delete page share

### 🏁 Checklist


max-nextcloud:
> What do you think @max-nextcloud?

@mejo- sounds like a good approach to me.

>  (e.g. attachments then always reside in the page folder, not in the parent page folder).

I thought they'd always live in a `.attachments.${fileId}` folder. But i guess that folder is either in the parent folder or in the same folder as the file itself (`Readme.md`)

mejo-:
> I thought they'd always live in a `.attachments.${fileId}` folder. But i guess that folder is either in the parent folder or in the same folder as the file itself (`Readme.md`)

Exactly :wink:

nimishavijay:
Super nice improvement! It looks great, only 2 small changes: 
- [ ] use a different icon for the copy button. My suggestion is the [copy MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%3Acontent_copy%3A). 
- [ ] unsure about wording "Manage share links" maybe just "Share" works? 

This is maybe a bigger discussion, but I wonder if it makes sense to combine the "Manage members" and "Manage sharing" to some degree. Both of those things are regarding who has access to the collective/page. This could be as simple as showing the members of the collective in the sharing tab as a "People with access" section (similar to the "Others with access" in the Files sharing) where there is also a button to "Manage members". But that's definitely a follow-up topic :)

lib/Service/CollectiveShareService.php
```diff
@@ -156,32 +197,46 @@ private function isShareEditable(IShare $folderShare): bool {
...
+		$pageId = 0;
+		$nodeId = 0;
+		if ($pageInfo) {
+			$pageId = $pageInfo->getId();
+			$file = $this->pageService->getPageFile($collective->getId(), $pageId, $userId);
+			if (NodeHelper::isIndexPage($file)) {
+				$nodeId = $file->getParent()->getId();
+			} else {
+				$nodeId = $pageInfo->getId();
+			}
```
max-nextcloud, line 220: I might be confusing things here. I thought you intended to turn normal `page.md` pages into a `Readme.md` within their own folder so you can share the folder and thus all future subpages.

This code looks more like you're handling the pages without pages just like any other file.

lib/Service/CollectiveShareService.php
```diff
@@ -156,32 +197,46 @@ private function isShareEditable(IShare $folderShare): bool {
...
+		}
+
+		if (null !== $this->findShare($userId, $collective->getId(), $nodeId)) {
 			throw new NotPermittedException($this->l10n->t('A share for collective %s exists already', $collective->getName()));
```
max-nextcloud, line 224: > A share for collective %s exists already

or the particular page in question, right?

max-nextcloud (commented):
Only reviewed the backend and cypress tests thus far.

Looks great. :100: 

I only added a few minor comments. I'll continue later on with the frontend code.

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -0,0 +1,57 @@
...
+			<template #icon>
+				<NcLoadingIcon />
+			</template>
+		</NcEmptyContent>
+
+		<!-- shares list -->
+		<ul v-else class="sharing-list">
+			<SharingEntryLink v-if="!shares.length" />
+			<SharingEntryLink v-for="(share, index) in shares"
+				v-else
```
max-nextcloud, line 14: I think you can drop this. if shares is empty the `v-for` will not do anything anyway.

src/store/pages.js
```diff
@@ -93,23 +93,23 @@ export default {
 			return `/${pagePath}?fileId=${id}`
 		},
 
-		pagePathTitle: (_state, getters) => (page) => {
+		pagePathTitle: (_state) => (page) => {
```
max-nextcloud, line 96: If you're neither using state nor getters you can leave out the parameters entirely.

```suggestion
		pagePathTitle: () => (page) => {
```

tests/Integration/features/bootstrap/FeatureContext.php
```diff
@@ -964,21 +1040,39 @@ public function userDeletesPublicShare(string $user, string $collective): void {
...
 	 *
+	 * @param string $page
 	 * @param string $collective
+	 * @param string $owner
 	 *
 	 * @throws GuzzleException
 	 */
-	public function anonymousFailsToSeePublicCollective(string $collective): void {
+	public function anonymousSeesPublicPageShare(string $page, string $collective, string $owner): void {
+		$this->setCurrentUser($owner);
```
max-nextcloud, line 1060: I don't quite understand what `setCurrentUser` does. But my intuition says for testing anonymous behavior it should not be set to the owner.

Is the current user only needed for the preparation and `sendRequest` then does not use it anymore?

tests/Integration/features/pages.feature
```diff
@@ -100,16 +100,6 @@ Feature: pages
...
-  Scenario: Trash and delete all subpages reverts subfolders
-    When user "jane" trashes page "subpage" in "BehatPagesCollective"
-    And user "jane" trashes page "subpage2" in "BehatPagesCollective"
-    And user "jane" trashes page "subpage3" in "BehatPagesCollective"
-    When user "jane" deletes page "subpage" from trash in "BehatPagesCollective"
-    When user "jane" deletes page "subpage2" from trash in "BehatPagesCollective"
-    When user "jane" deletes page "subpage3" from trash in "BehatPagesCollective"
-    Then user "jane" doesn't see pagePath "parentpage/Readme.md" in "BehatPagesCollective"
-    And user "jane" sees pagePath "parentpage.md" in "BehatPagesCollective"
-
```
max-nextcloud, line 112: Maybe keep this and update to the current behavior?

tests/Integration/features/publicPageShare.feature
```diff
@@ -0,0 +1,65 @@
+Feature: publicPageShare
+
+  Scenario: Create and share a single page publicly (read-only)
+    When user "jane" creates collective "BehatPublicPageCollective"
+    And user "jane" creates page "singlesharepage" with parentPath "Readme.md" in "BehatPublicPageCollective"
+    And user "jane" creates public page share for page "singlesharepage" in "BehatPublicPageCollective"
+    Then anonymous sees public page share "singlesharepage" in collective "BehatPublicPageCollective" with owner "jane"
```
max-nextcloud, line 7: Maybe check if anonymous cannot see the top level "Readme.md"?

max-nextcloud (approved):
Looks good. :tada: 

Had a look at the remaining pieces and added some minor comments. No blockers though.

appinfo/info.xml
```diff
@@ -22,7 +22,7 @@ In your Nextcloud instance, simply navigate to **»Apps«**, find the
 **»Circles«** and **»Collectives«** apps and enable them.
 
 	]]></description>
-	<version>2.9.2</version>
+	<version>2.10.0</version>
```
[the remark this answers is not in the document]
mejo-: Needed to bump the version due to the DB migration step. Otherwise the migration would not be applied.

lib/Controller/PublicPageController.php
```diff
@@ -99,6 +100,44 @@ private function checkEditPermissions(): void {
 		}
 	}
 
+	/**
+	 * @param int    $collectiveId
+	 * @param int    $pageId
+	 * @param int    $id
```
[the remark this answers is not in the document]
mejo-: Good point. I renamed `pageId` to `sharePageId` now.

lib/Service/CollectiveShareService.php
mejo-: Oh, good catch. This if condition can go. We already make sure that the `pageInfo` passed to `createShare` isn't a single page file anymore [here](https://github.com/nextcloud/collectives/blob/388224b3e9815653747995a226313dea677b7c8f/lib/Controller/ShareController.php#L126).

lib/Service/CollectiveShareService.php
mejo-: Same here. We want to support several shares now, so let's just get rid of this check.

src/components/Page/PagePicker.vue
```diff
@@ -5,7 +5,7 @@
 				{{ t('collectives', 'Copy or move page') }}
 			</h2>
 			<span class="crumbs">
-				<div class="crumbs-home">
+				<div v-if="!selectedCollective || !selectedCollective.isPageShare" class="crumbs-home">
 					<NcButton type="tertiary"
 						:aria-label="t('collectives', 'Breadcrumb for list of collectives')"
 						:disabled="!selectedCollective"
```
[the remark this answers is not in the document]
mejo-: Well, not exactly. The "list of collectives" breadcrumb is displayed if either `!selectedCollective` (no collective selected) or `!selectedCollective.isPageShare` (a collective is selected and this is not a page share). In other words, I only want to display it always, except for page shares. Unfortunately, `!selectedCollective.isPageShare` is not enough here, because `selectedCollective` might be null and in this case I also want to display the button.

The `disabled` condition on the other hand is different: the button should be disabled if no collective is selected.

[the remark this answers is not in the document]
mejo-: I agree though that the code is not exactly intuitive :zany_face:

src/store/pages.js
mejo-: Indeed, missed that one!

src/components/PageSidebar/SidebarTabSharing.vue
mejo-: I see. Still, isn't it better to make explicit in the code that either the first, or the second is active?

tests/Integration/features/publicPageShare.feature
mejo-: I'll add a check the anonymous cannot see `singlesharepage.md` in the share.

tests/Integration/features/pages.feature
mejo-: Well, I don't think this is worth the computing time. It would only check the default behaviour that is already checked by other tests. This test was only to check if reverting the subfolder structure happened on page deletes.

tests/Integration/features/bootstrap/FeatureContext.php
mejo-: Exactly. `setCurrentUser` is needed to retrieve the `pageId`, as we don't know that one. The request itself doesn't use it, because `sendRequest` get's `auth: false` passed.

cypress/e2e/collective-share.spec.js
```diff
@@ -123,20 +129,110 @@ describe('Collective Share', function() {
...
+			cy.wait('@createShare')
+			cy.get('.toast-success').should('contain', 'Page "Sharepage" has been shared')
+			cy.get('.sharing-entry .share-select')
+				.should('contain', 'View only')
+			cy.get('button.sharing-entry__copy')
+				.click()
+			cy.get('.toast-success').should('contain', 'Link copied')
+			cy.get('@clipBoardWriteText').should('have.been.calledOnce')
+		})
+		it('Allows opening a shared (non-editable) page', function() {
```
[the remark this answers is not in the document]
mejo-: I see your point. Let's move it into one big `it` block for now. The sole reason why I didn't do this is to logically separate the different things that get tested there. But I can do so with code comments as well.

cypress/e2e/collective-share.spec.js
```diff
@@ -123,20 +129,110 @@ describe('Collective Share', function() {
...
+			cy.get('.toast-success').should('contain', 'Collective "Share me" has been unshared')
 		})
 		it('Opening unshared collective fails', function() {
 			cy.logout()
 			cy.visit(shareUrl, { failOnStatusCode: false })
 			cy.get('.body-login-container').contains(/(File|Page) not found/)
 		})
 	})
+
+	describe('page share', function() {
```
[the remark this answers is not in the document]
mejo-: Agreed.

cypress/e2e/collective-share.spec.js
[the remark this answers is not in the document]
mejo-: Looking into this again, I decided to not do it for now. All in one big `it` block doesn't look right to me. Probably you're right and we should implement `cy.shareCollective` and `cy.sharePage` instead - using the API. We would have to do this refactoring for the collective share tests too. I think it's out of scope for this PR.

src/components/Page/PagePicker.vue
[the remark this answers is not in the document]
max-nextcloud: I think i just confused the `||` for an `&&`

src/components/PageSidebar/SidebarTabSharing.vue
max-nextcloud: :shrug: I don't know and trust your choice.

punkyard:
hi,
and could you add shares for each page? both public and internal shares in drop down menu, please
happy new year!

gymnae:
Second that question: The mockups in OP show a nice single page sharing dialogue, but it appears absent in the version shipped with NC 28

PR #1065, opened 2023-12-29
Set default language for TNTSearch

Koc:
### 📝 Summary
This change improves morphology in search functionality by providing correct Stemmer for TNTSearch

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image  | image |

### 🏁 Checklist

### DIFF
--- a/lib/Search/FileSearch/FileSearcher.php
+++ b/lib/Search/FileSearch/FileSearcher.php
@@ -21,13 +21,31 @@ class FileSearcher extends TNTSearch {
 		'storage' => ''
 	];
 
-	protected FileIndexer $indexer;
+	public const SUPPORTED_LANGUAGES = [
+		'ar' => 'Arabic',
+		'cr' => 'Croatian',
+		'fr' => 'French',
+		'de' => 'German',
+		'en' => 'Porter',
+		'it' => 'Italian',
+		'lv' => 'Latvian',
+		'pl' => 'Polish',
+		'pt' => 'Portuguese',
+		'ru' => 'Russian',
+		'uk' => 'Ukrainian',
+	];
 
-	public function __construct() {
+	private const UNSUPPORTED_LANGUAGE = 'No';
+
+	protected FileIndexer $indexer;
+	private ?string $language;
+
+	public function __construct(?string $language = null) {
 		parent::__construct();
 		$this->loadConfig();
 		$this->asYouType(true);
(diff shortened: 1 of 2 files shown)


Koc:
@mejo- @juliushaertl can you review?

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -21,13 +21,29 @@ class FileSearcher extends TNTSearch {
...
+		'pt' => 'Portuguese',
+		'ru' => 'Russian',
+		'uk' => 'Ukrainian',
+	];
+
 	protected FileIndexer $indexer;
+	private string $language;
 
-	public function __construct() {
+	public function __construct(string $language = 'no') {
```
juliushaertl, line 41:
```suggestion
	public function __construct(?string $language = null) {
```

Would prefer this to not be confused with no for Norwegian

juliushaertl (commented):
Looks good to me with one comment regarding the default value

lib/Search/FileSearch/FileSearcher.php
Koc: fixed, but we should to bypass `'No'` explicitly to `\TeamTNT\TNTSearch\Indexer\TNTIndexer::setLanguage`

Koc: @juliushaertl can you re-check/approve? :pray:

juliushaertl (approved):

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -21,13 +21,31 @@ class FileSearcher extends TNTSearch {
...
+		'en' => 'Porter',
+		'it' => 'Italian',
+		'lv' => 'Latvian',
+		'pl' => 'Polish',
+		'pt' => 'Portuguese',
+		'ru' => 'Russian',
+		'uk' => 'Ukrainian',
+	];
+
+	private const UNSUPPORTED_LANGUAGE = 'No';
```
mejo-, line 38: Shouldn't this be some string that is less likely to be similar to `no` (for "Norwegian")?

mejo-: Ok, thanks for the pointer. Then the PR is good to go :blush: Thanks for the contribution @Koc!

PR #1078, opened 2024-01-15
fix: Do not load editor on unrelated pages

juliushaertl:
Signed-off-by: Julius Härtl <jus@bitgrid.net>

### 📝 Summary

Make sure that unrelated pages do not load the editor as we don't need it there.

To reproduce open an unrelated page and see editor scripts as well as the reference provider scripts loaded though not needed e.g. https://nextcloud.local/index.php/s/ for a 404 page

### 🏁 Checklist

### DIFF
--- a/lib/Listeners/BeforeTemplateRenderedListener.php
+++ b/lib/Listeners/BeforeTemplateRenderedListener.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 
 namespace OCA\Collectives\Listeners;
 
+use OCA\Collectives\AppInfo\Application;
 use OCA\Collectives\Fs\UserFolderHelper;
 use OCA\Collectives\Service\NotPermittedException;
 use OCA\Text\Event\LoadEditor;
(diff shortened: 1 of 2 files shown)


max-nextcloud (approved):
Code changes make sense to me.
There's a psalm error in line 67 now for missing the `LoadEditor` class. I suspect that this was prevented by line 66 before and now psalm does not parse the more complex if condition anymore - but that's just a gut feeling. Maybe this can be fixed by two nested  if statements or an early return rather than one combined if statement.

juliushaertl:
Added the LoadEditor event to the psalm stub, which seems most sensible to me considering not having time right now to dive into reasons for the psalm failure

lib/Listeners/BeforeTemplateRenderedListener.php
```diff
@@ -41,7 +42,7 @@ public function __construct(IUserSession $userSession,
 	public function handle(Event $event): void {
 		if (!($event instanceof BeforeTemplateRenderedEvent)) {
 			return;
-		}
+		};
```
mejo-, line 45: This looks a bit weird to me :thinking: 

Leftover/typo?

mejo- (changes requested):

lib/Listeners/BeforeTemplateRenderedListener.php
juliushaertl: O I misread the diff when adding, thought i had removed that 🙈 Let me drop that

juliushaertl: Fixed

mejo- (approved):

PR #1126, opened 2024-02-22
Fix link handling with new link bubble from Nextcloud 29

mejo-:
### 📝 Summary

Starting with Nextcloud 29, clicking on a link opens the link bubble. This bubble contains a preview, clicking on it opens the link. It's no longer possible to pass a custom link handler into the Text app. Instead, all custom link handling needs to happen in the custom reference providers that provide the link preview.
 
Also starting with Nextcloud 29, all links to the collectives app open in the same tab, regardless whether in edit or view mode.

### 🏁 Checklist


juliushaertl (approved):

juliushaertl:
@mejo- I think the failure is expected on the upgrade test and will be fine after merge

PR #1130, opened 2024-02-27
feat: Implement a team resource provider

juliushaertl:
Requires https://github.com/nextcloud/server/pull/43855

### DIFF
--- a/lib/Team/CollectiveTeamResourceProvider.php
+++ b/lib/Team/CollectiveTeamResourceProvider.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OCA\Collectives\Team;
+
+use OCA\Collectives\AppInfo\Application;
+use OCA\Collectives\Db\CollectiveMapper;
+use OCA\Collectives\Service\CollectiveService;
+use OCP\IURLGenerator;
+use OCP\Teams\ITeamResourceProvider;
+use OCP\Teams\TeamResource;
+
+class CollectiveTeamResourceProvider implements ITeamResourceProvider {
+	public function __construct(private CollectiveMapper $collectiveMapper, private CollectiveService $collectiveService, private IURLGenerator $urlGenerator, private ?string $userId) {
+
+	}
+
+	public function getId(): string {
+		return Application::APP_NAME;
+	}
+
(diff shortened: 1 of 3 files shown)


juliushaertl:
@mejo- I rebased this and updated the psalm baseline as we still test against 25 there

lib/Team/CollectiveTeamResourceProvider.php
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	public function getIconSvg(): string {
+		return file_get_contents(__DIR__ . '/../../img/collectives.svg');
+	}
+
+	public function getSharedWith(string $teamId): array {
+		$collective = $this->collectiveMapper->findByCircleId($teamId);
+		if ($collective) {
+			$collective = $this->collectiveService->getCollectiveInfo($collective->getId(), $this->userId);
```
mejo-, line 34: This might throw `NotFoundException` if the user doesn't have access to the collective. Should we handle this gracefully, or it it ok to throw? When I read the server code correctly, at least here it would be a problem if one provider throws:

https://github.com/nextcloud/server/blob/7033967115ae7f5e5be8551f0530143d5b1a7885/lib/private/Teams/TeamManager.php#L80C18-L89C1

lib/Team/CollectiveTeamResourceProvider.php
```diff
@@ -0,0 +1,58 @@
...
+		}
+		return [];
+	}
+
+	public function isSharedWithTeam(string $teamId, string $resourceId): bool {
+		// TODO: Implement isSharedWithTeam() method.
+	}
+
+	public function getTeamsForResource(string $resourceId): array {
+		$collective = $this->collectiveMapper->findByIdAndUser((int)$resourceId, $this->userId);
```
mejo-, line 55: Theoretically, `$this-userid` could be null, which would return all collectives. I think it's better to use `$this->collectiveService->getCollective((int)$resourceId, $this->userId)` as this is guaranteed to only accept a non-null `userId`.

Conceptionally, `CollectiveMapper` isn't guaranteed to do permission checks, while `CollectiveService` should always do so. At least in theory :see_no_evil:

mejo- (changes requested):

Issue #1114, opened 2024-02-16
Heading anchors point to `#undefined` in view mode


mejo-:
**Describe the bug**
Heading anchors point to `#undefined` in view mode.

**To Reproduce**
1. Create a document with headings
2. Switch to view mode
3. Hover over the `#` symbol left to a heading and see the URL.

**Server details:**
- Collectives app version: latest
- Nextcloud version: 27.1.5

Koc:
not reproducible for me in current main branch after running `npm ci && npm run watch`

AngryNui:
I have the same problem. Only in view mode anchor points to #undefined. In edit mode the anchor points to the correct position. 
When view-mode is set as default and you open a link with the correct anchor it wont show the correct position. 

Collectives app version: 2.10.0
Nextcloud: 28.0.3

grnd-alt:
#### Findings
I looked into this issue, couldn't quite figure it out but still possibly made some findings.
The Links for headings are created in extractor.js
and called by either onUpdate or onCreate in nodes/Heading/index.js

the links rendering as undefined most likely has something to do with the lifecycle hooks not beeing called in the way they are ment to be. Somehow the setupReader and setupEditor functions in collectives that start those editors make it so the reader and editor influence one another.

Inspecting the stack trace with breakpoints it was visible that addNode in nodes/Heading/index.js was called before the onCreated Hook which could possibly cause issues. Another possibility would be that the influence of different orders of setContent-calls in collectives affects the link-values, which would bring up the question why setContent calls don't all trigger the same update cycle.

I don't know if any of that will be helpful but I did my best.

mejo-:
This finally got fixed in Text with nextcloud/text#5893.

PR #1150, opened 2024-03-09
Fix anchcors scrolling

Koc:
### 📝 Summary
Steps to reproduce:
- Create long page with multiple headings
- Click on heading in the bottom of the page
- Copy url from address bar with anchor and open it in new tab

Expected result:
- Content scrolled to heading

Actual result:
- There are no scroll

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 TODO
- [x] For some reason `onLoaded` callback is not called in my case, so I'm using `setTimeout`. Any idea how to fix that?
- [x] ~~Merge https://github.com/nextcloud-libraries/nextcloud-vue/pull/5435~~

### 🏁 Checklist

### DIFF
--- a/src/mixins/editorMixin.js
+++ b/src/mixins/editorMixin.js
@@ -86,7 +86,15 @@ export default {
 				onOutlineToggle: (visible) => {
 					this.toggleOutlineFromEditor(visible)
 				},
+				onLoaded: () => {
+					if (document.location.hash) {
+						// scroll to the corresponding header if the page was loaded with a hash
+						const element = document.querySelector(`[href="${document.location.hash}"]`)
+						element?.click()
+					}
+				},
 			})
+
 			if (!this.loading('pageContent')) {
 				this.reader.setContent(this.pageContent)
 			}
--- a/src/Collectives.vue
+++ b/src/Collectives.vue
@@ -183,4 +183,10 @@ export default {
 .app-navigation .app-navigation-toggle {
 	top: 0 !important;
 }
+
+/* Fix for anchor scroll and sticky header */
+.splitpanes__pane-details {
+	scroll-padding-top: 60px;
+	scroll-behavior: smooth;
+}
 </style>


mejo-:
> For some reason onLoaded callback is not called in my case, so I'm using setTimeout. Any idea how to fix that?

Currently only the session editor ([`Editor.vue` in the text app](https://github.com/nextcloud/text/blob/main/src/components/Editor.vue)) supports the `onLoaded` callback. But you're right, the editor used for the view mode ([`MarkdownContentEditor` in the text app](https://github.com/nextcloud/text/blob/main/src/components/Editor/MarkdownContentEditor.vue)) should do so as well. I opened https://github.com/nextcloud/text/pull/5489 to implement this.

src/mixins/editorMixin.js
```diff
@@ -86,10 +86,34 @@ export default {
 				onOutlineToggle: (visible) => {
 					this.toggleOutlineFromEditor(visible)
 				},
+				onLoaded: () => {
+					console.log('onLoaded')
```
mejo-, line 90: With https://github.com/nextcloud/text/pull/5489 this will work. Definitely cleaner than the `setTimeout()` approach :blush:

src/mixins/editorMixin.js
```diff
@@ -86,10 +86,34 @@ export default {
...
+				},
 			})
+
+			if (document.location.hash) {
+				window.setTimeout(() => {
+					const element = document.querySelector(document.location.hash)
+					const topPos = element.offsetTop
+
+					getScrollParent(element)?.scrollTo({
```
mejo-, line 99: Any reason why you don't just do `element?.scrollIntoView({ behavior: 'smooth' })` directly? I don't understand why you would want to get the parent node if `scrollHeight` is larger than `clientHeight` :thinking:

mejo- (changes requested):

src/mixins/editorMixin.js
Koc: I'm going to double check it one more time. But as far I can remember - there was an issue with sticky header overlapping

Koc: Just checked again

`element?.scrollIntoView({ behavior: 'smooth' })`:
image

Seems like we need to play with `scroll-margin-top` or something like that

Koc: Soo, I've investigated it little bit.

We need add few lines to [NcAppComponent](https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcAppContent/NcAppContent.vue#L432)
```css
  overflow-y: auto;
  scroll-padding-top: 60px;
  scroll-behavior: smooth;
```
This needed to fix scrolling when you clicking on anchors.

But it not helps to `scrollIntoView` :disappointed: . It scrolls more than needed, take a look on video

https://github.com/nextcloud/collectives/assets/191082/1f2ad586-2ad7-4b0e-9abf-f0537bdbf05f

src/mixins/editorMixin.js
Koc: It is working now! :tada:

Koc:
@mejo- I've updated current PR, and opened one more to nextcloud-vue. Please review one more time.

mejo- (changes requested):
Dear @Koc, great to hear that `onLoaded` works now :blush:

Regarding your PR against nextcloud-vue, I think that we should rather change the behaviour here in Collectives. It's not given that the contents given to `NcAppContent` contain a heading like in collectives. Adding something like the following to `Collectives.vue` should do the trick:

```scss
:deep(.splitpanes__pane-details) {
	scroll-padding-top: 60px;
	scroll-behavior: smooth;
}
```

Koc:
@mejo- thank you for suggestion!
I've pushed changes and closed related PR. The only one thing - I've omit `:deep` to be consistent with other styles. It works fine without it, as non-scopped styles used

mejo- (approved):
Awesome, works like a charm now. Thanks a lot for the contribution @Koc!

Only #1114 sometimes breaks the behaviour, but that's a separate issue.

Issue #1090, opened 2024-01-22
Page filter performance issues


mejo-:
**Describe the bug**
In a collective with many pages, the page filter in the page list has severe performance issues. For some clients, it freezes the browser for several seconds. Reported with Firefox 121.0 on Linux and with Safari on recent iOS.

juliushaertl:
Did a quick check and it seems mostly time spend on rendering the navigation components when typing, likely because we render all matched ones in sub categories once we start typing in the search.

Not sure what the ideal solution could be, but a quick fix would be to add some debounce, so we at least only filter once when typing a word

marcoambrosini:
The most important thing from a user's perspective is that the input field is frozen when the list is loading. It seems that there's some synchronous logic somewhere tieing the two together. It would be great if it could be decoupled.

PR #1155, opened 2024-03-13
Fix/page filter performance issues

grnd-alt:
### 📝 Summary

* Resolves: #1090 

Logic for filtering Search results in collectives was done in the search results' component itself mainly, this was extracted to be done at once in the parent component. To Improve performance even more and make the search responsive a virtual-scroller was added to only render visible search results.

⚠️ Filtered Results are now in a flat list. That improves performance, and also makes the list more intuitive, as having indented search results without the parent page can be confusing.

🏚️ Before | 🏡 After
---|---
Entering something in the search bar was very unresponsive, sometimes waiting multiple seconds until the first keystroke is handled | Entering something is very responsive, waiting very little to virtually no time until keystrokes are recognized

### 🚧 TODO
While developing, a bug was discovered. The subpages list is actually bigger than the content it displays (search bar is written above it).
Therefore when the virtual scroller comes in, there are two scrollbars. It's planned to work on this in a seperate bug ticket as it's not completely related/caused by the performance improvement.

### 🏁 Checklist

image

### DIFF
--- a/src/components/PageList.vue
+++ b/src/components/PageList.vue
@@ -99,13 +99,24 @@
 					:level="1"
 					:filter-string="filterString"
 					:is-template="true" />
-				<SubpageList v-for="page in subpages"
-					:key="page.id"
-					:data-page-id="page.id"
-					:page="page"
-					:level="1"
-					:filter-string="filterString"
-					class="page-list-drag-item" />
+				<div v-if="filterString !== ''">
+					<SubpageList v-for="page in filteredPages"
+						:key="page.id"
+						:data-page-id="page.id"
+						:page="page"
+						:level="1"
+						:filtered-view="true"
+						class="page-list-drag-item" />
+				</div>
+				<div v-else>
+					<SubpageList v-for="page in subpages"
+						:key="page.id"
+						:data-page-id="page.id"
+						:page="page"
+						:level="1"
+						:filtered-view="false"
+						class="page-list-drag-item" />
+				</div>
 			</Draggable>
 		</div>
 		<PageTrash v-if="displayTrash" />
@@ -174,7 +185,18 @@ export default {
 			'sortBy',
 			'showing',
 			'showTemplates',
+			'allPages',
 		]),
+
+		allpages() {
+			return this.allPages(this.rootPage.collectivePath)
+		},
+
+		filteredPages() {
+			return this.allpages.filter(p => {
+				return p.title.toLowerCase().includes(this.filterString.toLowerCase())
+			})
+		},
 
 		subpages() {
 			if (this.rootPage) {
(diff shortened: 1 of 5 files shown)


src/components/PageList.vue
```diff
@@ -174,8 +194,19 @@ export default {
 			'sortBy',
 			'showing',
 			'showTemplates',
+			'allPages',
 		]),
 
+		allpages() {
```
juliushaertl, line 200: Some different wording would be nice to make it easier to differentiate between this and the vuex store getter

src/components/PageList.vue
```diff
@@ -174,8 +194,19 @@ export default {
...
+			return this.allPages(this.rootPage.collectivePath)
+		},
+
+		filteredPages() {
```
juliushaertl, line 204: Could this just use the store getter?

juliushaertl: Ah no, I guess it needs to be based on the collective path, but then the store one is unused

src/components/PageList.vue
```diff
@@ -253,6 +284,11 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+
+.scroller{
+	height:100vh;
```
juliushaertl, line 289:
```suggestion
	height: 100vh;
```

mejo- (changes requested):
Nice one :tada: I didn't test yet, but code changes look sensible.

As discussed yesterday, pages would need to get sorted before they're filtered. I'd add a getter `sortedPages` similar to `sortedSubpages` to the page store that returns all pages sorted by the current sort order. Then you can apply the filtering on this getter.

mejo-:
@grnd-alt test ping

src/components/PageList.vue
```diff
@@ -97,15 +97,31 @@
...
+						key-field="id">
+						<SubpageList :key="item.id"
+							:data-page-id="item.id"
+							:page="item"
+							:level="1"
+							:filtered-view="true"
+							class="page-list-drag-item" />
+					</RecycleScroller>
+				</div>
+				<div v-if="filterString.length === 0">
```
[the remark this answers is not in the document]
grnd-alt: we can't use v-if and v-for on the same element as the list would still be rendered, though it's not shown.
here's the linting rule in place for that: [eslint-rule](https://eslint.vuejs.org/rules/no-use-v-if-with-v-for.html)

src/components/PageList.vue
grnd-alt: the unused store getter is removed.

src/components/PageList.vue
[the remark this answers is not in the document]
mejo-: Ah, makes sense :+1:

mejo- (approved):
Really nice work :tada:

Issue #1068, opened 2024-01-09
attachments (pictures/images) in subsites not loading in print-view / export-view


nordxONE:
**Describe the bug**
images of sub-sub-sites are not referenced correctly in the print-view / export-view.

the problem sometimes occures also in the first sub-site of the collective, but sometimes works.

**To Reproduce**
when you add a new collective and add a subsite and to this subsite another subsite, than add a picture to it, it wont show you the picture in the print-view /export-view cause the path of the file is wrong (it references to the main-folder - the folder main-folder of the collective - and "thinks" there is the attachement folder, but in reality it is in the subdirectory which is created when you add a subsite).
however it will show up correctly in the normal-view of the collective itself (where you can view or edit the collectives).

**Expected behavior**
it should show every image in the print-view as it is in the normal view/editor view.

**Screenshots**

When exporting the problematic collectiv it shows the following (not loading any image at all its always 0 out of all images) for a short period:
![image](https://github.com/nextcloud/collectives/assets/50635072/00f7bade-cac6-435a-b028-f19338f455c3)

after loading of the print-view it looks like this:
![image](https://github.com/nextcloud/collectives/assets/50635072/1b6d5303-0af0-42ba-97cc-c538643b6068)

**Server details:**
- Collectives app version: 2.9.2
- Nextcloud version: 27.1.5
- PHP Version: 8.1.26
- Database: MariaDB 10.5.21

**Client details:**
 - OS: Windows
 - Browser: Chrome / Firefox
 - Browser version: 120.0.6099.200 / 121.0
 - Device: Desktop

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
No specific error for this
```

#### Browser log
```
i get an 404 File not found. It wants to load the file from the main-directory (collective directory) and not the subfolder where the 
```

</details>

NoobyNiceDev:
Problem still persists in 
Nextcloud-Version 28.0.2 
Collectives: 2.9.2
PHP: 8.1.21
DB: MySQL 10.5.21

Aveyron-RetD:
I confirm the bug. In order to not display the bug I let collectives in edit mode.

PR #1157, opened 2024-03-14
fix(editor): Pass either `page` or `currentpage` to `createEditor`

mejo-:
E.g. in `PagePrint` we don't load the editor for `currentPage` but for other pages. Make sure we pass the right properties in this case.

Fixes image + attachment loading in print view.

Fixes: #1068

### 🏁 Checklist


src/mixins/editorMixin.js
```diff
@@ -42,6 +42,15 @@ export default {
...
 			return !!this.pageContent || !this.loading('pageContent')
 		},
+
+		/**
+		 * Use `this.page` if available (e.g. in `PagePrint`) and fallback to `currentPage`
+		 */
+		pageToUse() {
+			return this.page
+				? this.page
+				: this.currentPage
```
grnd-alt, line 52: might be optimized at build time anyway, but could it not just be like this?
```
this.page || this.currentPage
```

grnd-alt (commented):
looks good. Tested locally and working
one small possible improvement.
Could maybe also think about adding cypress tests for this behavior

src/mixins/editorMixin.js
mejo-: Good catch :fishing_pole_and_fish: Fixed now.

grnd-alt (approved):

PR #1163, opened 2024-03-16
Allow export collectives only if you have edit permission

Koc:
### 📝 Summary
Right now anybody can export Collectives. Would be nice to increase security and allow such export only for editors. In ideal world we would like to add an extra setting "allow exporting for"
image
. But this is not possible because of bitmask used for permissions, see `\OCA\Collectives\Db\Collective::editPermissions`.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/src/components/Collective/CollectiveActions.vue
+++ b/src/components/Collective/CollectiveActions.vue
@@ -17,7 +17,8 @@
 				<ShareVariantIcon :size="20" />
 			</template>
 		</NcActionButton>
-		<NcActionLink :close-after-click="true"
+		<NcActionLink v-if="collectiveCanEdit(collective)"
+			:close-after-click="true"
 			:href="printLink"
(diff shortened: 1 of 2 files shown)


mejo-:
Dear @Koc, what's your rationale for not allowing export/print to read-only users? The export/print functionality is nothing more than rendering all pages one below each other. It could easily be done anyways by read-only users by copy-pasting the individual pages. So I don't see any security gain by disabling this functionality for them.

Quite the contrary, I think the function to print (or export to PDF) the whole collective is especially useful to readers.

mejo-:
@Koc I'll close this PR for now. Feel free to comment on my questions above if you still think it's a good idea. I'm happy to reopen and further discuss this :relaxed:

Issue #1121, opened 2024-02-20
Overview page


juliushaertl:
* [ ] Add a button to the landing page that points to the corresponding teams overview page in contacts
* [ ] Only display the button if the contacts app is available

The idea would be to have the button right-aligned in the [MembersWidget](https://github.com/nextcloud/collectives/blob/main/src/components/Page/LandingPageWidgets/MembersWidget.vue) component.

The number of displayed members avatars is calculated dynamically depending on the element width, [see `updateShowMembersCount()`](https://github.com/nextcloud/collectives/blob/6718d83c388f880d9c6fe74cbab331aed6ff29eb/src/components/Page/LandingPageWidgets/MembersWidget.vue#L119-L124). This would need to be adjusted to always leave enough room for the new button.

PR #1168, opened 2024-03-20
add Team button that redirects to contacts app

grnd-alt:
### 📝 Summary

* Resolves: #1121 

adds a button to redirect to the team corresponding to the collective

#### 🖼️ Screenshots
image

🏚️ Before | 🏡 After
---|---
no button | button

### 🚧 TODO
write cypress tests

- [ ] ...

### 🏁 Checklist


src/components/Page/LandingPageWidgets/MembersWidget.vue
```diff
@@ -44,6 +44,12 @@ export default {
 		NcButton,
 		SkeletonLoading,
 	},
+	props: {
```
mejo-, line 47: Is there a reason why you pass `currentCollective` as property instead of using the vuex getter like before?

src/components/Page/LandingPageWidgets.vue
```diff
@@ -1,6 +1,16 @@
 <template>
 	<div class="landing-page-widgets">
-		<MembersWidget v-if="!isPublic" />
+		<div class="first-row-widgets">
+			<MembersWidget v-if="!isPublic" :current-collective="currentCollective" />
+			<NcButton v-if="'contacts' in OC.appswebroots" :href="teamUrl">
```
mejo-, line 5: Nitpicking, but maybe it makes sense to move condition `'contacts' in OC.appswebroots` into a computed to cache it?

src/components/Page/LandingPageWidgets.vue
```diff
@@ -1,6 +1,16 @@
...
+				<template #icon>
+					<TeamsIcon :size="20" />
+				</template>
+				<template #default>
+					{{ t('collectives','Team') }}
```
mejo-, line 10: I would call it "Team overview". What do others think @juliushaertl @nextcloud/designers?

mejo- (changes requested):
Thanks a lot, that was fast :blush: 

Overall looks good, but I have some comments.

src/components/Page/LandingPageWidgets.vue
szaimen: Team overview sounds good to me 👍

marcoambrosini (commented):
I think this should be in the action menu and not visible by default, and it should say "Team page"

mejo-:
> I think this should be in the action menu and not visible by default, and it should say "Team page"

To be honest I think that "Team page" is a bit ambiguous, given that "page" is a term used a lot in the collectives app for the pages of a collective. And from earlier discussions I remembered that we decided to have this button somehow visible and not hidden in an action menu, no? Maybe we can add it to both places?

src/components/Page/LandingPageWidgets/MembersWidget.vue
grnd-alt: LandingPageWidgets now needs the currentCollective as well. I feel like passing it as a prop is the better pattern than accessing the vuex store twice. If disagreed I can change it back.

mejo-: Interesting. I always thought that accessing it from the store was cleaner because it meant less passing on of properties. I don't have strong opinion here though and am open to arguments why passing it as a property is superior :wink:

grnd-alt: changed Back, MembersWidget already has vuex getters anyways, so might as well just leave it like it is.

mejo- (approved):
Nice, thank you!

PR #1193, opened 2024-04-09
add all subpages when filtering for collectives

grnd-alt:


max-nextcloud (commented):
Fix looks good.
No idea if the failing tests are related. It would also be great to have a test for this to prevent a regression.

Issue #505, opened 2023-01-16
It should be possible to share collectives with passwords enforced


brtptrs:
**Is your feature request related to a problem? Please describe.**
Currently, one can not share a collective when passwords are enforced for ecternal shares (via linl)

**Describe the solution you'd like**
Share collectives via a link, with password protection enabled, as discussed here:
https://help.nextcloud.com/t/why-can-collectives-not-be-shared-via-a-link-with-a-password

**Describe alternatives you've considered**
Share via a Circle with external contacts does not seem to work either.

**Additional context**
![grafik](https://user-images.githubusercontent.com/8102905/212719221-42fed6e3-590f-4274-a677-ebd085f8f86d.png)

Gymnae:
+1 for this request

max-nextcloud:
We need to figure out how to expose this in the user interface.

PR #1222, opened 2024-04-23
Allow to protect shares with a password

mejo-:
### 📝 Summary

Also supports adding shares when password protection for shares is enabled per default or enforced.

UI is more or less duplicated from the Files app.

* Fixes: #505 

#### 🖼️ Screenshots

Advanced share settings #1 | Advanced share settings #2
---|--- 
grafik | grafik

New share with password protection default  | New share with password protection enforced
--- | ---
grafik | grafik

### 🏁 Checklist


lib/Db/CollectiveShare.php
```diff
@@ -37,6 +38,14 @@ public function setEditable(bool $editable): void {
...
 	}
 
+	public function getPassword(): string {
+		return $this->password;
+	}
+
+	public function setPassword(string $password): void {
+		$this->password = $password;
+	}
+
```
max-nextcloud, line 48: Is there any difference between this and the `@method` comment used above for `getToken` and `setToken`?

lib/Db/CollectiveShare.php
```diff
@@ -45,6 +54,7 @@ public function jsonSerialize(): array {
 			'token' => $this->token,
 			'owner' => $this->owner,
 			'editable' => $this->editable,
+			'password' => $this->password,
```
max-nextcloud, line 57: Were does this get used / would it typically be used?

I'm wondering if we might want to leave the password hash out so it cannot leak - but then again we include the token, which we would not want to leak either. In logs I think the two will be filtered out.

src/store/collectives.js
```diff
@@ -373,12 +373,13 @@ export default {
 		 * @param {object} object the property object
 		 * @param {number} object.collectiveId ID of the collective to be shared
 		 * @param {number} object.pageId ID of the page to be shared
+		 * @param {number} object.password optional password for the share
```
max-nextcloud, line 376:
```suggestion
		 * @param {?string} object.password optional password for the share
```

max-nextcloud (approved):
## Looks great! Thanks for tackling this :partying_face: 

Only had a look at the code since you included a test.
I did not test the wrong password scenario and there's no test for it. Did you take a look how that works out?

Other than that I just have minor questions and a comment on jsdoc.

lib/Db/CollectiveShare.php
mejo-: Yes,  `@method` injects getters and setters for database fields. For transient attributes in the class that don't represent a database field, we have to define them manually. I just learned about this recently myself, when merging the `CollectiveShareInfo` and `CollectiveShare` classes. See https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/database.html#transient-attributes

lib/Db/CollectiveShare.php
mejo-: It's used in the frontend (sharing tab in sidebar, where shares are listed/configured), to determine if a password is set for the share. The files app does it exactly like this, but I agree that we could also just set a `hasPassword` flag in the `jsonSerialize()` function.

I don't think this will end up in logs, as its the payload that's sent as response body.

src/store/collectives.js
mejo-: Good catch, fixed now. It will always be an empty string tough, just empty if no password is set.

mejo-:
Thanks a lot for the quick review :heart: 

>  I did not test the wrong password scenario and there's no test for it. Did you take a look how that works out?

Sure, though the code that checks for a valid password is in `AuthPublicShareController` in the server code, so I thought it doesn't make much sense to add a test for it here.

I will look into adding a test for creating a share with enforced password protection though, since that's untested so far and probably something we don't use ourselves too often.

mejo-:
Mh, seems like something with password-protected shares is broken on stable27. Will have to further investigate.

mejo-:
> Mh, seems like something with password-protected shares is broken on stable27. Will have to further investigate.

Well, some debugging session revealed that the logic we use is broken with Nextcloud < 29 because this didn't happen before: [Router::generate()](https://github.com/nextcloud/server/blob/17725081bf4d5fbccab777f468f91437b50770b7/lib/private/Route/Router.php#L389). This results in [AuthPublicShareController::getAuthenticationRedirect()](https://github.com/nextcloud/server/blob/17725081bf4d5fbccab777f468f91437b50770b7/lib/public/AppFramework/AuthPublicShareController.php#L206) returning null.

We can either backport https://github.com/nextcloud/server/commit/66e7056c5ebd1538ca94ab3a1f04b6bf184170cb to stable27 and stable28 in server, or workaround this by adding extra camel-case entries for the corresponding routes:

```diff
diff --git a/appinfo/routes.php b/appinfo/routes.php
index 0e854889..fc9eff0e 100644
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -119,6 +119,10 @@ return [
                // Vue.js router public route (Vue.js frontend)
                ['name' => 'publicStart#showAuthenticate', 'url' => '/p/{token}/authenticate/{redirect}', 'verb' => 'GET'],
                ['name' => 'publicStart#authenticate', 'url' => '/p/{token}/authenticate/{redirect}', 'verb' => 'POST'],
+               // TODO: Remove the two uppercase PublicStart entries once Nextcloud 28 support gets removed
+               // See
+               ['name' => 'PublicStart#showAuthenticate', 'url' => '/p/{token}/authenticate/{redirect}', 'verb' => 'GET'],
+               ['name' => 'PublicStart#authenticate', 'url' => '/p/{token}/authenticate/{redirect}', 'verb' => 'POST'],
                ['name' => 'publicStart#showShare', 'url' => '/p/{token}/{path}', 'verb' => 'GET',
                        'requirements' => ['path' => '.*'],     'defaults' => ['path' => '']],
```

@juliushaertl @max-nextcloud what do you think? Is the workaround acceptable? Or would you prefer the server-side change to be backported?

mejo-:
> We can either backport https://github.com/nextcloud/server/commit/66e7056c5ebd1538ca94ab3a1f04b6bf184170cb to stable27 and stable28 in server, or workaround this by adding extra camel-case entries for the corresponding routes.

After discussing this with @provokateurin I decided to go with the workaround.

Issue #1094, opened 2024-01-26
Search page content when using the page filter in the left navigation


juliushaertl:
**Is your feature request related to a problem? Please describe.**
This came up in design discussions with @marcoambrosini 

We should think about providing page content search results when filtering in the page list of the left navigation.

**Describe the solution you'd like**
My initial suggestion would be to still list page title matches first by filtering the page list (as it currently works) but provide content matches in a second list below.

**Describe alternatives you've considered**
None so far

**Additional context**
This is also how the search in the Talk mobile app works for example, though I don't think we have a similar pattern yet on the web anywhere.

alexanderdd:
Also it would be great to then highlight the search term when you go to the page, as it is done in https://docs.nextcloud.com/server/latest/user_manual/en/ (which runs read the docs?)

But I think this also needs a higher-level discussion: There is the unified search in the top right, which is supposed to be THE way to find anything (right?). Afaik it already searches in collectives page contents. Then we don't need the same functionality in two places, right? But then how can users easily discover that the left search bar is just a page title filter, and the real search in in the top right?
Don't want to hijack this thread. If this discussion is already ongoing somewhere, please post a link.

juliushaertl:
Thanks for your feedback @alexanderdd 

> Also it would be great to then highlight the search term when you go to the page, as it is done in [docs.nextcloud.com/server/latest/user_manual/en](https://docs.nextcloud.com/server/latest/user_manual/en/) (which runs read the docs?)

Yes, though this is a separate issue. We are looking into providing search/highlight functionality in the text app which we first need to be able to use this to highlight results within Collectives then. https://github.com/nextcloud/collectives/issues/648

> But I think this also needs a higher-level discussion: There is the unified search in the top right, which is supposed to be THE way to find anything (right?). Afaik it already searches in collectives page contents. Then we don't need the same functionality in two places, right? But then how can users easily discover that the left search bar is just a page title filter, and the real search in in the top right?

While we already have the unified search it has quite a few design issues https://github.com/nextcloud/server/issues/24056 especially when it comes to providing quick access within the current visible app context. When discussing this in the past with the design team we came to the conclusion that local filtering also makes sense separately. I don't see why having the unified search would contradict with showing useful results also when performing local filtering in the current collective.

PR #1240, opened 2024-04-29
add content search to collectives

grnd-alt:
### 📝 Summary

* Resolves: #1094 

#### Screenshot
image

### 🏁 Checklist

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -80,6 +80,27 @@ class PageController extends Controller {
 		}, $this->logger);
 	}
 
+	public function contentFiltered(int $collectiveId, string $filterString): DataResponse {
+		return $this->handleErrorResponse(function () use ($collectiveId, $filterString): array {
+			$userId = $this->getUserId();
+			$pageInfos = $this->service->findAll($collectiveId, $userId);
+			$contentFilteredPages = [];
+			foreach ($pageInfos as &$pageInfo) {
+				$file = $this->service->getPageFile($collectiveId, $pageInfo->getId(), $userId);
+				$fileContent = $file->getContent();
+				if (str_contains(strtolower($fileContent), strtolower($filterString))) {
+					$contentFilteredPages[] = [
+						"page" => $pageInfo->jsonSerialize(),
+						"content" => $fileContent
+					];
+				}
+			}
+			return [
+				"data" => $contentFilteredPages
+			];
+		}, $this->logger);
+	}
+
 	/**
 	 * @NoAdminRequired
 	 */
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -88,11 +88,11 @@ class PageController extends Controller {
 	/**
 	 * @NoAdminRequired
 	 */
-	public function contentFiltered(int $collectiveId, string $filterString): DataResponse {
-		return $this->handleErrorResponse(function () use ($collectiveId, $filterString): array {
+	public function contentSearch(int $collectiveId, string $searchString): DataResponse {
+		return $this->handleErrorResponse(function () use ($collectiveId, $searchString): array {
 			$userId = $this->getUserId();
 			$collective = $this->collectiveService->getCollective($collectiveId, $userId);
-			$results = $this->indexedSearchService->searchCollective($collective, $filterString, 100);
+			$results = $this->indexedSearchService->searchCollective($collective, $searchString, 100);
 			$pages = [];
 			foreach ($results as $value) {
 				$pages[] = $this->service->find($collectiveId, $value['id'], $userId);
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace OCA\Collectives\Controller;
 
 use OCA\Collectives\Service\AttachmentService;
+use OCA\Collectives\Service\CollectiveService;
 use OCA\Collectives\Service\PageService;
+use OCA\Collectives\Service\SearchService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\IRequest;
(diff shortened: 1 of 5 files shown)


src/components/PageList.vue
```diff
@@ -99,8 +99,11 @@
 					:filtered-view="isFilteredview"
 					:is-template="true" />
 				<div v-if="isFilteredview">
-					<RecycleScroller v-slot="{item}"
+					<NcAppNavigationCaption v-if="filteredPages.length > 0" :name="t('Collectives','By Title')" />
```
jancborchardt, line 102:
```suggestion
					<NcAppNavigationCaption v-if="filteredPages.length > 0" :name="t('Collectives','Results in title')" />
```

src/components/PageList.vue
```diff
@@ -111,8 +114,28 @@
 							:filtered-view="true"
 							class="page-list-drag-item" />
 					</RecycleScroller>
+					<NcAppNavigationCaption v-if="loadingContentFilteredPages || contentFilteredPages.length > 0" :name="t('Collectives', 'By Content')" />
```
jancborchardt, line 117:
```suggestion
					<NcAppNavigationCaption v-if="loadingContentFilteredPages || contentFilteredPages.length > 0" :name="t('Collectives', 'Results in content')" />
```

jancborchardt (changes requested):
Looks nice! Just a small wording enhancement suggestion. :) (Note we use "Sentence case", so only first word capitalized.)

jancborchardt (approved):

lib/Controller/PageController.php
```diff
@@ -80,6 +80,27 @@ public function touch(int $collectiveId, int $id): DataResponse {
...
+			foreach ($pageInfos as &$pageInfo) {
+				$file = $this->service->getPageFile($collectiveId, $pageInfo->getId(), $userId);
+				$fileContent = $file->getContent();
+				if (str_contains(strtolower($fileContent), strtolower($filterString))) {
+					$contentFilteredPages[] = [
+						"page" => $pageInfo->jsonSerialize(),
+						"content" => $fileContent
+					];
+				}
+			}
```
juliushaertl, line 97: This is a bad idea performance wise as it will read the file content of every file. We had done this in the past but went with a different one.

We actually have a local full text search implementation within collectives that creates an index of the pages which we should rather use here. You can checkout the search provider of the server search for how that is being used: https://github.com/nextcloud/collectives/blob/1c7d1956cee227b2c1fa8df248249d3874927449/lib/Search/PageContentProvider.php#L69

References as we introduced this before moving collectives to github:
https://gitlab.com/collectivecloud/collectives/-/merge_requests/541

While there is also an existing search endpoint that the server provides, I think having our own one here makes sense as we could extend it further with highlighting that TNTsearch also provides.

lib/Trash/CollectivePageTrashItem.php
```diff
@@ -19,7 +19,7 @@ public function __construct(
 		IUser $user,
 		private string $mountPoint
 	) {
-		parent::__construct($backend, $originalLocation, $deletedTime, $trashPath, $fileInfo, $user);
+		parent::__construct($backend, $originalLocation, $deletedTime, $trashPath, $fileInfo, $user, null);
```
juliushaertl, line 22: @grnd-alt Maybe we can still split this, it just came to my mind that we'd probably want to have a version comparison here to only apply this for Nextcloud >= 30

src/components/PageList.vue
```diff
@@ -285,8 +333,13 @@ export default {
 <style lang="scss" scoped>
 
 .scroller {
-	// NC header bar 50px; page list header bar 52px; landing page 48px; page trash 76px
-	height: calc(100vh - 50px - 52px - 48px - 76px);
+	// NC header bar 50px; page list header bar 52px; landing page 48px; page trash 76px NcAppNavigationCaption 78px divided by 2 for multiple scrollers
+	max-height: calc((100vh - 50px - 52px - 48px - 76px - 78px * 2) / 2);
```
juliushaertl, line 337: Nitpick to use a css var where we have one:

```suggestion
	max-height: calc((100vh - var(--header-height) - 52px - 48px - 76px - 78px * 2) / 2);
```

lib/Controller/PageController.php
grnd-alt: I've implemented it to use the indexed Search. Should we change something abt the indexing? Idk how often it is done, but there are outdated results after writing to a page.

src/components/PageList.vue
```diff
@@ -245,6 +273,12 @@ export default {
 		},
 	},
 
+	watch: {
+		filterString() {
+			this.getContentFilteredPagesDebounced.apply(this)
```
mejo-, line 278: What's the reason for not just calling `this.getContentFilteredPageDebounced()` without the `.apply(this)`? Please forgive my question, it's just that I know the former and never saw the `.apply(this)` style so far.

src/apis/collectives/pages.js
```diff
@@ -1,5 +1,6 @@
 import axios from '@nextcloud/axios'
 import { pagesUrl } from './urls.js'
+import {generateUrl} from "@nextcloud/router";
```
mejo-, line 3: several eslint errors and warnings :wink:

mejo- (changes requested):
Sorry, some more nitpicking remarks. Apart from that good to go :blush:

src/apis/collectives/urls.js
```diff
@@ -34,3 +34,11 @@ export function pagesUrl(context, ...parts) {
 		? collectivesUrl('p', context.shareTokenParam, '_pages', ...parts)
 		: collectivesUrl(context.collectiveId, '_pages', ...parts)
 }
+
+/**
+ *
+ * @param {number} collectiveId collectiveId to search in
+ */
+export function getContentSearchUrl(collectiveId) {
```
mejo-, line 42:
```suggestion
export function contentSearchUrl(collectiveId) {
```

All the other functions in `urls.js` don't have `get` as prefix, so maybe stick to that for now?

Issue #1225, opened 2024-04-24
"Share with guests" menu entry on collective page is not consistent with guests meaning in Nextcloud (Guests app)


maximelehericy:
**Describe the bug**
"Share with guests" menu entry on collective page is not consistent with guests meaning in Nextcloud (Guests app)

"Share with guests" menu entry on collective page allows you to create a public link to a collective page. As it is public link, theoretically, the user accessing the page is anonymous to Nextcloud.

In Nextcloud, Guests accounts are specials user accounts. But from a Nextcloud standpoint, it is a known and registered user.

The naming "Share with guests" is confusing in that regard.
@mejo- @jancborchardt 

**To Reproduce**
1. Go to a collective
2. Click on the three dot menu
4. Click "Share with guests"
5. What you can do is create a public link, not sharing with a guest (user) in the sense of Nextcloud.

**Expected behavior**
I would rename Share with Guests to generate public link, or share publicly.

**Screenshots**

**Server details:**
- Collectives app version: [e.g. 2.8.2; see Nextcloud apps page]
- Nextcloud version: [e.g. 26.0.1]
- PHP Version: [e.g. 8.2]
- Database: [e.g. MariaDB 10.6]

**Client details:**
 - OS: [e.g. Windows/macOS/Ubuntu]
 - Browser: [e.g. Firefox, Chrome]
 - Browser version: [e.g. 22]
 - Device: [e.g. iPhone6, desktop]

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

jancborchardt:
Yep, to be consistent with how it is in Files, the correct wording would be "Share link".

We never use the wording "public" since it sounds like publishing something globally, which is not what is happening since the link is still quite private (and has a lot of security options too).

PR #1248, opened 2024-05-07
Fix of inconsistent and confusing public share wording

Jerome-Herbinet:
I'm correcting the wording of the share button because : 
- it creates confusion with the notion of “guest” (in relation to the “Guests” app). This app should have exclusive rights to the notion of guest.
- it is inconsistent in terms of vocabulary with what already exists in the Photos app for a rather similar context, which uses “Share via public link”.
- Fixes: #1225


juliushaertl (approved):

Jerome-Herbinet:
Do I have to worry of Cypress errors ?

jancborchardt (changes requested):
We should never use the wording "public link" since it sounds like publishing something. The link is still not guessable, thus not public.

In Files we use "Share link", which is what we should use here (and in Photos) as well. :)

Jerome-Herbinet:
> We should never use the wording "public link" since it sounds like publishing something. The link is still not guessable, thus not public.
> 
> In Files we use "Share link", which is what we should use here (and in Photos) as well. :)

@jancborchardt, thanks for your reply. "Share link" is relevant with "Files", in the sharing properties on the right-hand side panel. In "Share link" describes what it is, but not the action. Do you think that we can use share link like this ? (see screenshot) : 

I ask this question because, except “settings”, all the available options use an action verb. So “share link” doesn't seem to be well expressed in this part of the interface. So I still suggest using the verb “Share”, but with something else to complete instead of “via public link”. What do you suggest? (keep in mind that only “share” isn't enough, as users might think that sharing can also be done with other users (too general a word).

I also notice (see the screenshot) that en share entry is displayed twice. Strange, isn't it ?

mejo-:
> I still suggest using the verb “Share”, but with something else to complete instead of “via public link”.

Actually "Share link" contains an action verb, which is "share", no? You can read it as a short form of "Share the link" or "Share via link".

> I also notice (see the screenshot) that en share entry is displayed twice. Strange, isn't it ?

Oh, that's another bug. If you don't mind I push a fix for this into your PR.

mejo- (approved):

PR #1251, opened 2024-05-08
adapt collectives for NC 30

grnd-alt:
NC major 30 and above require a deletedby parameter for the Trashitem constructor. For now we should pass null to make collectives work again/ let CI be green again.

### DIFF
--- a/lib/Trash/CollectivePageTrashItem.php
+++ b/lib/Trash/CollectivePageTrashItem.php
@@ -8,6 +8,7 @@ use OCA\Files_Trashbin\Trash\ITrashBackend;
 use OCA\Files_Trashbin\Trash\TrashItem;
 use OCP\Files\FileInfo;
 use OCP\IUser;
+use OCP\Util;
 
 class CollectivePageTrashItem extends TrashItem {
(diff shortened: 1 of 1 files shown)


lib/Trash/CollectivePageTrashItem.php
```diff
@@ -19,7 +20,13 @@ public function __construct(
 		IUser $user,
 		private string $mountPoint
 	) {
-		parent::__construct($backend, $originalLocation, $deletedTime, $trashPath, $fileInfo, $user);
+		[$major] = Util::getVersion();
+		if ($major < 30) {
+			parent::__construct($backend, $originalLocation, $deletedTime, $trashPath, $fileInfo, $user);
+		} else {
+			// *TODO* add $deletedby parameter to constructor
```
juliushaertl, line 27:
```suggestion
			// *TOTO* Add support for deletedby to collectives trash backend table
```

But I'd suggest to also file an issue to not forget before the 30 final

juliushaertl (approved):
Nitpick but otherwise 👍

blizzz (approved):

PR #1256, opened 2024-05-13
test(cypress): Fix typing into editor

mejo-:
`cy.type()` requires a typable element. `cy.getEditor()` returns the wrapper which is not writable, so use `cy.getEditorContent(true)` to get the contenteditable element.


max-nextcloud (approved):
quick code review makes sense.

Issue #1259, opened 2024-05-14
Leaving a collective does not work


jakobroehrl:
I created so test collectives with only me as a user. 
But now I can't delete them. 
Leave collective works but the collective shows up again after a page reload
Latest NC and collective version, tested on two different servers.
Console error message:

`
Failed to leave collective 
[...]

Issue #1103, opened 2024-02-01
Add ability to delete a collective


marcoambrosini:
**Is your feature request related to a problem? Please describe.**
I'm not able to delete a collective for which I'm the only member. Right now it says "leave collective" and the action does not work. So I guess it's also a bug?  

**Describe the solution you'd like**
Button that says "delete collective" if I'm the only member and permanent deletion of it.

**Describe alternatives you've considered**

**Additional context**

marcoambrosini:
Nice, I think this should be available in the context menu in place of leave if you're the only member.

mejo-:
> Nice, I think this should be available in the context menu in place of leave if you're the only member.

Mh, this would mean the setting would be there for some collectives (where you're the sole member) and not for others. I think this would create a confusing user experience.

I agree though that "Leave collective" should be hidden if you're the sole member.

PR #1262, opened 2024-05-14
fix(CollectiveActions): Show 'leave collective' option conditionally

mejo-:
Don't show the option if either of:
* Team member has other type than `TYPE_USER` (e.g. is member through a group or another team).
* Team member is only member left.

Fixes: #1103
Fixes: #1259

### 🏁 Checklist


lib/Service/CircleHelper.php
```diff
@@ -260,6 +260,37 @@ public function getLevel(string $circleId, string $userId): int {
...
+	 * @throws NotPermittedException
+	 * @throws MissingDependencyException
+	 */
+	public function canLeave(string $circleId, string $userId): bool {
+		if (is_null($this->circlesManager)) {
+			throw new MissingDependencyException($this->dependencyInjectionError);
+		}
+
+		try {
+			$this->startSession($userId);
```
blizzz, line 274: not sure what it means, gives me shivers

lib/Service/CircleHelper.php
```diff
@@ -260,6 +260,37 @@ public function getLevel(string $circleId, string $userId): int {
...
+			}
+			$members = $circle->getMembers();
+		} catch (CircleNotFoundException $e) {
+			throw new NotFoundException($e->getMessage(), 0, $e);
+		}
+
+		foreach ($members as $member) {
+			if ($member->getSingleId() !== $initiator->getSingleId()) {
+				return true;
+			}
```
blizzz, line 288: This returns true with the first user that does not match the initiator. Is this intended?

blizzz (commented):
One question.

lib/Service/CircleHelper.php
mejo-: Hehe, I understand. That's how the Circles API works. It starts a session for `CirclesManager` as the user. Afterwards, we can probe the `CirclesManager` for circles in the name of this user. See https://github.com/nextcloud/circles/blob/3bf2a7bd617785d94c382222d52e96814d2569f3/lib/CirclesManager.php#L175

lib/Service/CircleHelper.php
mejo-: Exactly. Because this means that at least one other member exists in the circle, so we can leave it.

blizzz (approved):

Issue #1122, opened 2024-02-20
"Last changed by" is updating the changed by name of the parent page, but not the date


kevrl:
**Describe the bug**
When creating a sub page to a page in Nextcloud Collectives it will have a timestamp with the last edit like this: "Last changed by kevrl seconds ago".
the parent page to the sub page will have it's name updated to whoever edited or created this new sub page. 
Example: 
If I create a new sub page on an already existing page that have been edited by someone else, then my name will show up as "last changed by" but the timestamp will be unchanged for the parent page. 

**To Reproduce**
1. Open up Nextcloud Collectives
2. Select a collective
3. Select an existing page (That has been last changed by some other user than you) 
4. Press the "+" symbol to the right of the page name to create a sub page.
5. Write anything in the new page 
6. Save
7. Go to parent page
8. Check the "last changed by" 

**Expected behavior**
I would expect that one of the following would happen:  
A) The Name and Date of the parent page should not be updated at all  
B) The Name and Date of the parent page should both be updated.  

**Screenshots**
Screenshot of the parent page: 
![image](https://github.com/nextcloud/collectives/assets/160610709/675bf8f6-c4ea-4234-bfe1-204d9ce6b8da)  
  
Screenshot of the sub page:
![image](https://github.com/nextcloud/collectives/assets/160610709/310fdf68-9a6a-4bde-b87f-48c8dde1b9e6)

**Server details:**
- Collectives app version: TBA
- Nextcloud version: 27.1.5.1
- PHP Version: 8.2.15
- Database: MySQL 10.7.8

**Client details:**
 - OS: Arch Linux x86_64
 - Browser: Firefox
 - Browser version: 122.0.1 (64-bit)
 - Device: Desktop

arnowelzel:
In the latest version (2.10.1 running in Nextcloud 29.0.0) even the name is missing now. None of my documents has any name at all and the date is always referring to the creation date of the document and not to the last update.

![image](https://github.com/nextcloud/collectives/assets/6613614/4ad0f822-eefe-4b3f-a392-59ea34c55177)

PR #1284, opened 2024-05-21
Last edit user related fixes

mejo-:
### 📝 Summary

* fix(LastUserBubble): Fix displaying user bubble
* fix(PageService): Don't update last edit user with subpageOrder
* Resolves: #1122 

### 🏁 Checklist

### DIFF
--- a/lib/Service/PageService.php
+++ b/lib/Service/PageService.php
@@ -249,17 +249,29 @@ class PageService {
 		]);
 	}
 
-	private function updatePage(int $collectiveId, int $fileId, string $userId, ?string $emoji = null, ?string $subpageOrder = null): void {
+	private function updatePage(int $collectiveId, int $fileId, string $userId, ?string $emoji = null): void {
 		$page = new Page();
 		$page->setFileId($fileId);
 		$page->setLastUserId($userId);
 		if ($emoji !== null) {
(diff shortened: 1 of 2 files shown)


lib/Service/PageService.php
```diff
@@ -249,10 +249,12 @@ private function notifyPush(int $collectiveId, string $userId): void {
 		]);
 	}
 
-	private function updatePage(int $collectiveId, int $fileId, string $userId, ?string $emoji = null, ?string $subpageOrder = null): void {
+	private function updatePage(int $collectiveId, int $fileId, string $userId, ?string $emoji = null, ?string $subpageOrder = null, bool $updateEditUser = true): void {
```
max-nextcloud, line 252: That's a long list of arguments. I'd define a separate function for updating the subpage order instead. That way you'd have two functions with 4 arguments rather than one with 6.
As far as i can tell we never update the subpage order and the emoji at the same time. We also never update the last user when updating the subpage order etc. - It's nice to have a flexible and powerful function when you need it. Here it feels like you never need the added complexity it brings.

max-nextcloud (approved):

lib/Service/PageService.php
mejo-: Good point, thanks. Refactored now.

Issue #1118, opened 2024-02-18
Use notify_push for show add / remove of pages


wargreen:
**Is your feature request related to a problem? Please describe.**
When friends add / remove pages, i need to reload to see the change.
This lead multiple-create / remove when working together.

**Describe the solution you'd like**
Use notify_push (the  Client Push app) for live update the list of pages.

**Describe alternatives you've considered**
Use discussion for know when i need F5

**Additional context**
notify_push is used for desktop client sync, deck live update etc... It seem to me that can fluidize the collective workflow.

mejo-:
Dear @wargreen, this is already the case. Apparently the `notify_push` implementation broke with Collectives 2.11.0 due to a bug in `@nextcloud/axios` 2.5.0, but before that version changes to the page list should have get propagated via `notify_push`.

See [`initListenPush` in `Collective.vue`](https://github.com/nextcloud/collectives/blob/1b33d72e50e471febfd61b04188dd77e984fb38c/src/components/Collective.vue#L115) for the frontend part and [`notifyPush()` in `PageService`](https://github.com/nextcloud/collectives/blob/1b33d72e50e471febfd61b04188dd77e984fb38c/lib/Service/PageService.php#L241) for the backend part.

Did you actually test whether this works on your instance? As mentioned, with latest Collectives release (2.11.0) it's broken, but if you still run an older instance, you should see a websocket connection to your notify_push service when loading the Collectives app.

mejo-:
Looking into this again I now understood that synchronization via `notify_push` only works for sessions of the same user so far. So indeed there's more work needed to sync changes between different users.

wargreen:
Yay !
Thank you for all the work !

PR #1288, opened 2024-05-23
Notify_push sessions

mejo-:
### 📝 Summary

Adds session management to keep track which users are active in a collective. Send notify_push notifications to all active session users.

* Resolves: #1118 

### 🏁 Checklist


elzody (approved):
Front end changes look good

lib/Service/PageService.php
```diff
@@ -41,7 +41,8 @@ public function __construct(private IAppManager $appManager,
 		private UserFolderHelper $userFolderHelper,
 		private IUserManager $userManager,
 		private IConfig $config,
-		ContainerInterface $container) {
+		ContainerInterface $container,
+		private SessionService $sessionService) {
```
hweihwang, line 45: Should we inject the Iqueue here too

mejo-: IQueue might be unavailable if the notify_push app is not installed. That's why it's taken from the container with exception handling.

lib/Migration/Version021200Date20240522000000.php
```diff
@@ -0,0 +1,49 @@
...
+				'autoincrement' => true,
+				'notnull' => true,
+			]);
+			$table->addColumn('user_id', Types::STRING, [
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->addColumn('token', Types::STRING, [
+				'notnull' => true,
+				'length' => 64,
```
blizzz, line 30: 32?

lib/Service/PageService.php
juliushaertl: work for >28 https://github.com/nextcloud/server/pull/41265

lib/Service/SessionService.php
```diff
@@ -0,0 +1,79 @@
...
+		return $this->sessionMapper->deleteInactive();
+	}
+
+	public function getSessionUsers(int $collectiveId, ?string $sessionToken = null): array {
+		$activeSessions = $this->sessionMapper->findAllActive($collectiveId);
+		$userIds = [];
+		foreach ($activeSessions as $session) {
+			if ($sessionToken && $session->getToken() === $sessionToken) {
+				// Limitation:
+				// If same use ras a second active session, $sessionToken still gets notified
```
blizzz, line 65:
```suggestion
				// If same user has a second active session, $sessionToken still gets notified
```

lib/Service/SessionService.php
```diff
@@ -0,0 +1,79 @@
...
+				// because notify_push notifies users, not individual sessions
+				// https://github.com/nextcloud/notify_push/issues/195
+				continue;
+			}
+
+			// Don't notify same user multiple times
+			if (!in_array($session->getUserId(), $userIds, true)) {
+				$userIds[] = $session->getUserId();
```
blizzz, line 73: use user ids as array key, then you save the in_array calls, and can return the values with array_keys($userIds), e.g.

```php
$userIds[(string)$session->getUserId()] = 1;
…
return array_keys($userIds);
```

tests/Integration/features/collectiveSession.feature
```diff
@@ -0,0 +1,28 @@
...
+    When user "bob" fails to close session for "BehatSessionCollective" with token "invalid"
+
+  Scenario: Create, update and close a session as member
+    When user "jane" creates session for "BehatSessionCollective"
+    Then user "jane" fails to update session for "BehatSessionCollective" with token "invalid"
+    And user "jane" updates session for "BehatSessionCollective"
+    And user "jane" closes session for "BehatSessionCollective"
+
+  Scenario: Close a non-existing session as member
+    And user "jane" closes session for "BehatSessionCollective" with token "invalid"
```
blizzz, line 19:
```suggestion
    When user "jane" closes session for "BehatSessionCollective" with token "invalid"
```

tests/Integration/features/collectiveSession.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  Scenario: Create a session as member, fail to update after it got invalidated
+    When user "jane" creates session for "BehatSessionCollective"
+    And we wait for "95" seconds
+    Then user "jane" fails to update session for "BehatSessionCollective"
+
+  Scenario: Delete collective
+    And user "jane" trashes collective "BehatSessionCollective"
+    And user "jane" deletes collective+team "BehatSessionCollective"
```
blizzz, line 28:
```suggestion
    When user "jane" trashes collective "BehatSessionCollective"
    And user "jane" deletes collective+team "BehatSessionCollective"
```

blizzz (approved):
Nitpicks :smiley:

hweihwang (approved):

lib/Migration/Version021200Date20240522000000.php
mejo-: Not sure. The deck app uses a column size of `64`, so I copied that. Since we generate a random string with 32 characters, DB length of 32 should be enough in theory. I don't know enough about the internals if that's really the case though and feel a bit insecure to limit the DB field size that much. If you think it's safe, I'll change it :grimacing:

lib/Service/SessionService.php
mejo-: Nice, thanks :blush:

lib/Migration/Version021200Date20240522000000.php
blizzz: Since regular ascii chars are being used, that should be fine, but if ever the token length would be increased or multi-byte characters added, it would require a migration change. I would go with 32, but I do not have a strong opinion either. 

In DirectEditing, for example, length is 64, but the token is also generated with 64 characters (but the chosen pool for the token has 4 characters less)

PR #1315, opened 2024-06-10
fix(debounce): Register debounced functions in instance context

mejo-:
This fixes "Error: Debounced method called with different contexts." errors.

### 🏁 Checklist

### DIFF
--- a/src/components/Page/LandingPageWidgets/MembersWidget.vue
+++ b/src/components/Page/LandingPageWidgets/MembersWidget.vue
@@ -48,6 +48,7 @@ export default {
 	data() {
 		return {
 			showMembersCount: 3,
+			updateShowMembersCountDebounced: debounce(this.updateShowMembersCount, 50),
 		}
 	},
 
@@ -90,7 +91,7 @@ export default {
 	watch: {
 		'sortedMembers.length'() {
 			this.$nextTick(() => {
-				this.updateShowMembersCount()
+				this.updateShowMembersCountDebounced()
 			})
 		},
 	},
@@ -100,11 +101,11 @@ export default {
 	},
 
 	mounted() {
-		window.addEventListener('resize', this.updateShowMembersCount)
+		window.addEventListener('resize', this.updateShowMembersCountDebounced)
 	},
 
 	unmounted() {
-		window.removeEventListener('resize', this.updateShowMembersCount)
+		window.removeEventListener('resize', this.updateShowMembersCountDebounced)
 	},
 
 	methods: {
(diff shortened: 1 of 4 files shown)


src/components/Page/LandingPageWidgets/MembersWidget.vue
```diff
@@ -100,11 +101,11 @@ export default {
 	},
 
 	mounted() {
-		window.addEventListener('resize', this.updateShowMembersCount)
+		window.addEventListener('resize', this.updateShowMembersCountDebounced)
```
max-nextcloud, line 104: I'm not sure if this will work. I'd expect the function to be called without it being bound to `this`. So this inside the function would be undefined.

src/components/Page/LandingPageWidgets/MembersWidget.vue
```diff
@@ -116,14 +117,14 @@ export default {
 			'setMembersCollectiveId',
 		]),
 
-		updateShowMembersCount: debounce(function() {
+		updateShowMembersCount() {
 			// How many avatars (44px width + 12px gap) fit? Substract one for the more button.
 			const membersWidth = this.$refs.members?.clientWidth
 			if (membersWidth) {
 				const maxMembers = Math.floor(membersWidth / 56) - 1
 				this.showMembersCount = Math.min(this.sortedMembers.length, maxMembers)
```
max-nextcloud, line 125: here we're using `this`.

max-nextcloud (commented):
I have not looked at the code in detail yet. Just one thing that stood out to me.

My understanding is that event listeners will be called as plain functions without being bound to the object they come from. So they will have no notion of `this` unless you do something like:
```js
addEventListener('resize', () => this.handler())
```
or
```js
addEventListener('resize', () => this.handler.bind(this))
```

But I might be wrong there.

src/components/Page/LandingPageWidgets/MembersWidget.vue
mejo-: I did some testing around the resize listener because it indeed was broken (already before this change). Using `this` inside the debounced function seems to work well though.

I'll open a follow-up PR soon that fixes the broken resize logic soon.

max-nextcloud: You're right. I stepped through it and `this.updateShowMembersCound` is bound to `this`. I suspect that vue already binds the methods to `this`.

max-nextcloud (approved):
Looks good to me. Thanks for adressing this!

Issue #1249, opened 2024-05-07
Problem displaying members of collectives without members after creation 


Jerome-Herbinet:
**Describe the bug**
Check the video :

https://github.com/nextcloud/collectives/assets/33763786/4b970531-b2f4-4522-a1de-255c02b70a86

**Server details:**
Operating system: Linux 5.14.0-362.24.1.el9_3.0.1.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Apr 4 22:31:43 UTC 2024 x86_64

Webserver: nginx/1.25.5 (fpm-fcgi)

Database: mysql 10.6.17

PHP version: 8.2.18

Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, json, random, Reflection, SPL, session, standard, cgi-fcgi, bcmath, bz2, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, gmp, iconv, imap, intl, ldap, exif, mysqlnd, PDO, pgsql, Phar, posix, shmop, SimpleXML, sockets, sodium, sqlite3, sysvmsg, sysvsem, sysvshm, tokenizer, xml, xmlwriter, xsl, mcrypt, mysqli, pdo_mysql, pdo_pgsql, pdo_sqlite, xmlreader, xmlrpc, zip, apcu, imagick, msgpack, smbclient, igbinary, memcached, redis, libsmbclient, Zend OPcache

Nextcloud version: 29.0.0 - 29.0.0.19

Collectives app : 2.10.1

**Client details:**
 - OS: Ubuntu desktop 22.04 LTS
 - Browser: Firefox 25.0.1

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Can be provided if necessary
```

#### Browser log
```
Can be provided if necessary
```

</details>

PR #1316, opened 2024-06-10
fix(MembersWidget): Resize according to available space

mejo-:
Fixes: #1249

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist

### DIFF
--- a/src/components/Page/LandingPageWidgets/MembersWidget.vue
+++ b/src/components/Page/LandingPageWidgets/MembersWidget.vue
@@ -150,6 +150,10 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+.members-widget {
+	flex-grow: 1;
+}
+
 .members-widget-skeleton {
 	height: 44px;
 }


elzody (approved):

src/components/Page/LandingPageWidgets/MembersWidget.vue
```diff
@@ -118,9 +118,10 @@ export default {
 
 		updateShowMembersCount: debounce(function() {
 			// How many avatars (44px width + 12px gap) fit? Substract one for the more button.
-			const membersWidth = this.$refs.members?.clientWidth
-			if (membersWidth) {
-				const maxMembers = Math.floor(membersWidth / 56) - 1
+			const parentWidth = this.$refs.members?.parentElement?.clientWidth
+			const siblingWidth = this.$refs.members?.nextElementSibling?.clientWidth
+			if (parentWidth && siblingWidth) {
+				const maxMembers = Math.floor((parentWidth - (siblingWidth + 20)) / 56) - 1
```
max-nextcloud, line 124: This only works if there is only one sibling following the members widget, right?

Do we really need to rely on that or could we make the members ref always fill the available space and measure it instead?

max-nextcloud (approved):

src/components/Page/LandingPageWidgets/MembersWidget.vue
mejo-: Indeed, much better. The required fix turned out to be a simple `flex-grow: 1` on the right spot :magic_wand:

PR #1318, opened 2024-06-11
chore(release): Prepare 2.12.0 release

mejo-:


max-nextcloud (approved):
Thanks a lot for taking care of this! :heart: 

I imagine you also need to bump the app version but will do that separately?

Issue #1192, opened 2024-04-09
Share link users cannot close "Recent pages"


chmac:
**Describe the bug**
According to #835 I understood that the "Recent pages" should be closable and this would be remembered per user. But in our "share link" view, when the link has read only view to the page, we are unable to close it at all.

**To Reproduce**
1. Create a collective
2. Create a share link
3. Open share link in a private tab
4. Notice that clicking "Recent pages" does not fold the section

**Expected behavior**
I expected that the share link would be able to close the section, and ideally, that it would remain closed for future users of the same link.

**Screenshots**

<img width="677" alt="Screenshot 2024-04-09 at 08 20 17" src="https://github.com/nextcloud/collectives/assets/690997/197d2822-c56f-4724-97c2-e1b726493040">

I think the Recent pages is actually greyed out here.

**Server details:**
- Collectives app version: 2.10.1
- Nextcloud version: 28.0.4
- PHP Version: Version: 8.2.17
- Database: Type: mysql Version: 10.5.24

**Client details:**
 - OS: macOS
 - Browser: Chrome, Safari

estux:
I'm experiencing the same bug. 
Navigating a publicly shared Collective (with Only Read permission) I cannot fold the Recent Pages section on top of the Landing Page. 
The little arrow seems not "clickable" on mobile.
I'm using Fennec / Firefox android 125.3.0
Thank you

PR #1320, opened 2024-06-12
fix(recentPages): Allow toggleing in public share

mejo-:
Fixes: #1192

### 🏁 Checklist


max-nextcloud (approved):
Code looks good. Did not test.

Bit surprised this toggle changes the default for everyone - but i guess there's no other option.

mejo-:
> Bit surprised this toggle changes the default for everyone - but i guess there's no other option.

For non-public collectives, it persists the default just for the current collective and user.

Issue #648, opened 2023-04-18
Search highlight


vdtoorn:
**Is your feature request related to a problem? Please describe.**
When using collectives with somewhat large pages, it is very hard to find a searched text, because the search only every shows the page and does not highlight the text. Also, ctrl-F to search in the browser window does not work, because that is overruled by the global search option, and '/' to search is crippled somehow too.

**Describe the solution you'd like**
At the minimal it would be advantageous to highlight the current search terms in the current collective window. Alternatively, going through the result list could jump to the correct location within the collective page.

**Describe alternatives you've considered**
I can't think of alternatives other than '/', but that also does not work for me if the search term is in the pages list for instance.

**Additional context**
-

juliushaertl:
https://github.com/ueberdosis/tiptap/pull/2075 could be some useful inspiration for implementing highlights in text

PR #1368, opened 2024-07-15
feat: Search highlighting for Collectives

elzody:
### 📝 Summary

* Resolves: #648 

This pull request integrates search highlighting from the text app into collectives.

#### 🖼️ Video

https://github.com/user-attachments/assets/c8614202-f057-464f-94ed-de05b38249ef

### 🚧 TODO

- [ ] Tests, will be implemented later and tracked here #1381

### 🏁 Checklist

### DIFF
--- a/src/components/SearchDialog.vue
+++ b/src/components/SearchDialog.vue
@@ -1,28 +1,23 @@
 <template>
 	<div v-if="totalMatches !== null" class="search-dialog__container">
-		<div class="search-dialog__info">
-			<span v-if="matchAll">
-				{{ t('collectives', 'Found {matches} matches', {
-					matches: totalMatches,
-				}) }}
-			</span>
-
-			<span v-else>
-				{{ t('collectives', 'Match {index} of {matches}', {
-					index: matchIndex + 1,
-					matches: totalMatches,
-				}) }}
-			</span>
-		</div>
-
 		<div class="search-dialog__buttons">
+			<NcButton alignment="center-reverse"
+				type="tertiary"
+				:aria-label="t('collectives', 'Clear search')"
+				@click="clearSearch">
+				<template #icon>
+					<Close :size="20" />
+				</template>
+				{{ t('collectives', 'Clear search') }}
+			</NcButton>
+
 			<NcButton alignment="center-reverse"
 				:aria-label="t('collectives', 'Find previous match')"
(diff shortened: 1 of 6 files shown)


nimishavijay:
Looks really nice overall! Some comments from my side: 

- [ ] "Find prev" --> "Find previous" to be clear to non-native English speakers as well as to avoid translation issues
- [ ] add a few pixels more margin to the left of "Found 34 matches" 
- [ ] if possible, we could also show the search string in the bottom bar by changing the wording like `Found 34 matches for "she"`
- [ ] Usually the icons for buttons are on the left, so we should follow that pattern here too in the "Find next" and "Find previous" buttons
- [ ] the "Find all" toggle appears as a button and that may be an accessibility issue, so we can change that to a checkbox element and change the wording to "Highlight all matches" (ref Firefox's design)
- [ ] it would also be nice to have a tertiary "Clear search" button on the left of "Find previous" (it could have a `X` icon) 


Other than that, it looks really good! :)

jancborchardt (dismissed):
Nice! Only additional feedback in addition to @nimishavijay’s is:
- [ ] The elements in the left navigation directly below the search field jump around vertically a little bit. It would be nice if they stay in place, e.g. the main Collective page.

juliushaertl:
> Nice! Only additional feedback in addition to @nimishavijay’s is:
> 
> - [ ] The elements in the left navigation directly below the search field jump around vertically a little bit. It would be nice if they stay in place, e.g. the main Collective page.

Let's file that into a separate issue. Unrelated to this part of the implementation

AndyScherzinger (approved):

src/components/SearchDialog.vue
```diff
@@ -0,0 +1,159 @@
...
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+$button-gap: calc(var(--default-grid-baseline) * 3);
+
+.search-dialog__container {
+	width: 100%;
+	height: 50px;
```
mejo-, line 133: I would remove the `height` property altogether. With the `margin-bottom` and `margin-top` suggested fo rthe highlight button below, it's guaranteed that it's high enough. And on narrow screens where buttons become two lines high, `50px` are not enough.

mejo- (approved):

Issue #1359, opened 2024-07-10
Visual bugs after new design changes


nimishavijay:
![image](https://github.com/nextcloud/collectives/assets/52440189/8ef64657-20e1-4bac-b1ea-bb2bc2d29d52)

Page list

* [x] Items are too large --> should be var(--default-clickable-area) height  
* [x] There can be a bit more whitespace between the collective name and first page
* [x] Emojis are too big --> currently 1.3em, can be 1.2 or even 1.1em
* [x] All buttons and field at the top should be center aligned
* [x] Template button is too close to the search field (on hover their edges touch)

![image](https://github.com/nextcloud/collectives/assets/52440189/f838a836-b786-44c6-8968-076e726d7a5a)

Deleted pages item has too much spacing on the bottom

![image](https://github.com/nextcloud/collectives/assets/52440189/e01091e5-2326-42a3-a099-b3ec5768540b)
Top right buttons in a page are not aligned and spaced out properly

![image](https://github.com/nextcloud/collectives/assets/52440189/5472a925-042c-4f46-88cf-ee4b0ccb3201)

Top right action menu:

* [x] First item is too large, should be default clickable area
* [x] Slight misalignment of checkbox (component issue?)

![image](https://github.com/nextcloud/collectives/assets/52440189/2a80ae5f-d8d5-43db-bf79-6adaaf07f8cf)

Recent pages:

* [x] Items are too rounded --> currently using var(border-radius-rounded), can be var(border-radius-large)
* [ ] the chevron-down button has no hover feedback (but the cursor changes, so it's ok?)

![image](https://github.com/nextcloud/collectives/assets/52440189/f0a9195f-27c1-4257-abc9-db912c6b4b60)

Avatars and the button is too big, should be default clickable area

![image](https://github.com/nextcloud/collectives/assets/52440189/18529a52-42ed-48ff-bfe7-8030407821e5)

Copy/move page:

* [x] Items are too big, should be default clickable area like the page list
* [x] Items don't have any border radius at all
* [ ] Use NcDialog component
* [x] Add space below breadcrumbs element
* [x] Root button (the collectives icon) is too wide? (rectangular on hover)
* [ ] Last item of breadcrumbs is lighter in color than the rest

![image](https://github.com/nextcloud/collectives/assets/52440189/d6b7a41a-e091-42ca-a7a1-1daeec761815)
Creating a new collective
- [x] User items are too big, use default clickable area
- [x] Too much space below "Add accounts"

[...]

PR #1376, opened 2024-07-22
Adjust styling to not break with Nextcloud 30 design

mejo-:
### 📝 Summary

* Resolves: #1359 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist

### DIFF
--- a/src/components/Member/Member.vue
+++ b/src/components/Member/Member.vue
@@ -15,7 +15,7 @@
 			:disable-menu="true"
 			:disable-tooltip="true"
 			:show-user-status="false"
-			:size="44" />
+			:size="avatarSize" />
 
 		<!-- Data -->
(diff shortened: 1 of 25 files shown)


src/components/Member/Member.vue
```diff
@@ -15,7 +15,7 @@
 			:disable-menu="true"
 			:disable-tooltip="true"
 			:show-user-status="false"
-			:size="44" />
+			:size="34" />
```
juliushaertl, line 18: Should this also use the css variable value or is this expected to be also smaller on older versions?

juliushaertl (approved):

src/components/Member/Member.vue
mejo-: Good catch, changed now.

PR #1386, opened 2024-07-25
Save full width of page on server

Koc:
### 📝 Summary
Currently we are storing "Full width" inside local storage. This means, that this setting not shares between users and browsers. Would be nice to make this setting persistent on db level like in Confluence.

#### 🖼️ Screenshots
https://github.com/user-attachments/assets/8b7372fe-807c-4a69-8db3-6d359a76ac0b

### 🚧 TODO
- [x] add tests after code review

### 🏁 Checklist

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -61,6 +61,8 @@ return [
 			'verb' => 'POST', 'requirements' => ['collectiveId' => '\d+', 'parentId' => '\d+']],
 		['name' => 'page#touch', 'url' => '/_api/{collectiveId}/_pages/{id}/touch',
 			'verb' => 'GET', 'requirements' => ['collectiveId' => '\d+', 'id' => '\d+']],
+		['name' => 'page#setFullWidth', 'url' => '/_api/{collectiveId}/_pages/{id}/fullWidth',
+			'verb' => 'PUT', 'requirements' => ['collectiveId' => '\d+', 'id' => '\d+']],
 		['name' => 'page#contentSearch', 'url' => '/_api/{collectiveId}/_pages/search',
 			'verb' => 'GET', 'requirements' => ['collectiveId' => '\d+', 'filterString' => '\s+']],
 		['name' => 'page#moveOrCopy', 'url' => '/_api/{collectiveId}/_pages/{id}',
--- a/lib/Db/Page.php
+++ b/lib/Db/Page.php
@@ -24,6 +24,8 @@ use OCP\AppFramework\Db\Entity;
  * @method void setEmoji(?string $value)
  * @method string getSubpageOrder()
  * @method void setSubpageOrder(string $value)
+ * @method bool getFullWidth()
+ * @method void setFullWidth(bool $value)
  * @method int|null getTrashTimestamp()
  * @method void setTrashTimestamp(?int $trashTimestamp)
  */
@@ -32,7 +34,12 @@ class Page extends Entity implements JsonSerializable {
 	protected ?string $lastUserId = null;
 	protected ?string $emoji = null;
 	protected ?string $subpageOrder = null;
+	protected ?bool $fullWidth = null;
 	protected ?int $trashTimestamp = null;
+
+	public function __construct() {
+		$this->addType('fullWidth', 'bool');
+	}
 
 	public function jsonSerialize(): array {
 		return [
@@ -41,6 +48,7 @@ class Page extends Entity implements JsonSerializable {
 			'lastUserId' => $this->lastUserId,
 			'emoji' => $this->emoji,
 			'subpageOrder' => json_decode($this->getSubpageOrder() ?? '[]', true, 512, JSON_THROW_ON_ERROR),
+			'fullWidth' => $this->fullWidth,
 			'trashTimestamp' => $this->trashTimestamp,
 		];
 	}
--- a/lib/Migration/Version021200Date20240725154232.php
+++ b/lib/Migration/Version021200Date20240725154232.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Collectives\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version021200Date20240725154232 extends SimpleMigrationStep {
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('collectives_pages');
+		if (!$table->hasColumn('full_width')) {
+			$table->addColumn('full_width', Types::BOOLEAN, [
+				'notnull' => false,
+				'default' => false,
+			]);
+
+			return $schema;
+		}
+
+		return null;
+	}
+}
--- a/lib/Model/PageInfo.php
+++ b/lib/Model/PageInfo.php
@@ -24,6 +24,7 @@ class PageInfo implements JsonSerializable {
 	private ?string $lastUserId = null;
 	private ?string $lastUserDisplayName = null;
 	private ?string $emoji = null;
+	private bool $fullWidth = false;
 	private ?string $subpageOrder = null;
 	private ?int $trashTimestamp = null;
 	private string $title;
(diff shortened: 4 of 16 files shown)


appinfo/routes.php
```diff
@@ -60,6 +60,8 @@
 			'verb' => 'POST', 'requirements' => ['collectiveId' => '\d+', 'parentId' => '\d+']],
 		['name' => 'page#touch', 'url' => '/_api/{collectiveId}/_pages/{id}/touch',
 			'verb' => 'GET', 'requirements' => ['collectiveId' => '\d+', 'id' => '\d+']],
+		['name' => 'page#setFullWidth', 'url' => '/_api/{collectiveId}/_pages/{id}/fullWidth',
```
Koc, line 63: Mates, WDYT about creating some generic "update" method to store small things like Full width/Emoji/etc?

Koc:
I've added video to prove that it works. Please review @juliushaertl @mejo- @max-nextcloud and run pipeline

I will work on tests after initial review

Koc:
rebased, fixed merge conflicts and disabled checkbox for users without edit permission. Please review

lib/Db/Page.php
```diff
@@ -32,15 +34,21 @@ class Page extends Entity implements JsonSerializable {
 	protected ?string $lastUserId = null;
 	protected ?string $emoji = null;
 	protected ?string $subpageOrder = null;
+	protected ?bool $fullWidth = null;
 	protected ?int $trashTimestamp = null;
 
+	public function __construct() {
+		$this->addType('fullWidth', 'bool');
```
mejo-, line 41: We don't use `addType` in the constructors so far and I don't think it's necessary with typing in the class attributes. But not exactly sure about this. What was the reasoning behind adding it for you?

lib/Migration/Version021200Date20240725154232.php
```diff
@@ -0,0 +1,35 @@
...
+
+class Version021200Date20240725154232 extends SimpleMigrationStep {
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('collectives_pages');
+		if (!$table->hasColumn('full_width')) {
+			$table->addColumn('full_width', Types::BOOLEAN, [
+				'notnull' => false,
```
mejo-, line 26: Wouldn't it be better to set `notnull`? Given that you provide a default, there's also no scenario when the value would be set to `NULL`, or am I missing something?

lib/Model/PageInfo.php
```diff
@@ -170,7 +180,7 @@ public function jsonSerialize(): array {
 	 * @throws InvalidPathException
 	 * @throws NotFoundException
 	 */
-	public function fromFile(File $file, int $parentId, ?string $lastUserId = null, ?string $lastUserDisplayName = null, ?string $emoji = null, ?string $subpageOrder = null): void {
+	public function fromFile(File $file, int $parentId, ?string $lastUserId = null, ?string $lastUserDisplayName = null, ?string $emoji = null, ?string $subpageOrder = null, ?bool $fullWidth = false): void {
```
mejo-, line 183:
```suggestion
	public function fromFile(File $file, int $parentId, ?string $lastUserId = null, ?string $lastUserDisplayName = null, ?string $emoji = null, ?string $subpageOrder = null, bool $fullWidth = false): void {
```

Given the default for `$fullWidth` is a `bool` value, there's no need to make it nullable.

mejo- (approved):
Thanks so much for working on this @Koc :heart:

Code changes look good to me, I only have minor comments.

I'm in the process of migrating the frontend from vuex to pinia (#1398), so the frontend part will need some refactoring once this got merged. Sorry for the extra work :grimacing:

mejo- (changes requested):
Sorry, meant to request changes :wink:

lib/Db/Page.php
Koc: It not works without this change :) Postgres returns some error, like it can't set 1 to field which expects bool

Koc:
@mejo- all fine, I will do everything. No way to wait this changes before migration to Pinia? :pray:

appinfo/routes.php
Koc: Should I remove old api endpoint completely or just deprecate it?

lib/Db/Page.php
juliushaertl: That should be fine and the way go https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/database.html#entities to ensure that database column types are properly mapped

Those types will be used then in the mapper https://github.com/nextcloud/server/blob/2fb0ca9cf7929b3527f2068e19a690c371191b72/lib/public/AppFramework/Db/QBMapper.php#L213-L236

These days it would actually be nicer to inherit the type from the type hints but that's another story for improvement in server.

lib/Migration/Version021200Date20240725154232.php
juliushaertl: https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_22.html#boolean-database-columns

> Since Oracle can not store booleans in a non-nullable boolean column Nextcloud doesn’t support non-null boolean columns anymore. Apps have to migrate their schema to nullable boolean columns.

mejo-:
> @mejo- all fine, I will do everything. No way to wait this changes before migration to Pinia? 🙏

Sorry, pinia PR just got merged :grimacing:

lib/Model/PageInfo.php
Koc: done :heavy_check_mark:

lib/Db/Page.php
Koc: so, I guess we can keep it as is

appinfo/routes.php
Koc: I've created separate issue for that https://github.com/nextcloud/collectives/issues/1415. Better to not do everything in single PR.

Koc:
@mejo- @juliushaertl mates, how can I understand what thee root cause of Behat tests failure? How can I run this tests locally to debug failure?

mejo-:
> @mejo- @juliushaertl mates, how can I understand what thee root cause of Behat tests failure? How can I run this tests locally to debug failure?

@Koc I'm pretty sure it's unrelated, but let's see. Given that your local Nextcloud instance is available under `https://nextcloud.local`, you can run the behat tests with `make test-php-integration`.

tests/Integration/features/bootstrap/FeatureContext.php
```diff
@@ -534,6 +534,26 @@ public function userSetsPageEmoji(string $user, string $page, string $emoji, str
...
+	 * @When user :user sets full width for page :page to :fullWidth in :collective
+	 * @When user :user :fails to set full width for page :page to :fullWidth in :collective
+	 *
+	 * @throws GuzzleException
+	 */
+	public function userSetsFullWidth(string $user, string $page, string $fullWidth, string $collective, ?string $fail = null): void {
+		$this->setCurrentUser($user);
+		$collectiveId = $this->collectiveIdByName($collective);
+		$pageId = $this->pageIdByName($collectiveId, $page);
+		$formData = new TableNode([['emoji', $fullWidth === "true"]]);
```
mejo-, line 547:
```suggestion
		$formData = new TableNode([['fullWidth', (bool)$fullWidth]]);
```

tests/Integration/features/bootstrap/FeatureContext.php
```diff
@@ -534,6 +534,26 @@ public function userSetsPageEmoji(string $user, string $page, string $emoji, str
...
+		$this->sendRequest('PUT', '/apps/collectives/_api/' . $collectiveId . '/_pages/' . $pageId . '/fullWidth', $formData);
+		if ($fail === "fails") {
+			$this->assertStatusCode(403);
+		} else {
+			$this->assertStatusCode(200);
+			$this->assertPageKeyValue($pageId, 'fullWidth', $fullWidth);
```
mejo-, line 553:
```suggestion
			$this->assertPageKeyValue($pageId, 'isFullWidth', (bool)$fullWidth);
```

mejo- (changes requested):
Only some minor comments and suggestions to fix the behat tests left. Really nice that you already provided behat tests :heart_hands:

Cypress tests are already there and sufficient. The comment in https://github.com/nextcloud/collectives/blob/63eba4906b8e0a3f52908f80c086fa0c0d866bd7/cypress/e2e/pages.spec.js#L262 could use an update though (just remove the "with browser storage" part).

src/stores/pages.js
```diff
@@ -5,7 +5,6 @@
 
 import { defineStore } from 'pinia'
 import { set } from 'vue'
-import { getBuilder } from '@nextcloud/browser-storage'
```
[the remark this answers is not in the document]
Koc: indeed

tests/Integration/features/bootstrap/FeatureContext.php
Koc: ah, good catch, thanx!

mejo- (approved):
Really nice, thanks for your work @Koc :heart:

PR #1393, opened 2024-07-29
ci(cypress): Use cypress-split for running tests in parallel

mejo-:
### 🏁 Checklist

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -142,7 +142,9 @@ jobs:
         uses: buildjet/cache/save@e376f15c6ec6dc595375c78633174c7e5f92dc0e # v3
         with:
           key: cypress-context-${{ matrix.server-versions }}-${{ github.run_id }}
-          path: ./
+          path: |
+            ./
+            /home/runner/.cache/Cypress
 
   cypress:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/cypress-e2e.yml
```diff
@@ -233,37 +229,37 @@ jobs:
...
         with:
-          name: Screenshots.${{ matrix.server-versions }}-${{ matrix.containers }}
+          name: screenshots_${{ matrix.server-versions }}_${{ matrix.containers }}
           path: apps/${{ env.APP_NAME }}/cypress/screenshots/
           retention-days: 5
 
       - name: Upload NC logs
         uses: actions/upload-artifact@0b2256b8c012f0828dc542b3febcab082c67f72b # v4.3.4
-        if: always()
+        if: failure()
```
max-nextcloud, line 260: Is there a disadvantage to always uploading them?

Most of the time I think the logs should be empty, right? Sometimes i found it useful to check for that and investigate errors in the logs even if the tests passed.

max-nextcloud (approved):

.github/workflows/cypress-e2e.yml
mejo-: I changed this because I never look into logs when tests succeed, and it's like this in Text. Don't have strong feelings about it though :shrug:

PR #1394, opened 2024-07-29
controller: Migrate to attributes and add rate limits for public pages

mejo-:
### 🏁 Checklist

### DIFF
--- a/lib/Controller/CollectiveController.php
+++ b/lib/Controller/CollectiveController.php
@@ -15,6 +15,7 @@ use OCA\Collectives\Db\Collective;
 use OCA\Collectives\Fs\NodeHelper;
 use OCA\Collectives\Service\CollectiveService;
 use OCP\AppFramework\Controller;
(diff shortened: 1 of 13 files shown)


max-nextcloud (approved):
Looks good to me.
I'm somewhat unuser about the page get endpoint - it will easily exceed the limit if we fetch all info for pages in a collective via this endpoint. But to my understanding we only fetch it for one page when it's opened, right?

mejo-:
>  I'm somewhat unuser about the page get endpoint - it will easily exceed the limit if we fetch all info for pages in a collective via this endpoint. But to my understanding we only fetch it for one page when it's opened, right?

To be honest I don't think we use the get page endpoint at all in the frontend :upside_down_face:. We have `getPage()` defined in `src/apis/collectives/pages.js` and `GET_PAGE` in `src/store/pages.js`. But `GET_PAGE` doesn't get used anywhere.

PR #1403, opened 2024-08-05
perf: Optimize getPagesFromFolder to avoid requests for files that already get obtained through directory listing

juliushaertl:
I'm cleaning up some local stashed changes, opening a PR to not forget about this one, still needs proper testing and performance measurements

This one targets optimizing the initial page load on `PageController::index` (currently quite slow with P95 of 5.80s on our instance) where a list of all pages are loaded. The current logic:

- get index file (checks for a file that we could reuse the results from the getDirectoryListing to obtain it)
- check if folder has subpages (calls getDirectoryListing)
- get directory content (calls getDirectoryListing again)

New logic would be to only get the directory listing once and reuse it for other checks

Screenshot 2024-08-05 at 21 55 57

### 🏁 Checklist

### DIFF
--- a/lib/Service/PageService.php
+++ b/lib/Service/PageService.php
@@ -382,27 +382,31 @@ class PageService {
 	 * @throws NotPermittedException
 	 */
 	public function getPagesFromFolder(int $collectiveId, Folder $folder, string $userId, bool $recurse = false): array {
-		// Find index page or create it if we have subpages, but it doesn't exist
-		try {
-			$indexPage = $this->getPageByFile($this->getIndexPageFile($folder), $folder);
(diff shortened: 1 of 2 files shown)


juliushaertl:
Blackfire comparison with the company handbook as sample data: https://blackfire.io/profiles/compare/0d4caf55-5a57-4156-9694-a6a9ab4b5e59/graph

While the timing results locally are a bit flaky the number of saved method calls should be enough to proof that it is worth.

mejo- (approved):
Really nice optimization, and the blackfire call graph speaks for itself :guitar:

Issue #1397, opened 2024-07-30
Design review notes 2024-07-30


juliushaertl:
**Bold items are considered high priority**

- [x] New collective dialog
  - [x] Name too short should be left aligned, color max contrast, debounce 
  - [x] New collective should move from NcModal to NcDialog
  - [x] Title on the empty content should not have a title on the icon
  - [x] **In the user picker the text is not fully centered, check mark is not fully on the right**
  - [x] Own user should be primary highlight
  - [x] Can we show and suggest some users, like in talk
- [x] Manage members dialog
  - [x] Has two scrollbars
  - [x] Three dots is not fully right aligned
  - [x] Remove member should be red (maybe as a standard)
  - [x] (admin) and moderator text can be regular instead of light
- [x] **Title placeholder should be same size as the text and max contrast without opacity**
- [x] **Emoji in the page title header is not full opacity**
- [ ] **Maybe make the Test collective rather bold?**
- [x] **Make the collective title / page title regular h2 style**
- [x] Navigation icon for collective (emoji) is too big, should be 20px
- [ ] Team overview button should open in a new tab
- [x] "New Page" and "Team Overview" should have lower case letter on the second word
- [x] Sharing sidebar
  - [x] **Check error on daily**  @[Julius Härtl](mention://user/julius)  
  - [x] Jumping when clicking the button
- [x] Attachments sidebar
  - [x] Info text can replace the empty content message
- [x] Page menu (top right)
  - [x] **Open page sidebar can be removed (quick fix)**
  - [x] Full witdth checkbox out of place, mabye using the old component
- [ ] Collectives settings
  - [ ] Too much whitespace before headings (section element)

mejo-:
Thanks for collecting the feedback. While addressing the items, a few questions came up @juliushaertl @jancborchardt @nimishavijay:

>  Maybe make the Test collective rather bold?

Some time back we purposely changed the collective title from bold to a slightly bigger font in the pagelist due to negative user feedback about the bold title. Users thought it would mean "unread" and described the urge to click on it to mark it read. So I don't think it's a good idea to change this to bold again.

> Make the collective title / page title regular h2 style

What's the background here? Currently the whole titleform is inside a `h1` element.

> Sharing sidebar -> Check error on daily @Julius Härtl

What does this mean?

> Attachments sidebar -> Info text can replace the empty content message

What's meant here? Here's how an empty attachments sidebar currently looks. What can replace what?

[...]

PR #1407, opened 2024-08-07
Address Nextcloud 30 design review comments

mejo-:
### 📝 Summary

* Resolves: #1397 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik
---|---
grafik | grafik
---|---
grafik | grafik
---|---
grafik | grafik
--- | ---
grafik | grafik
---|---
grafik | grafik
---|---
grafik | grafik
---|---
grafik | grafik
---|---
grafik | grafik
--- | ---
grafik | grafik

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/collective-share.spec.js
+++ b/cypress/e2e/collective-share.spec.js
@@ -116,7 +116,7 @@ describe('Collective Share', function() {
 		it('Opening unshared collective fails', function() {
 			cy.logout()
 			cy.visit(shareUrl, { failOnStatusCode: false })
-			cy.get('.body-login-container').contains(/(File|Page) not found/)
+			cy.get('.body-login-container').contains(/(File|Page|Share) not found/)
 		})
 	})
 })
--- a/cypress/e2e/page-list.spec.js
+++ b/cypress/e2e/page-list.spec.js
@@ -82,7 +82,7 @@ describe('Page list', function() {
 				.click()
 			cy.get('.picker-move-buttons button .arrow-down-icon')
 				.click()
-			cy.get('.picker-buttons button')
+			cy.get('.dialog__actions button')
 				.contains('Move page here')
 				.click()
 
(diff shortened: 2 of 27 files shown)


jancborchardt (changes requested):
Nice! Some feedback:
- [ ] Modal first step:
  - [ ] The emptycontent icon seems placed too low, not vertically centered
  - [ ] "Cancel" button was fine on the left and without icon. Placing it next to the "Add members" button can lead to misclicks, and giving it an icon gives it too much importance.
- [ ] Add members step:
  - [ ] The emptycontent text seems a bit superfluous, duplicate from the search label?
  - [ ] Same here with the "Back" button as with the "Cancel" button before
- [ ] Pages: We noted during the design review that page titles would look better in bold as proper h2 heading styles.
- [ ] Attachments sidebar: This seems to have a misunderstanding :) What I meant was to adjust it so the bold text stays "No attachments available" (otherwise this is too much text for a heading). The current text "If the page has attachments, they will be listed here" is replaced with the text of the infobox as it is more informative, and the infobox style is removed as it is an unnecessary different style.

Thank you! :)

mejo-:
>  "Cancel" button was fine on the left and without icon. Placing it next to the "Add members" button can lead to misclicks, and giving it an icon gives it too much importance.

This is how buttons are aligned in `NcDialog`. I could hack this with a `:deep()` selector, but should I also do this for the second step of the new collectives modal than as well?

mejo-:
> Add members step:
> - The emptycontent text seems a bit superfluous, duplicate from the search label?

What do you mean here? In the second step of the new collective modal, there's no empty content displayed anymore, because it loads a list of potential members to add straight away :thinking:

mejo-:
Thanks @jancborchardt, addressed all your follow-up comments with one further question and updated the screenshots. Would you mind to review again?

jancborchardt (approved):
Awesome @mejo-, looks really great! :)

juliushaertl (approved):

Issue #1461, opened 2024-09-04
Search highlighting does not persist when switching pages


elzody:
**Describe the bug**
When executing a search, the matching text is highlighted in the current page. This does not persist when switching pages, meaning the highlighting is lost.

**To Reproduce**
1. Go to any page in any collective
2. Type some text to find
3. Make sure some text is found and highlighted
4. Switch to another page in the list
5. The search dialog is still at the bottom, but no matching text is highlighted

**Expected behavior**
The expected behavior is that the matching text will still be highlighted when switching pages (if there is still some text being searched)

PR #1462, opened 2024-09-04
fix: search highlighting persists between pages

elzody:
### 📝 Summary

* Resolves: #1461 

Previously, the search highlighting did not persist when changing pages. This PR uses changes in the editor content to re-execute the search, meaning it should now re-apply the search highlighting properly when switching pages.

### 🏁 Checklist


mejo-:
@elzody no need to backport, we don't have separate branches for supported Nextcloud versions in Collectives :wink:

src/mixins/editorMixin.js
```diff
@@ -89,6 +89,10 @@ export default {
 				this.reader?.setSearchQuery(this.searchQuery, newValue)
 			}
 		},
+		'editorContent'() {
```
mejo-, line 92: I wonder whether it's really required to watch for changes to `editorContent`. Wouldn't it be enough to watch for `currentPage.id`?

mejo- (changes requested):

PR #1476, opened 2024-09-11
fix(mount): public shares always get permissions of a simple user

mejo-:
### 🏁 Checklist


max-nextcloud (approved):
only had a quick look. Thanks for adding tests. Looks convincing to me.

PR #1491, opened 2024-09-18
fix(CollectiveStorage): adjust `getOwner` to NC31

mejo-:
### 🏁 Checklist

### DIFF
--- a/lib/Mount/CollectiveStorage.php
+++ b/lib/Mount/CollectiveStorage.php
@@ -38,9 +38,9 @@ class CollectiveStorage extends Wrapper {
 	/**
 	 * @param string $path
 	 *
-	 * @return false|string
+	 * @return string|false
 	 */
-	public function getOwner($path) {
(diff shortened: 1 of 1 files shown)


max-nextcloud:
Wait... that change looks like it's only adding a return type to the function.

Would this break with NC <= 30 because PHP does not yet support that syntax

max-nextcloud (commented):
Not sure if this actually has the full change.

mejo-:
> Wait... that change looks like it's only adding a return type to the function.
> 
> Would this break with NC <= 30 because PHP does not yet support that syntax

Exactly, it only adds the return type to be in line with the parent class on Nextcloud 31. I first thought it would then be incompatible with the parent classes on Nextcloud 30 and below, but manual testing and CI both proved me wrong. So should be good to be merged :blush:

provokateurin (approved):
Sorry for this :see_no_evil:

blizzz (approved):

PR #1505, opened 2024-09-25
fix(member): close actions menu when scrolling members

mejo-:
### 🏁 Checklist

### DIFF
--- a/src/components/Member/Member.vue
+++ b/src/components/Member/Member.vue
@@ -41,6 +41,7 @@
 		<!-- Action menu -->
 		<NcActions v-else-if="currentUserIsAdmin && !isSearched && !isCurrentUser"
 			:force-menu="true"
+			:open.sync="showActionMenu"
 			class="member-row__actions">
 			<NcActionButton v-if="!isAdmin"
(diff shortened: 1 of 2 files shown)


nickvergessen (approved):
At least prevents the menu from being visible while scrolled away.
But could also set the container/parent (@Antreesy might be able to point you into the right places)

Antreesy:
container is needed to mount popover in specific place (useful for popover inside of modal to keep correct order): https://floating-vue.starpad.dev/api/#container

boundary (boundaries-element in NcActions) fits better here, but i haven't tested it yet with scrolling lists: https://floating-vue.starpad.dev/api/#boundary

mejo-:
> boundary (boundaries-element in NcActions) fits better here, but i haven't tested it yet with scrolling lists: https://floating-vue.starpad.dev/api/#boundary

That would probably be `$parent.$parent.refs.memberPickerList` (the `div.member-picker-list` element) in this case as it's the one that has the scroll container. Unfortunately passing this as `:boundaries-element` property to `NcActions` doesn't solve the problem at hand.

Issue #1494, opened 2024-09-20
Could not fetch pages (Readme.md)


avsob:
**Describe the bug**
Error "Could not fetch pages" on Collective page.

**To Reproduce**
1. Create new Collective "Group1" (or use existing one) 
2. Go to Files app and create folder "Folder1". 
3. Into "Folder1" create folder "Folder2"
4. Into "Folder2" create file "file.md"
5. Go to Collective app
6. On collective "Group1" have error "Could not fetch pages: 
/<user.name>/files/Collectives/Group1/Folder1/Folder2/Readme.md"
And "Group1" page not loaded.

**Expected behavior**
No error.  The group page loads fine

**Screenshots**
![2024_09_20_Error_Could not fetch pages](https://github.com/user-attachments/assets/94cc7bcb-32f0-4458-ae90-fe950c99dc1f)

**Server details:**
- Collectives app version: 2.14.3
- Nextcloud version: 29.0.7
- PHP Version: 8.2
- Database: PostgreSQL 14

**Client details:**
 - OS: Windows 10
 - Browser: Firefox v130, Chrome v129 and Edge v128
 - Device: desktop

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
Errors are not recorded in the nextcloud.log file

</details>

Issue #943, opened 2023-10-15
Cannot access collection anymore


rrrnld:
**Describe the bug**
I cannot open my collection anymore. Instead, I always see: "Page not found: {page}", with no option to change to a different page

**To Reproduce**

1. Create new page
2. Click on 'Open in files'
3. Create a folder with the same name

Now, no matter what I do, I end up with the message above. The files are still on the disk, but I can't open them in the Collectives app anymore.

**Expected behavior**
I wanted to copy a README.md from one folder to another.

**Screenshots**

**Client details:**
 - OS: [e.g. iOS] Linux
 - Browser: [e.g. chrome, safari] Firefox
 - Version: [e.g. 24] 118
 - Device: [e.g. iPhone6, desktop] Laptop

<details>
<summary>Server details</summary>

**Collectives app version:** (see Nextcloud apps page) 2.8.2

**Operating system:** Linux 5.15.0-78-generic

**Web server:** nginx version: nginx/1.18.0

**Database:**mysql  Ver 15.1 Distrib 10.6.12-MariaDB

**PHP version:** 8.2.8

**Nextcloud version:** (see Nextcloud admin page) 27.1.2

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
[...]

PR #1506, opened 2024-09-25
fix(PageService): Handle errors with missing index files

mejo-:
### 📝 Summary

* fix(PageService): Create landing page if missing
* fix(PageService): Ignore folders if parent folder has no index page
* Fixes: #943
* Fixes: #1494

### 🏁 Checklist

### DIFF
--- a/lib/Service/PageService.php
+++ b/lib/Service/PageService.php
@@ -396,7 +396,12 @@ class PageService {
 		// Add subpages and recurse over subfolders
 		foreach ($folder->getDirectoryListing() as $node) {
 			if ($node instanceof File && NodeHelper::isPage($node)) {
-				$page = $this->getPageByFile($node, $folder);
+				try {
+					$page = $this->getPageByFile($node, $folder);
+				} catch (NotFoundException) {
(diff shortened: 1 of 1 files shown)


blizzz (approved):

lib/Service/PageService.php
```diff
@@ -389,14 +389,19 @@ private function getIndexPageFile(Folder $folder): File {
...
 		$indexPage = null;
 
 		// Add subpages and recurse over subfolders
 		foreach ($folder->getDirectoryListing() as $node) {
 			if ($node instanceof File && NodeHelper::isPage($node)) {
-				$page = $this->getPageByFile($node, $folder);
+				try {
+					$page = $this->getPageByFile($node, $folder);
+				} catch (NotFoundException) {
+					// If parent folder doesn't have an index page, it throws NotFoundException. Let's ignore it.
```
blizzz, line 402: Should the NotFoundExceptions then removed from the doc block?

mejo-: Other code lines in this function can still throw `NotFoundException`.

PR #1547, opened 2024-10-23
fix(deps): Remove dev dependency on nextcloud/ocp

mejo-:
Only needed for psalm checks, which got moved to its own vendor-bin environment.

This reverts commit df190af3720be2073aaa80a5868a853733dbf3ba.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/behat-sqlite-encryption.yml
+++ b/.github/workflows/behat-sqlite-encryption.yml
@@ -90,10 +90,7 @@ jobs:
 
       - name: Install composer dependencies
         working-directory: apps/${{ env.APP_NAME }}
-        run: |
-          composer i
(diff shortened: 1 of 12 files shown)


max-nextcloud:
Maybe add a `.git-blame-ignore-refs` file and add the code style fixes commit to it like in https://github.com/nextcloud-libraries/eslint-config/pull/854/files

mejo-:
> Maybe add a .git-blame-ignore-refs file and add the code style fixes commit to it like in https://github.com/nextcloud-libraries/eslint-config/pull/854/files

Interesting, didn't know about this feature. But it feels a bit off to start this now with a rather small code style commit. Maybe instead do a separate PR that adds all recent code style commits to this file?

juliusknorr (approved):

PR #1551, opened 2024-10-27
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
         "ext-dom": "*",
         "bamarni/composer-bin-plugin": "^1.8",
         "guzzlehttp/guzzle": "^7.8",
+        "nextcloud/ocp": "dev-stable27",
         "rector/rector": "^1.2.0"
     },
     "scripts": {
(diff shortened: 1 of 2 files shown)


max-nextcloud:
behat tests fail consistently with current master branch for server. Looks like something changed there.
Restarted mysql upgrade test as that seemed to have failed due to mysql failing to start.

max-nextcloud:
grafik

We don't seem to store the nextcloud.log though it seems - so no way from the failed CI run to tell what happened.

mejo-:
I removed `nextcloud/ocp` from `composer.json` in https://github.com/nextcloud/collectives/pull/1547, because it's only needed by psalm which got moved to a dedicated `vendor-bin/psalm` environment. Thus is version bump is obsolete. It causes problems because `nextcloud/ocp` requires a newer `psr/log` version that conflicts with some `behat` dependencies. That's why Julius excluded `nextcloud/ocp` in https://github.com/nextcloud/collectives/pull/1522/commits/df190af3720be2073aaa80a5868a853733dbf3ba before, which I reverted due to being obsolete in #1547

PR #1561, opened 2024-10-30
fix(css): Several print style fixes

mejo-:
* Fixed alignment of narrow page content
* No cropped page content on the right side
* No page break between page title and content when printing single page
* Print page content after first page when printing single page

### 🏁 Checklist

### DIFF
--- a/src/Collectives.vue
+++ b/src/Collectives.vue
@@ -108,12 +108,11 @@ export default {
 	overflow-y: auto;
 }
 
-@page {
-	size: auto;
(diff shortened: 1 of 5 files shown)


mejo-:
> Is there a visible before and after?

Unfortunately hard to screenshot because only visible in print view. That's also why I listed exactly what this fixes.

mejo-:
> Uh, where's the list please?

Haha sorry, I only added it later to the commit message. Copied it now to the PR description above.

PR #1564, opened 2024-11-04
fix: Preserve anchors in links that rendered as widget

Koc:
### 📝 Summary
1. Create long page with anchor in the bottom
2. Create another page and insert link to first page with anchor
3. Open preview widget

Expected result: anchor is preserved
Actual result: anchor is missing

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist


src/mixins/editorMixin.js
```diff
@@ -128,9 +128,11 @@ export default {
...
 					this.reader.setSearchQuery && this.reader.setSearchQuery(this.searchQuery, this.matchAll)
 					if (document.location.hash) {
-						// scroll to the corresponding header if the page was loaded with a hash
-						const element = document.querySelector(`[href="${document.location.hash}"]`)
-						element?.click()
+						window.setTimeout(() => {
+							// scroll to the corresponding header if the page was loaded with a hash
+							const element = document.querySelector(`[href="${document.location.hash}"]`)
+							element?.click()
+						}, 1000)
```
Koc, line 135: This looks ugly but the problem is that `element` is null here once we click on link in widget. I've try to use `this.$nextTick()` instead but still null

mejo-: The problem is that `reader` is initiated before the dav content is fetched [in `mounted()` of `TextEditor.vue`](https://github.com/nextcloud/collectives/blob/9729f643247dfd7be117565a7a58576c0a3729b1/src/components/Page/TextEditor.vue#L99-L101), so the `loaded` callback gets fired before the content is set.

A better place for scrolling to the anchor link is the [end of `initEditMode()` in `TextEditor.vue`](https://github.com/nextcloud/collectives/blob/9729f643247dfd7be117565a7a58576c0a3729b1/src/components/Page/TextEditor.vue#L150).

mejo- (approved):
Thanks a lot for tackling this @Koc, much appreciated. I pushed a small improvement, see my comment. I hope you don't mind.

src/mixins/editorMixin.js
Koc: awesome, thank you for investigation! :handshake:

PR #1568, opened 2024-11-04
chore: Remove Github workflow to update `nextcloud/ocp`

mejo-:
We no longer have `nextcloud/ocp` in our dev dependencies, it got moved to a dedicated psalm vendor-bin environment.


juliusknorr:
No auto updating then for the new one?

mejo-:
I agree that we should find away to auto-update the one in `vendor-bin/psalm`, but the current workflow doesn't do this either, it only creates PRs that break the behat tests :zany_face:

max-nextcloud (approved):

PR #1574, opened 2024-11-06
fix(PageService): Create index page if folder or subfolders have pages

mejo-:
Fixes the scenario where a folder is missing the index page but one of its subfolders contains pages.

### 🏁 Checklist

### DIFF
--- a/lib/Service/PageService.php
+++ b/lib/Service/PageService.php
@@ -390,36 +390,62 @@ class PageService {
 	 * @throws NotPermittedException
 	 */
 	public function getPagesFromFolder(int $collectiveId, Folder $folder, string $userId, bool $recurse = false, bool $forceIndex = false): array {
-		$pageInfos = [];
-		$indexPage = null;
+		$subPageInfos = [];
+		$folderNodes = $folder->getDirectoryListing();
 
-		// Add subpages and recurse over subfolders
-		foreach ($folder->getDirectoryListing() as $node) {
-			if ($node instanceof File && NodeHelper::isPage($node)) {
+		$hasPages = false;
+		$pageFiles = [];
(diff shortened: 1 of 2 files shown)


lib/Service/PageService.php
```diff
@@ -391,35 +391,54 @@ private function getIndexPageFile(Folder $folder): File {
...
+			foreach ($subfolders as $subfolder) {
+				array_push($pageInfos, ...$this->getPagesFromFolder($collectiveId, $subfolder, $userId, true));
 			}
 		}
 
-		if (!$indexPage) {
-			if (!$forceIndex && count($pageInfos) === 0) {
+		$pageFiles = array_filter($folderNodes, static function (Node $node) {
+			return $node instanceof File && NodeHelper::isPage($node);
+		});
```
blizzz, line 408: There is a potentially duplicated iteration over the same array. The array_ functions also copy the whole array into RAM, which might not be desired on big collections. Maybe classic foreach is the better option?

lib/Service/PageService.php
```diff
@@ -391,35 +391,54 @@ private function getIndexPageFile(Folder $folder): File {
...
+
+		// Create missing index page if folder or subfolders have page files (or forceIndex)
+		$hasIndexFile = (bool)array_filter($folderNodes, static function (Node $node) {
+			return $node instanceof File && NodeHelper::isIndexPage($node);
+		});
```
blizzz, line 413: Okay, up to three times. bool-casting an array? array_reduce would be more straight forward, but this can be embedded in the same single foreach, no?

blizzz (commented):
remarks

blizzz (approved):

PR #1617, opened 2024-12-19
Revert "test(cypress): Fix flaky template test by disabling a check"

mejo-:
This reverts commit de20a9fb94331dce234102faec1cb904bbdd667e.

### 🚧 TODO

- [ ] Find out why in Cypress new pages don't open in edit mode when created from a template. Manually tests work as expected, but in CI this fails in the majority of test runs. I suspect a race condition that only happens because Cypress is faster than a human.

### 🏁 Checklist


mejo-:
This is obsolete now after the templates feature rework.

PR #1625, opened 2025-01-03
ci: Hotfix to make login work as long as we are on older cypress version

Koc:
This is replication of https://github.com/nextcloud/text/pull/6819 to Collectives repository in order to fix Cypress tests


juliusknorr (approved):

juliusknorr:
Thank you very much

Issue #1619, opened 2024-12-20
Sharding compatibility


juliusknorr:
On daily we have the following error, trace looks like some join is not working anymore due to the sharding enabled there:

```
{
  "reqId": "OV4plHan5yZytg1DqGS8",
  "level": 3,
  "time": "2024-12-20T08:44:05+00:00",
  "remoteAddr": "95.114.120.194",
  "user": "julius",
  "app": "index",
  "method": "GET",
  "url": "/apps/theming/icon/settings?v=a33f1d43",
  "message": "An exception occurred while executing a query: SQLSTATE[42703]: Undefined column: 7 ERROR:  column \"fileid\" does not exist\nLINE 1: SELECT \"co\".\"id\" as \"folder_id\", \"fileid\", \"storage\", \"path\"...\n                                         ^",
  "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:133.0) Gecko/20100101 Firefox/133.0",
  "version": "31.0.0.6",
  "exception": {
    "Exception": "Doctrine\\DBAL\\Exception\\InvalidFieldNameException",
    "Message": "An exception occurred while executing a query: SQLSTATE[42703]: Undefined column: 7 ERROR:  column \"fileid\" does not exist\nLINE 1: SELECT \"co\".\"id\" as \"folder_id\", \"fileid\", \"storage\", \"path\"...\n                                         ^",
    "Code": 7,
    "Trace": [
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1939,
        "function": "convert",
        "class": "Doctrine\\DBAL\\Driver\\API\\PostgreSQL\\ExceptionConverter",
        "type": "->"
      },
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1881,
        "function": "handleDriverException",
        "class": "Doctrine\\DBAL\\Connection",
        "type": "->"
      },
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1106,
        "function": "convertExceptionDuringQuery",
        "class": "Doctrine\\DBAL\\Connection",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/DB/Connection.php",
        "line": 415,
        "function": "executeQuery",
        "class": "Doctrine\\DBAL\\Connection",
        "type": "->"
      },
      {
[...]

PR #1630, opened 2025-01-10
fix: hint table for columns for sharded queries

juliusknorr:
Fix https://github.com/nextcloud/collectives/issues/1619

Based on https://github.com/nextcloud/groupfolders/pull/3483

For queries to the filecache we need to provide the sharding key (storage id) so the sharding logic in server can put the query to the right shard. https://github.com/nextcloud/collectives/pull/1630/files#diff-070d4f3f410c3f2fe7c8c4ab40af4884d62b754f28350d2bb0ebc4fc3f846117R208 is the most relevant change

CI run in https://github.com/nextcloud/collectives/pull/1634
Failure shown in [nextcloud/collectives/actions/runs/12777873556/job/35619586742?pr=1634](https://github.com/nextcloud/collectives/actions/runs/12777873556/job/35619586742?pr=1634)

### DIFF
--- a/lib/Mount/CollectiveFolderManager.php
+++ b/lib/Mount/CollectiveFolderManager.php
@@ -37,7 +37,7 @@ class CollectiveFolderManager {
 	private const SUFFIX = '.md';
 
 	private ?string $rootPath = null;
-	private ?int $rootFolderStorageId = null;
+	private ?int $rootFolderId = null;
 
 	public function __construct(
 		private IRootFolder $rootFolder,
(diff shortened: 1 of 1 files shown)


juliusknorr:
@icewind1991 Do we have some docs on how to setup sharding and anything on how it works concept wise?

max-nextcloud:
Thanks for looking into this @juliusknorr 

As there are no instructions for testing i will compare with the groupfolders code you linked to and perform a general code check. Hope that works for you.

juliusknorr:
Sorry forgot to update, we tried on daily, but didn't work, waiting for feedback from @icewind1991 first. now

lib/Mount/CollectiveFolderManager.php
```diff
@@ -149,13 +150,13 @@ private function getJailPath(int $folderId): string {
...
 
-			$qb->select('fileid')
-				->from('filecache')
-				->where($qb->expr()->eq('storage', $qb->createNamedParameter($this->getRootFolder()->getStorage()->getCache()->getNumericStorageId())))
+			$qb->select('f.fileid')
+				->from('filecache', 'f')
+				->where($qb->expr()->eq('storage', $qb->createNamedParameter($this->getRootFolderStorageId())))
 				->andWhere($qb->expr()->eq('path_hash', $qb->createNamedParameter(md5($this->getRootPath()))));
 
 			$this->rootFolderStorageId = (int)$qb->execute()->fetchColumn();
```
max-nextcloud, line 162: Please rename this instance variable to `$this->folderRootId` to avoid confusion.

max-nextcloud (approved):

lib/Mount/CollectiveFolderManager.php
```diff
@@ -194,14 +199,13 @@ public function getLandingPagePath(string $path, string $lang): string {
...
 			->from('collectives', 'co')
-			->leftJoin('co', 'filecache', 'fc', $qb->expr()->andX(
-				// concat with empty string to work around missing cast to string
-				$qb->expr()->eq('fc.name', $qb->func()->concat('co.id', $qb->expr()->literal(''))),
-				$qb->expr()->eq('parent', $qb->createNamedParameter($this->getRootFolderStorageId()))))
-			->where($qb->expr()->eq('co.id', $qb->createNamedParameter($id)));
+			->leftJoin('co', 'filecache', 'fc', $qb->expr()->eq('fc.name', $qb->expr()->castColumn('co.id', IQueryBuilder::PARAM_STR)))
+			->where($qb->expr()->eq('co.id', $qb->createNamedParameter($id)))
+			->andWhere($qb->expr()->eq('fc.parent', $qb->createNamedParameter($this->getFolderRootId())))
+			->andWhere($qb->expr()->eq('fc.storage', $qb->createNamedParameter($this->getRootFolderStorageId(), IQueryBuilder::PARAM_INT)));
```
blizzz, line 208: cannot judge whether specifying the storage id is safe here. Could there be, for example, incoming shares?

lib/Mount/CollectiveFolderManager.php
juliusknorr: Used `getRootFolderId` / `$this->rootFolderId` now as it is in line with the new `getRootFolderStorageId`

blizzz (approved):

PR #1634, opened 2025-01-14
ci: Run tests with sharding

juliusknorr:
Draft to run tests on db sharding


juliusknorr:
I'll see if we can somehow add this to the existing action, closing this one for now

Issue #1469, opened 2024-09-09
Mentions & Notifications


kaffeeringe:
**Is your feature request related to a problem? Please describe.**
I use collectives for meeting minutes. I prepare each meeting with agenda and I add all people that shoud attend with an @-mention to the list of attendees. So everybody gets a notification that there is a new agenda and they are able to prepare for the meeting. 

But the notification doesn't link to the page in the collective but instead to the file in the filesystem. That is confusing.

**Describe the solution you'd like**
It would be netter if the notification linked to the page in the collective.

**Describe alternatives you've considered**
They could use the files. But the notification doesn't open the file. It only links to the folder where the files are sorted by title or whatever they set. They don't even really know which file to open.

Aveyron-RetD:
That's why no one participate in my collective /-)
Yes it could be done  like how spreed does it show it in the discussion. Show it in collectives

juliusknorr:
I think this should be rather easy to implement, we can do this in text.

The URL is generated in https://github.com/nextcloud/text/blob/ce13589f65505ad8b867e6c93bebdc275d0daaed/lib/Notification/Notifier.php#L76 so we should check if the file is within a collectives app mountpoint and use the collective route instead.

juliusknorr:
https://github.com/nextcloud/text/pull/6923
https://github.com/nextcloud/collectives/pull/1675

juliusknorr:
Took an attempt to implement since I've been navigating around in the code areas anyways

PR #1675, opened 2025-02-27
feat: Implement a custom notification message for mentions

juliusknorr:
Requires https://github.com/nextcloud/text/pull/6923

Fixes https://github.com/nextcloud/collectives/issues/1469

This makes sure that for mentions in collectives we have a proper notification message and redirect the user to the collective page instead of the file.

Screenshot 2025-02-27 at 17 00 41

### 🏁 Checklist

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -91,6 +91,14 @@ jobs:
           path: apps/files_pdfviewer
           ref: ${{ matrix.server-versions }}
 
+      - name: Checkout notifications app
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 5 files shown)


marcoambrosini (approved):

jancborchardt (changes requested):
Small detail: In Collectives we never refer to a page as "Readme" so it would be good to use the page name. Meaning:
- If it's the Readme of a folder, use the folder name
- If it's the Readme of a Collective, just use only the Collective name and don't append the "- Readme" part

Is that possible @juliusknorr?

mejo- (approved):
Thanks @juliusknorr, works well! I pushed some minor changes (displaying "Landingpage" instead of "Readme"), a cypress test and some fixes to make psalm happy.

PR #1685, opened 2025-03-17
Fix CI and update github actions

juliusknorr:
- Adapting https://github.com/nextcloud/text/commit/64d9887bf94c7d23d9f8094b427ba03b5fbfd4ce to make cypress pass again
- Update action templates
- Pin actions to hashes
- Fix psalm


max-nextcloud:
Cypress sharing tests are failing on stable27 with 500 responses from the server.

max-nextcloud (approved):

Issue #1709, opened 2025-04-09
Hide the templates when looking at a document read-only


tintou:
When I share a document, the "Template" button near the search bar is still clickable and one can then list the temples, on read-only mode this isn't really useful.

PR #1711, opened 2025-04-09
fix(landingpage): Do not show the Show Templates button on public share

tintou:
### 📝 Summary

* Resolves: #1709

The user is not going to do some edition at this point, do not allow them to list the templates.

### 🏁 Checklist

- [X] Tests (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [X] Documentation ([README](https://github.com/nextcloud/collectives/blob/main/README.md) or [documentation](https://github.com/nextcloud/collectives/blob/main/docs/)) has been updated or is not required

### DIFF
--- a/src/components/PageList.vue
+++ b/src/components/PageList.vue
@@ -14,7 +14,7 @@
 				trailing-button-icon="close"
 				:show-trailing-button="isFilteredView"
 				@trailing-button-click="clearFilterString" />
-			<NcActions class="toggle toggle-push-to-right">
+			<NcActions v-if="!isPublic" class="toggle toggle-push-to-right">
 				<NcActionButton class="toggle-button"
 					:aria-label="labels.showTemplates"
 					@click="toggleTemplates()">


src/components/PageList.vue
```diff
@@ -14,7 +14,7 @@
 				trailing-button-icon="close"
 				:show-trailing-button="isFilteredView"
 				@trailing-button-click="clearFilterString" />
-			<NcActions class="toggle toggle-push-to-right">
+			<NcActions v-if="!isPublic" class="toggle toggle-push-to-right">
```
mejo-, line 17:
```suggestion
			<NcActions v-if="currentCollectiveCanEdit" class="toggle toggle-push-to-right">
```

Public collectives shares can be read-write, so there's no reason not to edit templates in public shares. But I agree that we should only show the button if the collective is editable.

mejo- (approved):
Thanks for your contribution, much appreciated! I left a comment.

mejo- (changes requested):
Requested changes.

mejo- (approved):

Issue #1319, opened 2024-06-12
Allow multiple Templates for a page, Allow page Template


AnjaSimons:
Sometimes, for instance for project management, it is nice to have a page with many subpages as a template. And a template make have constant texts and modifiable fields. So it would be nice, to add this feature, so a user can create a new collective from a template or a new page from a template, which can contains multiple subpages, which are also templates.

Aveyron-RetD:
I agree a collective as a template would be handy

jimdaza:
Agree! that would be incredibly useful! 
Thanks to the team!

Aveyron-RetD:
=> Ok as a workaround I found this which is not very difficult to do
=> you need to have your collective template already prepared
Do the following in the specific order
=> Create a collective
=> Go to file
=> Go to the folder collective
=> In here you should see the template collective and your newly created collective
=> enter the template collective
=> Select all => Copy => move to the newly created collective => When asked about the readme.md replace it.

TADA new collective ready to go!!!

alexanderdd:
@mejo- I find the title a little confusing, maybe rename this issue? "multi-page templates for whole collectives"?

related issues: (I see you are aware of them but i still think it is good if they are connected here in github)
https://github.com/nextcloud/collectives/issues/1195
https://github.com/nextcloud/collectives/issues/267

Issue #1195, opened 2024-04-10
Page emoji in template


siccovansas:
**Is your feature request related to a problem? Please describe.**
I can't set a default page emoji in a template. This means that I have to manually change the page emoji every time I create a new page based on a template. I expect to be able to specify this in a template.

**Describe the solution you'd like**
I want to specify a default page emoji in a template.

Issue #267, opened 2021-12-09
Improve user experience around templates


mejo-:
Using the templates feature is confusing to users.

1. Not everybody understands the impacts of toggling 'view templates' in the page list (it lists templates **and** adds the option to create a template subpage for pages). We should rethink the user interface here.
2. When selecting "Create a template subpage" from the page actions three-dot-menu (next to the edit/done button), the new template page is opened but not listed in the page list at all. That's confusing.

mejo-:
I spent some time to think about implementation details for improved template handling in Collectives, see below. I might or might not start to implement this in next weeks in case nobody objects.

I'd especially appreciate feedback from @max-nextcloud, @juliusknorr and @nextcloud/designers 😉 

## User interface

When creating a new page and templates exist, a modal to select empty page or a template opens, like it's already when creating a new text file from the files app:

![Image](https://github.com/user-attachments/assets/db69a013-d1f5-4f54-972f-af0f5393ae2d)

To manage collective-wide templates, add a "manage templates" action to the collective actions. This switches the page list to a list of existing templates where one can add, remove and edit collective-wide templates.

![Image](https://github.com/user-attachments/assets/3e81e1a6-2548-4dc8-b8ff-fe385b951780)

**Open questions:**
- List personal markdown templates from the files Templates folder?
  -> I'd tend do not do this as this will clutter the template selection a lot. If people want their personal templates in the collective, they have to copy it once.
- Support instance-wide templates that are available for each collective?
  -> I'd say not for now, we can still consider to add it later.

## Possible implementation

**Backend:**
* Store template files in special folder `appdata/collectives/templates/<collective_id>`
  - Problem: how to expose template files via DAV when editing them in Text? Alternatively we could use a subfolder of the collectives folder, like `appdata/collectives/<collective_id>/.templates`. This would mean the folder is exposed via DAV automatically, but we'd have to special-treat it then in several ways:
    - Disallow creating a top-level page with the name `.templates`.
    - Ignore the folder when listing normal pages and subpages.
* Implement custom TemplateController + TemplateService that allows to list, add and remove template files for a collective.
  * There's no use for the global ITemplateManager class as it's bound to the personal template folder.

## Migration

We have to think about how to migrate existing template folders to the new system.

I'd suggest to move all existing `Template.md` files in a Collective into their respective templates folder and rename them to the parents page name - as they're at the moment effectively templates for subpages of their parent page.

max-nextcloud:
[...]

PR #1727, opened 2025-04-22
Template improvements

mejo-:
### 📝 Summary

* Resolves: #267
* Resolves: #1195 
* Resolves: #1319 

#### 🖼️ Screencast

| Screencast |
| --- | 
| a |

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] Tests
- [ ] Copy full pages (with emojis etc,) not only page files

### 🏁 Checklist


mejo-:
I changed the title for the blank page item in template selection from "Blank" to "Blank page" in the meantime so it's more explicit.

max-nextcloud:
:heart_eyes: Just had a look at the screenshots and the screencast so far and I like them a lot!
Super clean and seems very usable.

max-nextcloud (commented):
:microscope: minor nitpick:
 `fix(pages): Unset trash pages when switching the page `
 is when switching the collective - not when switching the page, right?

max-nextcloud (commented):
When renaming a template I would expect any click outside the text input to save the new name rather than leaving the input active.

grafik

lib/Controller/TemplateController.php
```diff
@@ -0,0 +1,89 @@
...
+	#[NoAdminRequired]
+	public function setEmoji(int $collectiveId, int $id, ?string $emoji = null): DataResponse {
+		return $this->handleErrorResponse(function () use ($collectiveId, $id, $emoji): array {
+			$userId = $this->getUserId();
+			$templateInfo = $this->templateService->setEmoji($collectiveId, $id, $emoji, $userId);
+			return [
+				'data' => $templateInfo
+			];
+		}, $this->logger);
+	}
```
max-nextcloud, line 88: Why not use the corresponding page actions for this?

max-nextcloud: We talked this through and decided it may be useful to have separate api endpoints in case we want to for example restrict work on templates to admins later on.

mejo-:
> When renaming a template I would expect any click outside the text input to save the new name rather than leaving the input active.

Good catch. Fixed this now along with some other problems I spotted in the frontend implementation. Also rebased on latest main.

mejo-:
Also added the repair step for migrating old templates now. To test it, add some `Template.md` pages in your collectives, bump the app version in `appinfo/info.xml` and run `occ upgrade`.

marcoambrosini (approved):
Very nice @mejo- 
A few comments from design, none of them blocking. Adding checkboxes so you can easily convert to issue should you need to.

## Editing state
- [ ] There should be some clear indication that I'm editing a template when I'm doing so. Could we embed something like this in the viewer title slot? (screenshot comes from anytype):
Screenshot 2025-04-30 at 17 16 14

Otherwise we could play with the string in the title but I'm worried that it wouldn't be very clear if we do that.

## New document creation flow
- [ ] Template picker should be added to the suggestion bar and not interrupt the flow of creating a new document

## Templates list
- [ ] Remove margins between list items
- [ ] Remove border radius of list items
- [x] Icons should have no background
Example screenshot with these tweaks: Screenshot 2025-04-30 at 17 29 26

mejo-:
> Icons should have no background

The icons are a button to edit the emoji for the template. I wonder whether that's still discoverable with a tertiary button without background.

marcoambrosini:
I think it's ok since this can be edited while editing the template. I would be ok with removing this functionality here and simplify the interaction with this list

mejo-:
> I think it's ok since this can be edited while editing the template. I would also be ok with removing this functionality here and simplify the interaction with this list

Nope, unfortunately that's not the case. Now that we open the template markdown file in the viewer, there's no way to edit the emoji while editing the page.

mejo-:
> > I think it's ok since this can be edited while editing the template. I would also be ok with removing this functionality here and simplify the interaction with this list
> 
> Nope, unfortunately that's not the case. Now that we open the template markdown file in the viewer, there's no way to edit the emoji while editing the page.

We could add an action "Set emoji" in the three-dots menu though to highlight the possibility of editing the emoji.

max-nextcloud (approved):
Thanks a lot! :heart:

Issue #1736, opened 2025-04-29
Anchor links not working


bentuna:
**Describe the bug**
Anchor links to headings do not scroll the page to the right point.

**To Reproduce**
1. Click on the hashtag left of a heading in a collectives page
2. Copy the current URL
3. Open that URL in a new browser tab
4. See how the collectives page gets opened, but not scrolled to the linked heading

**Expected behavior**
The page should be scrolled automatically to the linked heading.

**Additional Context**
This issue is applicable for internal links as well as share links of collective pages.

**Server details:**
- Collectives app version: 2.16.1
- Nextcloud version: 31.0.4

PR #1748, opened 2025-05-08
fix: Scroll to location hash when opening an anchor link

mejo-:
Fixes: #1736

### 🏁 Checklist

### DIFF
--- a/src/mixins/editorMixin.js
+++ b/src/mixins/editorMixin.js
@@ -27,6 +27,7 @@ export default {
 			editorContent: null,
 			pageInfoBarPage: null,
 			updateEditorContentDebounced: debounce(this.updateEditorContent, 200),
+			updateCounter: 0,
 		}
 	},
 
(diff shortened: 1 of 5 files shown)


elzody (approved):
I gave it a test, it works well for me. I also think the code is quite elegant

src/mixins/editorMixin.js
```diff
@@ -185,6 +200,31 @@ export default {
...
+
+					headingEl.scrollIntoView({ behavior: 'instant' })
+					scrolled = true
+				}
+
+				if (scrolled === true) {
+					// Scroll up a bit as the heading is hidden behind the menubar otherwise
+					const scrollContainerEl = document.querySelector('.page-scroll-container')
+					scrollContainerEl.scrollBy({ top: -43, behavior: 'instant' })
+				}
```
max-nextcloud, line 224: There are two alternatives to this:
* in text we scroll the element to the middle of the screen https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#block (block: 'center')
* There's the [scroll-margin css property](https://developer.mozilla.org/de/docs/Web/CSS/scroll-margin) when scrolling

PR #1768, opened 2025-05-20
Bump psalm error level and fix issues spotted by psalm

mejo-:
### 🏁 Checklist

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <psalm
-	errorLevel="5"
+	errorLevel="4"
 	resolveFromConfigFile="true"
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xmlns="https://getpsalm.org/schema/config"
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,46 +1,46 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="5"
-    resolveFromConfigFile="true"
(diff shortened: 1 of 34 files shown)


psalm.xml
```diff
@@ -1,46 +1,81 @@
...
+				<referencedClass name="Doctrine\DBAL\Schema\Schema" />
+				<referencedClass name="Doctrine\DBAL\Schema\SchemaException" />
+				<referencedClass name="Doctrine\DBAL\Schema\Table" />
+			</errorLevel>
+		</UndefinedDocblockClass>
+		<DeprecatedClass>
+			<errorLevel type="suppress">
+				<directory name="lib" />
+			</errorLevel>
+		</DeprecatedClass>
```
max-nextcloud, line 49: I'm not quite sure I understand what this does...

`type="suppress"` sounds like it suppresses the errors rather than complaining.


Since the commit message sounds like it treats more things as errors than before I assume that just mentioning these issue handlers activates them and they will only be suppressed in the `lib` directory. But then again... isn't most php code in lib? This seems to stem from server where there's also `core`, `aps`, `ocs`... so there it makes sense to me. But in collectives? Do we actually need to exclude `lib`?

mejo-: Indeed, the changes were wrong. I started working on bumping the psalm error level now instead. This PR bumps the error level to `4` and brings many fixed reported by error level `3`. There's still around 57 errors to be fixed though, which I'd tackle later on in a follow-up PR.

PR #1771, opened 2025-05-21
test(ci): Run psalm against app store release with latest migrations

mejo-:
We had upgrade failures already a few times in the past because a migration or repair step referenced a new constant or function in an existing class.

Due to the way app upgrades are handled in Nextcloud, the same PHP process that patches the app sources with the new ones from the app store then runs the new migrations and repair steps. Existing classes don't get refreshed before that, so depending on new objects in an existing PHP class can cause problems here.

This commit adds a CI check that statically checks the migrations and repair steps from latest development branch along with the app code of the former app store release.

This should catch situations where we accidentally depend on new objects in migrations.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/psalm-migrations.yml
+++ b/.github/workflows/psalm-migrations.yml
@@ -1,0 +1,82 @@
+# This workflow is provided via the organization template repository
+#
+# https://github.com/nextcloud/.github
+# https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization
+#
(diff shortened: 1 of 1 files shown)


mejo-:
The PR is ready for review. I'll keep the second commit for now to illustrate how the check is failing when running against Collectives appstore release 2.16.3 with the broken Migration that used `NodeHelper::putContent()`. Will remove the commit once reviews are done.

max-nextcloud (approved):
Nice... using psalm to check the code in the old context seems like a very clean approach that also points out the problems in question.

blizzz (approved):

Issue #234, opened 2021-08-28
Named versions


mejo-:
In GitLab by @drachenzauberei on Aug 28, 2021, 20:09

I'd like to be able to enter a comment for an edit like "New section on XYZ" which will end up in the history in the sidebar. Only seeing "5 minutes ago" is not that helpful.

max-nextcloud:
Named versions are supported by the server for a while now. We're not supporting them in collectives yet.

PR #1784, opened 2025-05-28
Overhaul versions support

mejo-:
### 📝 Summary

* Resolves: #234

**Backend:**
* Implement version metadata support (also usable via files app frontend)

**Frontend:**
* Show version label and author if existent
* Allow to compare versions in viewer
* Allow to download versions
* Allow to restore versions from sidebar
* Allow to delete versions
* Allow to label versions

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
grafik | grafik

### 🚧 TODO

- [ ] Tests

### 🏁 Checklist


mejo-:
This PR is mostly ready for review. I have to figure out why the restore cypress test is failing while it works in manual tests, but apart from that it's ready from my side. So would be happy about reviews already :blush:

cypress/e2e/page-versions.spec.js
max-nextcloud, line 1: :heart:

lib/Db/CollectiveVersionMapper.php
```diff
@@ -0,0 +1,54 @@
...
+
+use OCP\AppFramework\Db\QBMapper;
+use OCP\IDBConnection;
+
+/**
+ * @extends QBMapper<CollectiveVersion>
+ */
+class CollectiveVersionMapper extends QBMapper {
+	public function __construct(IDBConnection $db) {
+		parent::__construct($db, 'collectives_p_versions', CollectiveVersion::class);
```
max-nextcloud, line 20: I was briefly wondering what the `p` is about. I suspect it's `page`, right?

mejo-: Exactly. `collectives_page_versions` is too long as table name. That's why I went with `collectives_p_versions` as I didn't want to strip `collectives`. We also already have `collectives_u_settings` or something alike, so I thought I'll follow this scheme.

juliusknorr:
I also had a look through the code just didn't get through it fully before it was merged, but seems good. Generally I'm wondering if at some point we should consider reusing the files sidebar to some degree, but that requires more upstream work in server then I guess to make it more flexible. Very nice feature addition in general 👍

PR #1799, opened 2025-06-04
test(cy): Don't just run specific tests

mejo-:
### 🏁 Checklist


juliusknorr (approved):
🙈

PR #1424, opened 2024-08-18
Slugify Collectives and Pages

Koc:
### 📝 Summary

This is a very beginning of my PR to slugify urls for Collectives and Pages. For now urls are too long and hard to understand if using cyrylic in this entities.

Here an example of some of them: `http://localhost/index.php/apps/collectives/%D0%A3%D0%BA%D1%80%D0%B0%D1%97%D0%BD%D1%81%D1%8C%D0%BA%D1%96%20%D1%84%D1%83%D1%82%D0%B1%D0%BE%D0%BB%D1%96%D1%81%D1%82%D0%B8/%D0%AF%D1%80%D0%BC%D0%BE%D0%BB%D0%B5%D0%BD%D0%BA%D0%BE%20%D0%90%D0%BD%D0%B4%D1%80%D1%96%D0%B9%20%D0%9C%D0%B8%D0%BA%D0%BE%D0%BB%D0%B0%D0%B9%D0%BE%D0%B2%D0%B8%D1%87?fileId=187`. Just compare it with slugified version: `http://localhost/index-php/apps/collectives/ukrayinski-futbolisti-1/page-1-yurchenko-andriy-mykolayovich`. Easy to read, 3 times shorter.

#### 🖼️ Video
[Screencast from 2024-09-01 17-18-46.webm](https://github.com/user-attachments/assets/47e96118-8425-4926-bcbd-605f7042740d)

### 🚧 TODO
- [x] resolve updater [issue](https://github.com/nextcloud/collectives/pull/1424#discussion_r1819174500)
- [x] update previewer
- [x] add/fix tests

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [ ] Tests (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [ ] Documentation ([README](https://github.com/nextcloud/collectives/blob/main/README.md) or [documentation](https://github.com/nextcloud/collectives/blob/main/docs/)) has been updated or is not required

PR #1802, opened 2025-06-08
Slugify Collectives and Pages

Koc:
This is continuation of #1424

This PR aims to slugify urls for Collectives and Pages. For now urls are too long and hard to understand if you're using cyrylic in this entities.

Here an example of some of them: `http://localhost/index.php/apps/collectives/%D0%A3%D0%BA%D1%80%D0%B0%D1%97%D0%BD%D1%81%D1%8C%D0%BA%D1%96%20%D1%84%D1%83%D1%82%D0%B1%D0%BE%D0%BB%D1%96%D1%81%D1%82%D0%B8/%D0%AF%D1%80%D0%BC%D0%BE%D0%BB%D0%B5%D0%BD%D0%BA%D0%BE%20%D0%90%D0%BD%D0%B4%D1%80%D1%96%D0%B9%20%D0%9C%D0%B8%D0%BA%D0%BE%D0%BB%D0%B0%D0%B9%D0%BE%D0%B2%D0%B8%D1%87?fileId=187`. Just compare it with slugified version: `http://localhost/index-php/apps/collectives/ukrayinski-futbolisti-1/page-1-yurchenko-andriy-mykolayovich`. Easy to read, 3 times shorter.

#### 🖼️ Video
[Screencast from 2024-09-01 17-18-46.webm](https://github.com/user-attachments/assets/47e96118-8425-4926-bcbd-605f7042740d)

### 🚧 TODO

- [ ] Fix Cypress tests

### 🏁 Checklist


max-nextcloud:
I tried this locally:

* checked out the latest main
* ran `make setup-dev && npm run dev`
* opened a preexisting collective with a few pages
* Added new links between the pages of various forms (preview or plain link)
* checked out this branch
* ran the upgrade in the browser

Afterwards:

- file that I was on could not be loaded
- list of collectives is empty
- `Could not fetch collectives` and `could not fetch collectives from trash` warnings are showing.
- The two requests in question are receiving 404 responses.

mejo-:
@max-nextcloud the problems you report stem from a problem with older versions of Collectives not being compatible with latest server master branch. This got fixed by using servers `LazyFolder` some time ago, as you already figured out.

> update: looks like this branch has not been rebased on top of https://github.com/nextcloud/collectives/commit/7035d47d0fbd3153b5a9c573acb4fde16f6ce205 yet and that's why it's failing while latest main branch succeeds.

Weird, could it be that you're on an old version of this PR? I also still had a very old version checked out at some time with still the old branch upstream (Koc's private repo). This PR was already updated to latest Collectives main branch. I pushed another rebase now, but it only rebases on the l10n commit from tonight.

max-nextcloud:
Tested this including the migration some and it works really smooth.

Only thing I noticed is no slugs were generated for pages with subpages during the migration.

grafik

This is also visible when navigating the collective:

https://github.com/user-attachments/assets/3c1e6f64-f188-48cb-9427-4583205c76f0

lib/Command/GenerateSlugs.php
```diff
@@ -0,0 +1,106 @@
...
+			->where($update->expr()->eq('file_id', $update->createParameter('file_id')));
+
+		while ($rowCollective = $resultCollectives->fetch()) {
+			$circle = $this->circleHelper->getCircle($rowCollective['circle_unique_id'], null, true);
+			$pageInfos = $this->pageService->findAll($rowCollective['id'], $circle->getOwner()->getUserId());
+
+			foreach ($pageInfos as $pageInfo) {
+				if ($pageInfo->getFileName() === PageInfo::INDEX_PAGE_TITLE . PageInfo::SUFFIX) {
+					continue;
+				}
```
max-nextcloud: This is leaving out all parent pages, right?

lib/Migration/GenerateSlugs.php
```diff
@@ -0,0 +1,116 @@
...
+		$update = $this->connection->getQueryBuilder();
+		$update->update('collectives_pages')
+			->set('slug', $update->createParameter('slug'))
+			->where($update->expr()->eq('file_id', $update->createParameter('file_id')));
+
+		$rootFolder = $this->collectiveFolderManager->getRootFolder();
+
+		while ($row = $result->fetch()) {
+			$pageFile = $rootFolder->getById($row['file_id'])[0];
+			if (!($pageFile instanceof File) || NodeHelper::isIndexPage($pageFile)) {
```
max-nextcloud: Also leaving out the parent pages - so I suspect this is on purpose?

max-nextcloud (approved):
Approving - but wondering if we could include parent pages in the slug generation. I suspect this is meant to only exclude the landing pages - but I'll leave that to you @mejo- 

I also tried renaming collectives and pages. With the slugs links still work nicely after a rename. The link preview shows the old name both of the collective and the page - I guess it's cached. But the link works fine. Links that existed before the migration stop working once a collective is renamed. Renaming the page is handled just fine.

cypress/e2e/page-slug-urls.spec.js
```diff
@@ -0,0 +1,50 @@
...
+	const nameUrl = `/apps/collectives/${encodeURIComponent(collectiveName)}/${encodeURIComponent(pageName)}`
+
+	before(function() {
+		cy.loginAs('bob')
+		cy.deleteAndSeedCollective(collectiveName)
+			.seedPage(pageName, '', 'Readme.md').then(({ collectiveId: cid, pageId: pid }) => {
+				collectiveId = cid
+				pageId = pid
+				slugUrl = `/apps/collectives/${collectiveSlug}-${collectiveId}/${pageSlug}-${pageId}`
+			})
```
max-nextcloud: Is there any reason for renaming the ids back and forth?

This should work as well, right?
```suggestion
			.seedPage(pageName, '', 'Readme.md').then(({ collectiveId, pageId }) => {
				slugUrl = `/apps/collectives/${collectiveSlug}-${collectiveId}/${pageSlug}-${pageId}`
			})
```

cypress/e2e/pages-links.spec.js
```diff
@@ -351,8 +385,7 @@ describe('Page link handling', function() {
 		})
 	})
 
-	// TODO: Adapt tests to Nextcloud 29+
-	describe.skip('Link handling public share', function() {
+	describe('Link handling in public share', function() {
```
max-nextcloud, line 388: uhh nice... thanks for bringing that back!

lib/Migration/GenerateSlugs.php
```diff
@@ -0,0 +1,116 @@
...
+		$output->startProgress();
+		$this->generateCollectiveSlugs($output);
+		$output->finishProgress();
+		$output->info('done');
+
+		$output->info('Generating slugs for pages ...');
+		$output->startProgress();
+		$this->generatePageSlugs($output);
+		$output->finishProgress();
+		$output->info('done');
```
max-nextcloud, line 55: Somehow the output during `occ upgrade` is not as nice as the code sounds.
It seems to leave out the info but prints the name instead.

grafik

lib/Migration/Version030000Date20240820000001.php
```diff
@@ -0,0 +1,36 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('collectives_pages');
+		if (!$table->hasColumn('slug')) {
+			$table->addColumn('slug', Types::STRING, [
+				'notnull' => false,
+				'default' => false,
+				'length' => 255,
```
max-nextcloud, line 28: I checked and trying to create longer page names or collective names runs into other errors (filename or team name not allowed) before hitting this limit. So seems fine.

max-nextcloud (approved):
Did a small code review. Looks really nice as well :heart_eyes:

cypress/e2e/page-slug-urls.spec.js
mejo-: Yes, `pageId` is a global variables in the context. It's also used in later tests. I changed `collectiveId` to be local only now.

lib/Command/GenerateSlugs.php
mejo-: Good catch! Fixed now.

lib/Migration/GenerateSlugs.php
mejo-: Same here.

cypress/e2e/page-slug-urls.spec.js
max-nextcloud: ahh.. makes sense. Should have seen there is no const in front of collectiveId and pageId.

lib/Migration/GenerateSlugs.php
mejo-: I looked into this briefly and even if I completely remove the `$output->info()` calls from the repair step, the output stays the same. All the output comes from `$output->startProgress()` and `$output->finishProgress()`. I'd leave it as is as it's a one-time thing and I don't think it warrants to put more time and energy into it.

Koc:
thank you so much! I can't believe that it happened :smiley_cat: . Going to be drunk today :beers:

arnowelzel:
I missed that update first - but I'd like to thank you for this great improvement for Collectives!

Issue #1680, opened 2025-03-05
Sorting of pages in reverse order


ronnybremer:
**Is your feature request related to a problem? Please describe.**
Meeting minutes and other pages are prefixed with the date they were held on, like "2024-10-22 Testpage". The default sort order is ordering them ascending, leading to the newer pages at the very bottom. No issue with five pages, but after a while it gets cumbersome to scroll all the way down and you might miss important updates.
Custom order is error prone in this context and order by most recently changed doesn't work if older pages get updated frequently.

**Describe the solution you'd like**
In addition to order pages by name ascending I would like to see order pages by name descending.

**Describe alternatives you've considered**
Custom order, but we have to add back logs and dragging them to the right position is not very efficient.

**Additional context**

ronnybremer:
@mejo- Thank you for taking care of this!

PR #1803, opened 2025-06-08
feat(PageList): Allow to sort pages descending by title/time

mejo-:
Fixes: #1680

#### 🖼️ Screenshots

| 🏡 After |
|--- |
| image |

| Screencast |
| --- |
| recording1 |

### 🏁 Checklist

### DIFF
--- a/lib/Db/Collective.php
+++ b/lib/Db/Collective.php
@@ -43,8 +43,10 @@ class Collective extends Entity implements JsonSerializable {
 
 	public const pageOrders = [
 		0 => 'byOrder',
-		1 => 'byTimestamp',
-		2 => 'byTitle',
+		1 => 'byTimeAsc',
+		2 => 'byTitleAsc',
+		3 => 'byTimeDesc',
+		4 => 'byTitleDesc',
 	];
 	public const defaultPageOrder = 0;
 
--- a/src/util/sortOrders.js
+++ b/src/util/sortOrders.js
@@ -4,23 +4,27 @@
  */
 
 const byName = (a, b) => a.name.localeCompare(b.name, OC.getLanguage(), { numeric: true })
-const byTitle = (a, b) => a.title.localeCompare(b.title, OC.getLanguage(), { numeric: true })
-const byTimestamp = (a, b) => b.timestamp - a.timestamp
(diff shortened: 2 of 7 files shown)


lib/Db/Collective.php
```diff
@@ -43,8 +43,10 @@ class Collective extends Entity implements JsonSerializable {
 
 	public const pageOrders = [
 		0 => 'byOrder',
-		1 => 'byTimestamp',
-		2 => 'byTitle',
+		1 => 'byTimeAsc',
+		2 => 'byTimeDesc',
```
max-nextcloud: This index now has a new meaning. Is it stored somewhere? Wouldn't it be better to preserve the existing meaning and append the new options?

src/util/sortOrders.js
```diff
@@ -4,23 +4,27 @@
  */
 
 const byName = (a, b) => a.name.localeCompare(b.name, OC.getLanguage(), { numeric: true })
-const byTitle = (a, b) => a.title.localeCompare(b.title, OC.getLanguage(), { numeric: true })
-const byTimestamp = (a, b) => b.timestamp - a.timestamp
+const byTitleAsc = (a, b) => a.title.localeCompare(b.title, OC.getLanguage(), { numeric: true })
+const byTitleDesc = (a, b) => b.title.localeCompare(a.title, OC.getLanguage(), { numeric: true })
```
max-nextcloud: Maybe easier to understand / parse when expressed in terms of the ascending option?

```suggestion
const byTitleDesc = (a, b) => byTitleAsc(b, a)
```

max-nextcloud (approved):
Screenshare looks great. Code looks good. Did not test it but confident it works as described iwht the tests included.

Just some minor comments.

src/util/sortOrders.js
mejo-: Great idea, changed!

lib/Db/Collective.php
mejo-: Yes, indeed it is persisted in the database. So far I thought I prefer to have the options in a meaningful order for cleaner code and that users using `byTitle` sorting will have to adjust it one time after the upgrade. But now thinking about it again I'll adjust it as you suggested, so the existing option indices don't change their meaning.

PR #1809, opened 2025-06-10
feat: Migrate API to OCS API with OpenAPI support

mejo-:
### 🚧 TODO

- [ ] Migrate remaining controllers to `OCSController`
- [ ] Add annotations and description to controller functions
- [ ] Fix typing errors reported by psalm
- [ ] Generate OpenAPI specification

### 🏁 Checklist


appinfo/routes.php
```diff
@@ -160,18 +24,162 @@
...
+		// Public pages trash API
+		['name' => 'publicPageTrash#index', 'url' => '/api/v{apiVersion}/p/collectives/{token}/pages/trash', 'verb' => 'GET',
+			'requirements' => ['apiVersion' => '(1.0)']],
+		['name' => 'publicPageTrash#delete', 'url' => '/api/v{apiVersion}/p/collectives/{token}/pages/trash/{id}', 'verb' => 'DELETE',
+			'requirements' => ['apiVersion' => '(1.0)', 'id' => '\d+']],
+		['name' => 'publicPageTrash#restore', 'url' => '/api/v{apiVersion}/p/collectives/{token}/pages/trash/{id}', 'verb' => 'PATCH',
+			'requirements' => ['apiVersion' => '(1.0)', 'id' => '\d+']],
+
+		// Public template pages API
+		['name' => 'publicTemplate#index', 'url' => '/api/v{apiVersion}/p/collectives/{token}/pages/templates', 'verb' => 'GET',
```
max-nextcloud, line 182: :sparkles: Routes look good and consistent to me.
I wonder if it would make sense to replace
`/p/collectives/{token}` with `/p/{token/`. The latter looks familiar to me and signals a share rather than a collective.
Do these routes always point to an entire collective or can it also be a share of a page and its subpages? The latter would be a case for using `/p/{token}/` from my point of view.

lib/Controller/CollectiveController.php
```diff
@@ -35,118 +44,179 @@ public function __construct(
 		private IFactory $l10nFactory,
 		private LoggerInterface $logger,
 		private NodeHelper $nodeHelper,
+		private string $userId,
```
max-nextcloud, line 47: uh... nice. Where does that come from? Seems like that will save quite a few `getUserId()` functions.

src/apis/collectives/pages.js
```diff
@@ -41,32 +75,32 @@ export function createPage(context, page) {
  * Get a page in the given context (collective or public share)
  *
  * @param {object} context - either the current collective or a share context
- * @param {number} pageId - Id of the page to retrieve
+ * @param {number} pageId - ID of the page to retrieve
```
max-nextcloud, line 78: Mainly out of curiosity ... why capitalize both chars in ID?

max-nextcloud (approved):
I have a few minor comments. Looked over all parts of the code changes but not at everything.

max-nextcloud:
I'm curious to see if we can eventually use https://openapistack.co/docs/openapi-client-axios/intro/ or something similar to make use of the openapi spec for a strongly typed client in the frontend.

lib/Controller/CollectiveController.php
mejo-: `OCSController` allows it to be injected directly. I guess `lib/private/AppFramework/DependencyInjection/DIContainer.php` takes care of it.

src/apis/collectives/pages.js
mejo-: Well, probably a matter of taste. I always did it that way. Just looked it up on and seems like Wikipedia agrees with me that `ID` should be fully capitalized :laughing: https://en.wikipedia.org/wiki/Identifier

appinfo/routes.php
mejo-: Well, that's a good question. Technically spoken, a page share is also a "stripped down collective".

This very route always points to the templates of a "collectives app share", whether it's a shared full collective or a shared page+subpages.

I don't have strong feelings here, we have `/collectives/` in front of all the OCS API routes related to a collective now, and as shares always belong to a particular collective (and are not agnostic to collectives, like e.g. app-wide user settings), it's more consistent with the session routes this way.

PR #1827, opened 2025-06-23
enh!: require Nextcloud 30

max-nextcloud:
`nextcloud/openapi-extractor` requires at least php 8.1
but Nextcloud 29 still supports 8.0.

grafik

See https://github.com/nextcloud/collectives/pull/1821#issuecomment-2993728398

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -70,16 +70,16 @@
         },
         {
             "name": "teamtnt/tntsearch",
-            "version": "v5.0.1",
+            "version": "v5.0.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/teamtnt/tntsearch.git",
-                "reference": "1a86ca3431dce7fcd816b153d5a340f5665bf40e"
+                "reference": "3f6078c37d55feab3927d8f988f9e1e8b3aaa2a0"
             },
             "dist": {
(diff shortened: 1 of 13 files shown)


max-nextcloud:
I'm not sure we actually need to bump the requirement. We could also run openapi extractor on php > 8.1 and still support Nextcloud 29.
The error mainly originates from the fact that the workflow uses the minimum supported php version (which makes sense). I think with #1826 it will not adjust the composer.lock file anymore. So it ends up being incompatible.

max-nextcloud:
@provokateurin maybe you have an idea how best to handle this.
openapi-extractor requires at least php 8.1 - which normally is not a problem. But the php-cs workflow will run on the lowest supported php version of all supported Nextcloud versions. And that is 8.0 if we want to keep support for Nextcloud 29. I see several options:
1) drop support for Nextcloud 29
2) run php-cs on a later php version
3) remove openapi-extractor before installing the dependencies similar to how we remove nextcloud/ocp

mejo-:
Thanks for looking into this @max-nextcloud. Given that Nextcloud 29 is EOL, I'm all for dropping support for it.

mejo- (approved):

PR #1828, opened 2025-06-23
fix(command): extend `Symfony\Component\Console\Command\Command`

mejo-:
### 🏁 Checklist


lib/Fs/NodeHelper.php
```diff
@@ -126,7 +126,10 @@ public static function getContent(File $file): string {
 		}
 
 		if (!mb_check_encoding($content, 'UTF-8')) {
-			$content = mb_convert_encoding($content, 'UTF-8');
+			$convertedContent = mb_convert_encoding($content, 'UTF-8');
+			if ($convertedContent !== false) {
```
max-nextcloud, line 135: If I understand this correctly this will happen if the content is in a different encoding than the default or contains invalid characters for some reason, right?
I think it would be good to log this. Maybe even as a warning.

mejo-:
> I don't understand what the benefits of extending Symfonies Command over OC/Cores are. Did not see any deprecation notice in the OC file. But I trust your judgement. Just one minor comment other than that.

`Base` is in `OC` namespace, i.e. not part of the public API. We should avoid to use private server classes where possible because they might change their interfaces. Most other apps I checked use `Symfony\...\Command`.

lib/Fs/NodeHelper.php
mejo-: Agreed, doesn't hurt to log a warning here. Added now.

PR #1832, opened 2025-06-27
fix(RecentPageTile): Correct semantics for hover

kra-mo:
### 📝 Summary

Objects which are hovered should float instead of pressing down, I'm guessing this was an oversight :)

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist


mejo- (approved):
Thanks @kra-mo :pray:

PR #1837, opened 2025-06-30
fix: string validation fixes

mejo-:
Source string validation was turned on for l10n strings, which caught a few usages of `...` instead of `…`. I took the opportunity to search for other occurrences of this.

### 🏁 Checklist


max-nextcloud (approved):
Not sure how the changes to the l10n files behave with the transifex sync. But I guess it will work out fine... otherwise we learn something. :wink:

PR #1843, opened 2025-07-01
Persist parts of pinia store in browser localStorage

mejo-:
### 📝 Summary

Use `useLocalStorage` composable from vueuse to persist parts of the pinia store state.

### 🏁 Checklist

### DIFF
--- a/src/stores/root.js
+++ b/src/stores/root.js
@@ -5,15 +5,17 @@
 
 import { defineStore } from 'pinia'
 import { set } from 'vue'
-import { editorApiReaderFileId, editorApiUpdateReadonlyBarProps, pageModes } from '../constants.js'
+import { useLocalStorage } from '@vueuse/core'
+import { editorApiReaderFileId, editorApiUpdateReadonlyBarProps } from '../constants.js'
+
+const STORE_PREFIX = 'collectives/pinia/root/'
 
 export const useRootStore = defineStore('root', {
 	state: () => ({
-		textMode: pageModes.MODE_VIEW,
-		showings: {},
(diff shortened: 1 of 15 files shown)


src/stores/root.js
```diff
@@ -28,8 +31,9 @@ export const useRootStore = defineStore('root', {
 
 		isPublic() { return !!this.shareTokenParam },
 
-		isTextEdit: (state) => state.textMode === pageModes.MODE_EDIT,
-		isTextView: (state) => state.textMode === pageModes.MODE_VIEW,
+		getTextMode: (state) => (pageId) => state.textMode[pageId] ?? pageModes.MODE_VIEW,
+		isTextEdit: (state) => (pageId) => state.getTextMode(pageId) === pageModes.MODE_EDIT,
+		isTextView: (state) => (pageId) => state.getTextMode(pageId) === pageModes.MODE_VIEW,
```
max-nextcloud: My impression is that those are only ever called with `currentpage.id` (or `pageId` in a component where `pageId` is a prop that is fed `currentpage.id`).

Do you expect this to change? I.e. do you expect to call these with other page ids than the current page?
Otherwise we could also get the `currentpage.id` from the store and keep these as plain getters, right?

max-nextcloud (approved):
Looks great. I'm happy it was so easy. I have one comment about the getters and setters that now take a `pageId` argument.

I did not test this - but I assume you did.

src/stores/root.js
mejo-: Good point. I thought about it as well but was reluctant to move `textMode` to the pages store. But in the end this is where it belongs, even more when it depends on `currentPage`. Did so now.

PR #1845, opened 2025-07-01
chore: Remove unused linkHandler mixin

mejo-:
It used to be the link handler passed to Text. But nowadays, links don't get handled by Text anymore as we have the link bubble that shows the link preview.

### 🏁 Checklist


max-nextcloud (approved):
Great catch!

I think we also still have a link handler in text. Maybe we can also remove that one then?

mejo-:
> I think we also still have a link handler in text. Maybe we can also remove that one then?

Yes, probably. I'm still a bit unsure whether there are edge cases when the link handler might still come into play. But given it only handled normal clicks (and not stuff like <kdb>Ctrl</kbd>-click) I wouldn't see when that would be. But anyway something to be discussed somewhere else :laughing:

PR #1870, opened 2025-07-12
chore(update): drop outdated cypress dependencies

max-nextcloud:
cypress/browserify is not maintained anymore and pulls in even more outdated dependencies.


mejo- (approved):
Thank you :pray:

Issue #1703, opened 2025-04-05
Expose File Tags to the Collectives UI


noantiq:
While the hierarchical structure that Collectives allows you to build is already nice to organize different kinds of notes to a certain extent, I would like to be able to further categorize and tag certain pages to quickly find all notes on a specific topic across all folders or sub pages.

As pointed out in #482, you can already use Files to assign tags to the corresponding markdown files. I did not find a way to view or manage these tags from within Collectives and switching to Files to assign and search for tags is rather cumbersome. Therefore, exposing the existing tag system to the Collectives UI would really help here. Ideally I would want to be able to see which tags are assigned to each page and be able to manage as well as search for them from within Collectives.

This touches on #470 and #1604 .

Issue #1604, opened 2024-12-04
Tag system Feature with Diary style and Collectives


4NBK:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Hi,

I use "Logseq" since two years and it's a good one but too much "technical style" and "comfortless" for "classicals users".

The big feature i love with it is the "tag system" that allow you to put in the "journal page" all the things you have in mind and find them later with a "search fonction". The other things is that "link" this part with the page which have the same name or create it if it's not already done.

That's feature to probably do witch will "change the game" i think in the futur of KPM expansion tools.

But i'am not a Developer... 😥 I would like to have your feelings about that.

And i really do the parallel in my mind with   "Diary + Collectives" on Nextcloud.

Example : 

I write something in my Journal page or "Diary"

![Image](https://github.com/user-attachments/assets/08854513-3df5-4e46-83a2-f66d34c23f4c)

Then i just want to see the informations about "Nextcloud Subject". So when i click on it : 

![Image](https://github.com/user-attachments/assets/58878ae9-7fd7-434e-942a-d6d1f1b6c83f)

Same thing for application tag : 

![Image](https://github.com/user-attachments/assets/5957d413-0563-4b2e-ba9b-8abf9df735ee)

I could also have written a tag on a page other than the "journal" and it would also have been referenced too.

Do you thinks that will be a good thing too and possible to have the same interaction and features type ?

Have a good day

4NBK:
Thanks for your feed back. @dasaweb . And thanks for the link.

I 'll follow these. And switch "on" my notifiactions... ;p

Issue #470, opened 2022-12-24
Tags / Categorisation of pages


thopanX:
**Is your feature request related to a problem? Please describe.**
I'm currently using templates to automatically "assign a category" to a subpage of a parent page. I'm writing this in "", because this is simply a link to the parent page and not a real categorisation, I guess? It does work for me, but some sort of (semi-)automatic procedure would be even better in my optinion / for my usecase.

![grafik](https://user-images.githubusercontent.com/1814897/209442408-6f06d8c8-9208-4d3c-ac9f-7fd384401063.png)

**Describe the solution you'd like**
A parent page could automatically become a category with the subpages inheriting the category. Or this could be configurable per parent page to be able to include/exclude parent pages from categorisation. 

**Describe alternatives you've considered**
An alternative is to use a template for new subpages of a parent page. This is what I'm currently doing.

kizm0:
I think this would be quite useful, for example i have a collective "bookshelf" where i collect all my books that i read, i would like to apply categories like scifi, on the top of the searchbar it would be useful to filter all books within this category...

estux:
I think this would be more useful e.g. also for indicating the "status" of a page, to collaborate better with the others working on the same Collective. 
E.g. I would like to let the others know the pages to modify (as we're working on them) and the finished pages (to not touch anymore). 
A more complex use-case would be a multi-level stages work on the pages. (e.g. To do, frozen, next, WIP, ready)

I hope this makes sense. :)

mejo-:
@max-nextcloud and me discussed this feature a bit and came up with the following thoughts:

## Backend implementation and scope of tags

Nextcloud server already provides a tags API that's used for files, but it suggest to not use it for tags in collectives.

File tags are visible instance-wide (see e.g. https://github.com/nextcloud/server/issues/32455). Especially on big instances, this is somewhat unexpected.

Therefore we suggest a collectives-specific tags/labels implementation like the deck app does. New tags will be bound to the collective in question and not visible/usable in other collectives.

## User interface

- In the page list, no tags will be displayed per default, to not clutter the list too much.
- The page search/filter (at the top of the page list) allows to filter by tags.
- Tags for pages shall be displayed below the title of the currently open page, more or less like the following:
  ![Image](https://github.com/user-attachments/assets/64ac49ad-4c8e-4e81-b399-b059588e57fa)
- Clicking on a tab anywhere will add the tag filter to the page list and limit the page list to pages with this tag

## Open question

[...]

PR #1877, opened 2025-07-14
Page tags

mejo-:
### 📝 Summary

* Resolves: #470
* Resolves: #1604
* Resolves: #1703 

#### 🖼️ Screencast

| Screencast |
| --- |
| recording1 |

#### 🖼️ Screenshots (slightly outdated)

🏚️ Dark mode | 🏡 Bright mode
---|---
grafik | grafik
grafik | grafik
grafik | grafik
grafik | grafik
grafik | grafik

### 🚧 TODO

- [x] Page filtering via tags
- [x] Cypress tests
- [ ] Cleanup tags when removing collective from trash

### 🏁 Checklist


lib/Migration/Version030100Date20250711000000.php
```diff
@@ -0,0 +1,51 @@
...
+			'notnull' => true,
+			'length' => 250,
+		]);
+		$table->addColumn('color', Types::STRING, [
+			'notnull' => true,
+			'length' => 6,
+		]);
+
+		$table->setPrimaryKey(['id']);
+		$table->addUniqueIndex(['id', 'collective_id'], 'c_tags_unique_idx');
```
max-nextcloud: That seems odd to me. `id` itself is autoincrement and I would expect it to be unique. So I assume you meant `name` and `collective_id` to be uniq.
But maybe you already fixed this in a later commit.

lib/Service/TagService.php
```diff
@@ -0,0 +1,83 @@
...
+	 */
+	private function checkPermissions(int $collectiveId, string $userId, bool $edit = false): void {
+		if ($this->collectiveMapper->findByIdAndUser($collectiveId, $userId) === null) {
+			throw new NotFoundException('Collective not found: ' . $collectiveId);
+		}
+
+		if ($edit && !$this->collectiveService->getCollective($collectiveId, $userId)->canEdit()) {
+			throw new NotPermittedException('Not allowed to edit collective');
+		}
+	}
```
max-nextcloud: Minor nitpick:

I'd split this into `checkReadPermission` which only performs the first check and `checkEditPermission` which only performs the latter.

That's one db query less in the edit case as `$this->collectiveService->getCollective(...)` also does the check against `=== null`.

It's also slightly more explicit on the calling side than the parameter `true`.

tests/Integration/features/publicTags.feature
```diff
@@ -0,0 +1,30 @@
...
+
+  Scenario: Prepare collective
+    When user "jane" creates collective "BehatTagsPublicCollective"
+    And user "jane" creates public share for "BehatTagsPublicCollective"
+    And user "jane" sets editing permissions for collective share "BehatTagsPublicCollective"
+
+  Scenario: Create and update tag in public collective share
+    When anonymous creates tag "test1" with color "FF0000" for collective "BehatTagsPublicCollective" with owner "jane"
+    Then anonymous fails to create existing tag "test1" with color "FFFFFF" for collective "BehatTagsPublicCollective" with owner "jane"
+    And anonymous updates tag "test1" with color "FF0000" for collective "BehatTagsPublicCollective" with owner "jane"
```
max-nextcloud, line 16: Nothing changes here, right?

lib/Service/PageService.php
```diff
@@ -961,6 +985,50 @@ private function removeFromSubpageOrder(int $collectiveId, int $pageId, int $rem
...
+		$pageInfo = $this->getPageByFile($file);
+
+		$collectiveTags = array_map(static fn ($tag) => $tag->getId(), $this->tagMapper->findAll($collectiveId));
+		if (!in_array($tagId, $collectiveTags, true)) {
+			throw new NotFoundException('Tag ' . $tagId . ' not found for collective');
+		}
+
+		$tags = PageTagHelper::add($pageInfo->getTags(), $tagId, $collectiveTags);
+		$pageInfo->setTags($tags);
+		$this->updateTags($collectiveId, $pageInfo->getId(), $userId, $tags);
```
max-nextcloud: I smell a race condition. If two updates happen at the same time and both first read the current tags and then add theirs and write the result the first write will be overwritten and only one tag will be added.

First of all I'd move the `in_array` check to the start of the function right below verifying the edit permissions. That's one query less during the crucial time.

Then I see two options:
* locking the row between the read and the write. Both processes would need to acquire a write lock before performing the read so that in the scenario I described before the lock actually has an effect.
* Adding a `where` condition to the update that checks the tags field has not changed in the meantime, retrying if it did change.

All in all this seems like an unlikely thing to happen often. But I can imagine scenarios like the client was offline and fires off a batch of tagging requests when coming back online.

lib/Service/PageService.php
```diff
@@ -961,6 +985,50 @@ private function removeFromSubpageOrder(int $collectiveId, int $pageId, int $rem
...
+		$pageInfo = $this->getPageByFile($file);
+
+		$collectiveTags = array_map(static fn ($tag) => $tag->getId(), $this->tagMapper->findAll($collectiveId));
+		if (!in_array($tagId, $collectiveTags, true)) {
+			throw new NotFoundException('Tag ' . $tagId . ' not found for collective');
+		}
+
+		$tags = PageTagHelper::remove($pageInfo->getTags(), $tagId, $collectiveTags);
+		$pageInfo->setTags($tags);
+		$this->updateTags($collectiveId, $pageInfo->getId(), $userId, $tags);
```
max-nextcloud: Same thing here with the race condition

lib/Migration/Version030100Date20250711000000.php
mejo-: Good catch. Indeed I meant `name` and `collective_id`. Will fix.

lib/Service/TagService.php
mejo-: Agreed, cleaner!

tests/Integration/features/publicTags.feature
mejo-: That's already fixed in a later commit. Sorry for not fixing it in the original commit.

lib/Service/PageService.php
mejo-: Agreed regarding moving up the `in_array()` check. I'm a bit unsure whether the extra complexity with implementing locks and retries is in relation to the likeliness of race conditions.

mejo-: We decided against implementing the extra complexity for now. Instead, I added a comment in code.

tests/Unit/Service/PageTagHelperTest.php
max-nextcloud, line 1: Great to see logic moved to it's own helper and unit tested :heart:

src/components/Collective.vue
```diff
@@ -219,6 +221,9 @@ export default {
 		async getAllPages(setLoading = true) {
 			await this.getPages(setLoading)
 				.catch(displayError('Could not fetch pages'))
+			await this.getTags()
+				.catch(displayError('Could not fetch tags'))
+			await this.getTags()
```
max-nextcloud: Seems like a typo - I don't see why you'd want to get them twice.

src/components/Page/PageTags.vue
```diff
@@ -0,0 +1,132 @@
...
+				// Filter out undefined (if tag got removed)
+				.filter(Boolean)
+		},
+
+		pageTagsVisible() {
+			return this.pageTagsDecorated.slice(0, TAGS_LIMIT)
+		},
+
+		pageTagsInvisible() {
+			return this.pageTagsDecorated.slice(TAGS_LIMIT, this.pageTagsDecorated.length)
```
max-nextcloud: minor: You can leave out the second parameter if you want all elements to the end.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#end

src/components/Collective.vue
mejo-: Yes, indeed. But already fixed in a subsequent commit :face_with_peeking_eye:

src/stores/tags.js
```diff
@@ -0,0 +1,102 @@
...
+			const response = await api.getTags(this.context)
+			set(this.allTags, this.collectiveId, response.data.ocs.data.tags)
+			set(this.allTagsLoaded, this.collectiveId, true)
+		},
+
+		_addOrUpdateTagState(tag) {
+			if (!this.allTags[this.collectiveId]) {
+				set(this.allTags, this.collectiveId, [])
+			}
+			const idx = this.tags.findIndex(t => t.id === tag.id)
```
max-nextcloud, line 85: I find the mixture of `this.allTags[this.collectiveId]` and `this.tags` slightly confusing here and also in the `delete` function.

~~I suggest to move the check for `!this.allTags[this.collectiveId]` into the `tags` getter and then use `this.tags` everywhere.~~ Okay... no setting in getter. How about watching `this.collectiveId` and when it changes make sure `this.allTags[this.collectiveId]` is at least an empty array?

src/components/Page/PageTags.vue
```diff
@@ -74,10 +91,22 @@ export default {
 			return this.pageTagsDecorated.slice(TAGS_LIMIT, this.pageTagsDecorated.length)
 		},
 
+		pageTagTitle() {
+			return t('collectives', 'Add tag to page list filter')
```
max-nextcloud: wording: add to filter sounds a bit technical. "Filter page list by this tag" ?

max-nextcloud:
Done with the code review. Will try it out later.

src/stores/tags.js
mejo-: Seems like with the current options syntax, there's [no way to define watchers within a pinia store](https://github.com/vuejs/pinia/discussions/794).

I'd say let's leave it as is for now and if we ever migrate to composition syntax stores, we can switch to using a watcher? :thinking:

max-nextcloud (approved):

PR #1889, opened 2025-07-17
fix(RecentPage): Fix landing page detection

mejo-:
### 🏁 Checklist

### DIFF
--- a/src/components/Collective.vue
+++ b/src/components/Collective.vue
@@ -93,16 +93,7 @@ export default {
 		},
 		'currentPage.id'() {
 			this.selectVersion(null)
-
-			// Redirect to slugified URL if possible
-			if (this.currentCollective
-				&& this.isLandingPage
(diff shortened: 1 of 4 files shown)


src/components/Collective.vue
```diff
@@ -240,6 +232,18 @@ export default {
 		closeNav() {
 			emit('toggle-navigation', { open: false })
 		},
+
+		slugUrl() {
+			// Redirect to slugified URL if possible
+			if (this.currentCollective
+				&& this.isLandingPage
+				&& this.$route.fullPath !== this.currentCollectivePath) {
+				this.$router.replace(this.currentCollectivePath)
```
max-nextcloud: I think you'd also want to maintain the hash here so one can link to headings on the landing page.

mejo-: Good catch. Changed now.

Issue #1780, opened 2025-05-26
"Last changed" when only changing title emoji


janbaum:
**Describe the bug**
When only changing the title emoji of a file, in the "last changed" info, the user changes correctly, but the date keeps the old one.

**To Reproduce**
1. Create a file with user1 and select a title emoji
2. wait one minute
3. log in with user2, select the page and **just** change the emoji _(without enabling "edit"-mode)_ 
4. The "last changed" info now shows user2 as the one, who last changed the document, but the time is still the one, when user1 edited the file the last time

**Expected behavior**
Is a changing the emoji equivalent to chaning the file? If so, also update the last-changed-date, if not, keep user1 as the last editing user

**Server details:**
- Nextcloud version: 30.0.10.1
- PHP Version: ?
- Database: ?

**Client details:**
 - OS: Linux
 - Browser: Firefox, Chrome
 - Browser version: latest
 - Device: Desktop

max-nextcloud:
#1895 bumps the version timestamp of the md file which causes text to assume it was updated and in turn causes conflicts with currently running editing sessions.

So we need another way to fix this. Ideas that come to my mind:
* Have a separate modification timestamp for the metadata in the pages database and render the later one.
* touch the file and update the metadata in the documents table in text
* have text recognize that the text content did not change at all
* figure out why the etag changes if a file is only touched but not altered.

mejo-:
Thinking about this again, I would say let's just update the last changed user when changing the file contents. When changing emoji, neither user nor timestamp should get updated.

PR #1901, opened 2025-07-23
test(cy): Only test link handling against `https` URLs

mejo-:
Opening `http://example.org` (instead of `https`) resulted in `ERR_EMPTY_RESPONSE (-324) loading 'http://example.org/'` recently.

### 🏁 Checklist

### DIFF
--- a/.github/workflows/cypress-e2e.yml
+++ b/.github/workflows/cypress-e2e.yml
@@ -42,7 +42,7 @@ jobs:
               - 'babel.config.js'
               - 'composer.json'
               - 'composer.lock'
-              - 'cypress.config.js'
+              - 'cypress.config.mjs'
               - 'package.json'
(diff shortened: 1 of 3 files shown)


max-nextcloud (approved):

max-nextcloud:
Right now even the https version is down. So I will change this to `https://github.com`. We rely on github being accessible for our CI already anyway.

mejo-:
> Ahh... I see. I think this was caused by https://github.com/nextcloud/collectives/commit/06d58cdc9b9867d70df1bc60d36852bf1d982219

Oh wow, good catch and great that our CI caught it :disappointed_relieved:

I agree that we should revert 06d58cdc9b9867d70df1bc60d36852bf1d982219 and reopen #1780 for now.

PR #1911, opened 2025-07-30
Use Notify push body to update page list

### LINKED ISSUE #1908: Collectives will flood server with requests if lots of notify push updates come in
**Describe the bug**
We had a high load on our server tonight.
Part of it was open collectives tabs sending requests for pages, templates and the trash all of the time.

It looks like due to a text issue some file was updated all the time - leading to constant notify_push updates.

So even if there's a lot of notify push updates - let's not flood the server with even more requests.

**To Reproduce**

I'd rather not. 😉 

**Expected behavior**

Only send the requests needed to update the user.

## Possible solution

Include the information about the updated page in the notify push message. That way no requests are needed on top of it.

## Other possible improvements

* [ ] only send requests when collectives tab is visible. (This will update other tabs via local storage anyway)
* [ ] only request pages by default.
* [ ] only request trash when it's open.
* [ ] Maybe only request templates when creating a new page.
* [ ] Throttle to 1 request per 30 seconds? (First request immediate and then only 2 per minute.)

max-nextcloud:
fixes #1908 

## Server side

Send a list of the updates pages as well as the ids of pages who were removed entirely.

- **enh(notifyPush): allow sending a body**
- **chore(refactor): move notifyPush out of updatePage**
- **chore(refactor): move notifyPush out of updateSubpageOrder**
- **chore(refactor): move notifyPush out of add/removeSubpageOrder**
- **chore(refactor): move notifyPush out of updateTags**
- **chore(refactor): remove notifyPush from newPage**
- **enh(notifyPush): send page info in body**

### DIFF
--- a/src/stores/pages.js
+++ b/src/stores/pages.js
@@ -10,7 +10,7 @@ import { getCurrentUser } from '@nextcloud/auth'
 import { generateRemoteUrl } from '@nextcloud/router'
 import { useRootStore } from './root.js'
 import { useCollectivesStore } from './collectives.js'
-import { INDEX_PAGE, pageModes } from '../constants.js'
+import { INDEX_PAGE, PAGE_SUFFIX, pageModes, TEMPLATE_PATH } from '../constants.js'
 /* eslint import/namespace: ['error', { allowComputed: true }] */
 import * as sortOrders from '../util/sortOrders.js'
 import * as api from '../apis/collectives/index.js'
@@ -71,7 +71,7 @@ export const usePagesStore = defineStore('pages', {
 				: (!rootStore.pageId && !rootStore.pageParam) || rootStore.pageParam === INDEX_PAGE
 		},
 		isIndexPage(state) {
-			return state.currentPage.fileName === INDEX_PAGE + '.md'
+			return state.currentPage.fileName === INDEX_PAGE + PAGE_SUFFIX
 		},
 
 		rootPageForCollective(state) {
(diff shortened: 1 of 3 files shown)


src/stores/pages.js
```diff
@@ -923,18 +894,22 @@ export const usePagesStore = defineStore('pages', {
 				return
 			}
 			for (const page of pages) {
+				if (page.filePath === '.templates' || page.filePath.startsWith('.templates/')) {
```
mejo-: The corresponding code in templates store is slightly different as it uses the `TEMPLATE_PATH` const there. I wonder if this can be harmonized and the const be used in both cases. We already have `src/constants.js` anyway, maybe just move `TEMPLATE_PATH` there and use it in both stores? :thinking:

src/stores/templates.js
```diff
@@ -12,8 +12,11 @@ import { usePagesStore } from './pages.js'
 import * as api from '../apis/collectives/index.js'
 import { byTitleAsc } from '../util/sortOrders.js'
 import { TEMPLATE_PAGE } from '../constants.js'
+import { removeFrom, updateOrAddTo } from './collectionHelpers.js'
 
 const STORE_PREFIX = 'collectives/pinia/templates/'
+const TEMPLATE_PATH = '.templates'
+const INDEX_PAGE_NAME = 'Readme.md'
```
mejo-: We have the `INDEX_PAGE` const in `src/constants.js`, maybe reuse it here.

mejo- (approved):
Thanks again for tackling this :heart: 

Only two nitpicking comments, nothing blocking.

PR #1913, opened 2025-08-01
fix(session): Raise session valid time to 120 seconds

mejo-:
Should fix racy situations where update session request comes too late and as a result a new session needs to get created.

### 🏁 Checklist


max-nextcloud (approved):
Thanks!

Issue #1924, opened 2025-08-11
Automatically opening the first collective does not work anymore


SebieF:
**Describe the bug**
As described in #1241 and changed in #1264 , if only one collective exists, the first one should be opened automatically. In my test for version 3.0.3 on Firefox, this does not work anymore. Instead, the usual start page of the collective app is shown. I verified that only one Collective exists.

**To Reproduce**
1. Install version 3.0.3
2. Go to "Collectives" App
3. Start Page is shown instead of first collective

**Expected behavior**
The first collective should be automatically opened if only one exists.

**Server details:**
- Collectives app version: 3.0.3
- Nextcloud version: 30.0.5

**Client details:**
 - OS: Ubuntu
 - Browser: Firefox

PR #1928, opened 2025-08-11
fix(home): Open collective straight away if only one exists

mejo-:
Fixes: #1924

### 🏁 Checklist


src/views/Home.vue
```diff
@@ -50,16 +50,16 @@ export default {
...
 
 	watch: {
-		// Open the navigation if we already have collectives.
-		// Only has an effect on mobile (where navigation is closed per default).
-		'collectives'(val, oldval) {
-			if (oldval.length === 0 && val.length === 1) {
-				this.$router.push(`/${encodeURIComponent(val[0].name)}`)
-			} else if (oldval.length === 0 && val.length > 1) {
+		'collectives'(val) {
+			if (val.length === 1) {
```
max-nextcloud: Isn't this triggered a little to often? Everytime `collectives` changes and the length is 1, right? So if I only have one collective and I rename it it will be opened? If I had two collectives, remove one the other one will open.

Maybe we could watch `collectives.length` instead?

mejo-: Good point. I don't understand why this is in a watcher at all, probably because `this.collectives` was empty during mount of the component earlier, or maybe just because I didn't know better back when I implemented it.

Moved the logic to `mounted()` where it belongs in my opinion. The `Home.vue` view is only used in the app start page, i.e. when no collective is selected. And there I don't see a reason why we would want to react on changing number of collectives. Removing a collective should not result in rerouting or opening the navigation in my opinion, and creating a collective will open it anyway.

It tested both rerouting to the only collective and opening the navigation on mobile. For the latter we now also have a Cypress test.

PR #1973, opened 2025-09-09
chore(migrate) editorMixin to useEditor composable

max-nextcloud:
Some preparation for using a single editor instance.


src/composables/useEditor.js
```diff
@@ -0,0 +1,197 @@
+/**
+ * SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
```
mejo-:
```suggestion
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

nitpick :wink:

PR #1983, opened 2025-09-16
fix(pageList): Current page loading only when moving itself

mejo-:
No need to mark current page as loading when copying a page or moving another page.

### 🏁 Checklist


max-nextcloud (approved):
Code seems legit - did not test but I assume you did / this came out of observing the issue and fixing right away.

Issue #1950, opened 2025-08-31
Tag search ignores all capitalized characters


noantiq:
**Describe the bug**
When searching for a tag that begins with a capitalized character, it can't be found. I can only find tags when searching for substrings that are in lowercase in the tag.

I'd suspect that on one side the string is converted to lowercase and that is missing on the other side, but that's just a guess when observing the behaviour.

**To Reproduce**
1. Open a collective
2. Tag a page with 'PageTag'
3. Start typing 'PageTag' into the search
4. No tag to select

**Expected behavior**
The tag is found when searching for it, even when it has capitalized characters.

**Screenshots**

<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/39112752-ae7a-4851-a034-ad6e5afd80bf" />

<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/4fc6b703-5882-4677-80ff-a481044f7e05" />

<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/b47e0bbf-b4ca-47b5-aab7-de0aba247b70" />

<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/62aeeab3-2035-47df-8dbe-ec6b5a9140d0" />

<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/6c588d79-c841-4c4a-aa01-ff8fedd77964" />
<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/d3f63c79-396b-4da1-8c64-7b9903b6e500" />
<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/65a6f4b7-b824-4adc-85ab-c41a55e1abd6" />
<img width="534" height="142" alt="Image" src="https://github.com/user-attachments/assets/4560edc1-28c2-4c0f-9399-e9d85ba7db6b" />

**Server details:**
- Collectives app version: 3.1.1
- Nextcloud version: 31.0.6

PR #1987, opened 2025-09-17
Make tag filtering and sorting case-insensitive

silverkszlo:
### 📝 Summary

* Resolves: #1950 

### Problem
When searching for tags that contain uppercase letters, they are not found because the search and sorting logic was case-sensitive. For example, a tag named "PageTag" could only be matched using lowercase input.

### Solution
Convert both the tag names and the search string to lowercase before filtering and sorting. This makes the search and sorting logic case-insensitive, so tags can be found regardless of capitalization.

### 🏁 Checklist


mejo- (approved):
Thanks so much for tackling this :pray:

Issue #1961, opened 2025-09-04
Track online / offline state and disable page requests when offline


max-nextcloud:
**Is your feature request related to a problem? Please describe.**

When being offline these warnings are shown regularly:

<img width="344" height="342" alt="Image" src="https://github.com/user-attachments/assets/71f9d0df-28d0-4dd3-9c77-18f3f88ac548" />

Instead indicate the offline state somewhere all the time and show no additional warnings / issue no requests.

Issue #1796, opened 2025-06-03
Disable sidebar when offline


mejo-:
None of the sidebar tabs are available when offline and it will stay like this for a while.

## Proposal

Replace the error indicators in the tabs with an offline empty content and also change the toggle to indicate the offline state.

## Current state

The sidebar currently includes:
* Attachments
* Backlinks
* Sharing
* Versions

<img width="344" height="263" alt="Image" src="https://github.com/user-attachments/assets/8f2cc941-bde2-4475-89bf-11205d455cb3" />

<details>
<summary>
Original issue ( was only about versions ).
</summary>
Versions is something we don't want to support while being offline, at least for now. With offline support slowly coming now, we should show an `NcEmptyContent` with a note while being offline.
</details>

max-nextcloud:
We discussed this further and ended up with the idea to have an offline indicator at the lower left corner of the page content:

<img width="408" height="270" alt="Image" src="https://github.com/user-attachments/assets/55b3d803-d86f-45d4-9d60-f4544133ede9" />

The sidebar would remain as is - except for other messages.

PR #1999, opened 2025-09-23
feat: track online/offline state

mejo-:
### 📝 Summary

* Handle offline state in all page sidebar tabs
* Add offline state indicator
* Don't send page API requests when offline
* Resolves: #1796
* Resolves: #1961 

#### 🖼️ Screenshots offline indicator

Desktop | Mobile | Mobile clicked
--- | --- | ---
image | image | image

#### 🖼️ Screenshots page sidebar tabs

Versions tab | Attachments tab | Shares tab
---|--- | ---
image | image | image

#### 🖼️ Screenshots request errors

Before | Pages request error new | Any other request error new
--- | --- | ---
image | image | image

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/collective-offline.js
+++ b/cypress/e2e/collective-offline.js
@@ -1,0 +1,52 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+describe('Collective offline', function() {
+	let collectiveId, pageId, pageSlugUrl
+	const collectiveName = 'Offline'
+	const pageName = 'First Page'
+	const pageSlug = 'First-Page'
+
+	before(function() {
+		cy.loginAs('bob')
+		cy.deleteAndSeedCollective(collectiveName)
+			.seedPage(pageName, '', 'Readme.md').then(({ collectiveId: cid, pageId: pid }) => {
+				collectiveId = cid
+				pageId = pid
+				pageSlugUrl = `/apps/collectives/${collectiveName}-${collectiveId}/${pageSlug}-${pageId}`
+			})
+	})
+
+	beforeEach(function() {
+		cy.loginAs('bob')
+		cy.visit(pageSlugUrl)
+		// make sure the page list loaded properly
+		cy.contains('.app-content-list-item a', pageName)
+	})
+
+	describe('Offline mode', function() {
+		it('Shows offline indicator', function() {
+			cy.get('.offline-indicator').should('not.exist')
+			cy.goOffline()
+			cy.get('.offline-indicator').should('be.visible')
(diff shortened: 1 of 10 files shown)


cypress/e2e/collective-offline.js
max-nextcloud, line 1: 😍 tests included 🥲

src/components/PageSidebar/SidebarTabAttachments.vue
```diff
@@ -12,8 +12,11 @@
 			</template>
 		</NcEmptyContent>
 
+		<!-- offline -->
+		<OfflineContent v-else-if="!loaded && !networkOnline" />
+
 		<!-- error message -->
-		<NcEmptyContent v-else-if="error" :name="error">
+		<NcEmptyContent v-else-if="error " :name="error">
```
max-nextcloud:
```suggestion
		<NcEmptyContent v-else-if="error" :name="error">
```

src/views/CollectiveView.vue
```diff
@@ -193,18 +198,32 @@ export default {
...
+					promises.push(this.getTrashPages())
 				}
 			}
+
+			try {
+				await Promise.all(promises)
+				this.loadPending = false
+			} catch (e) {
+				displayError('Could not fetch collective details')(e)
+			}
```
max-nextcloud, line 223: Should `loadPending` remain true if a request fails? In fact I don't see it being read anywhere. I think we could do without it.

src/components/PageSidebar/SidebarTabAttachments.vue
```diff
@@ -272,14 +294,17 @@ export default {
 
 	watch: {
 		'page.id'() {
-			this.load('attachments')
-			this.getAttachmentsForPage()
+			this.getAttachmentsForPage(true)
```
max-nextcloud, line 298: I think you need to set `loaded` to `false` here. Otherwise the offline indicator won't show even when navigating to pages whose attachments have never been loaded.

src/views/CollectiveView.vue
mejo-: Indeed, `getAllPages()` is called unconditionally whenever the network comes back online. So no need for the flag in `CollectiveView`.

src/components/PageSidebar/SidebarTabAttachments.vue
mejo-: Good catch as well. Fixed now.

max-nextcloud (approved):
Looks good now.

marcoambrosini (commented):
Nice addition, from my pov I would:

- [x] Shrink the size of the indicator to 12px
- [ ] Only use a button element on mobile, where it should be clickable/tappable
- [x] Remove border on mobile 
- [ ] If possible use a radial gradient background on mobile:

Issue #1912, opened 2025-07-31
Arrow buttons in move/copy modal broken


mejo-:
**Describe the bug**
Positioning the target page with arrow up/down button within the move/copy modal doesn't work.

Hitting the buttons still seems to have an effect in positioning the page after move/copy, but there's no visual effect.

PR #2001, opened 2025-09-24
Fix arrow buttons visual feedback in move/copy modal

silverkszlo:
### 📝 Summary

* Resolves: #1912 

**Before**:
- When users clicked the up/down arrow buttons in the "Move or Copy" modal, the list items did not move inside the modal, i.e., there was no visual feedback.

**How it works now**:
- When users click the up/down arrow buttons in the "Move or Copy" modal, they'll see immediate visual feedback as the page moves in the list
- The reordered position is preserved while they're in the same location
- When they navigate away and come back, or when they click "Move page here", the actual move operation will use the new position they selected
- The visual feedback makes it clear where the page will be moved to before they confirm the action
→ This gives users the visual feedback they need to understand where their page will be positioned before completing the move operation.

#### 🖼️ Screenshots

Screenshot from 2025-09-24 15-36-55

### 🏁 Checklist

### DIFF
--- a/src/components/Page/PagePicker.vue
+++ b/src/components/Page/PagePicker.vue
@@ -315,15 +315,12 @@ export default {
 		swapSubpages(from, to) {
 			const length = this.subpages.length - 1
 			if (from >= 0 && from <= length && to >= 0 && to <= length) {
-				// Create a copy of the subpages array to manipulate
-				const reorderedPages = [...this.subpages]
-				// Swap the elements
-				const temp = reorderedPages[from]
-				reorderedPages[from] = reorderedPages[to]
-				reorderedPages[to] = temp
-
-				// Update the component data to trigger reactivity
-				this.$set(this, 'reorderedSubpages', reorderedPages)
+				// Initialize reorderedSubpages if not already set
+				if (!this.reorderedSubpages) {
+					this.reorderedSubpages = [...this.subpages]
+				}
+				// Use splice to swap elements while maintaining reactivity
+				this.reorderedSubpages.splice(from, 1, this.reorderedSubpages.splice(to, 1, this.reorderedSubpages[from])[0])
 			}
 
 			// Scroll current page into view


src/components/Page/PagePicker.vue
```diff
@@ -309,7 +315,15 @@ export default {
...
-				this.subpages.splice(from, 1, this.subpages.splice(to, 1, this.subpages[from])[0])
+				// Create a copy of the subpages array to manipulate
+				const reorderedPages = [...this.subpages]
+				// Swap the elements
+				const temp = reorderedPages[from]
+				reorderedPages[from] = reorderedPages[to]
+				reorderedPages[to] = temp
+
+				// Update the component data to trigger reactivity
+				this.$set(this, 'reorderedSubpages', reorderedPages)
```
mejo-: Just curious: is there a reason why you don't just use the splice method on `this.reorderedSubpages`? As far as I know, splice keeps reactivity in Vue, so there's no need to create a temporary variable and overwrite the original one using `this.$set()`. Like this way:

```suggestion
				if (!this.reorderedSubpages) {
					this.reorderedSubpages = this.subpages
				}
				this.reorderedSubpages.splice(from, 1, this.subpages.splice(to, 1, this.subpages[from])[0])
```


It also has the (neglectable) advantage that subsequent replacement happen on the visible state (`this.reorderedSubpages`) directly. But that's just nitpicking 🙃

mejo- (approved):
Super nice, works well. While reviewing it i also found the commit where I broke it. 6c23c042cc9f45d399e955878598fe988f0fbebe changed `this.subpages` from a data property to a computed 🙃 

I only have a minor comment out of curiosity.

src/components/Page/PagePicker.vue
silverkszlo: Oh I didn't know that Vue tracks `splice`, so your suggestion is much better :)

And about your code suggestion in line 319: wouldn’t `this.reorderedSubpages = [...this.subpages]` be safer than `this.reorderedSubpages = this.subpages`, since otherwise both would reference the same array and mutations to `reorderedSubpages` might unintentionally affect the original computed/store data?

mejo-: > And about your code suggestion in line 319: wouldn’t this.reorderedSubpages = [...this.subpages] be safer than this.reorderedSubpages = this.subpages, since otherwise both would reference the same array and mutations to reorderedSubpages might unintentionally affect the original computed/store data?

So true, good catch :fishing_pole_and_fish:!

PR #2004, opened 2025-09-25
fix(a11y): Make sure all NcButton components have title or aria-label

mejo-:
Fixes warnings "[WARN] @nextcloud/vue: You need to fill either the text or the ariaLabel props in the button component."

### 🏁 Checklist


src/components/Page/PagePicker.vue
```diff
@@ -41,6 +41,7 @@
 					class="crumbs-level">
 					<ChevronRightIcon :size="20" />
 					<NcButton type="tertiary"
+						:aria-label="t('collectives', 'Navigate to page {page}', {page: page.title })"
```
max-nextcloud: This seems quite verbose given the label of the surrounding div. In the short run I would suggest to remove the `aria-label` on the surrounding div.
In the long run (?) we might want to use a list with a single label for the breadcrumbs and then one label per button: https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/

I'm also not sure if the `Navigate to page` part is necessary.

mejo-: Good catch, will remove the `aria-label` property from surrounding div. About being verbose, my understanding was that `aria-label` ideally should be a bit more verbose than the button text itself, as e.g. blind people don't see the context where the button is. But I'm really not sure here.

src/components/PageSidebar/SharingEntryLink.vue
```diff
@@ -170,10 +170,10 @@
 				:label="t('collectives', 'Password')"
 				@update:value="onPasswordChange" />
 			<div class="button-group">
-				<NcButton @click="cancelSettings">
+				<NcButton :aria-label="t('collectives', 'Cancel')" @click="cancelSettings">
 					{{ t('collectives', 'Cancel') }}
 				</NcButton>
```
max-nextcloud, line 175: My understanding is that you could use the text prop here instead of having the same text in the `aria-label` and the button itself:


```suggestion
				<NcButton :text="t('collectives', 'Cancel')" @click="cancelSettings" />
```

src/components/PageSidebar/SharingEntryLink.vue
```diff
@@ -170,10 +170,10 @@
...
-				<NcButton type="primary" @click="saveSettings">
+				<NcButton type="primary" :aria-label="t('collectives', 'Update share')" @click="saveSettings">
```
max-nextcloud, line 176: Maybe use text prop instead like above?

src/components/PageSidebar/VersionLabelDialog.vue
```diff
@@ -25,11 +25,13 @@
 			<NcButton v-if="versionLabel"
 				type="error"
 				native-type="reset"
+				:aria-label="t('collectives', 'Remove version name')"
```
max-nextcloud, line 28: Maybe use text prop instead like above?

src/components/PageSidebar/VersionLabelDialog.vue
```diff
@@ -25,11 +25,13 @@
...
 				@click="$emit('label-update', '')">
 				{{ t('collectives', 'Remove version name') }}
 			</NcButton>
 			<NcButton type="primary"
 				native-type="submit"
+				:aria-label="t('collectives', 'Save version name')"
```
max-nextcloud, line 34: Maybe use text prop instead like above?

src/components/PageSidebar/SharingEntryLink.vue
mejo-: Unfortunately this seems only possible from `@nextcloud/vue` 9.0.0 onwards. See https://stable8--nextcloud-vue-components.netlify.app/#/Components/NcButton vs. https://nextcloud-vue-components.netlify.app/#/Components/NcButton and the comment "Since 9.0.0" in the latter one. But good to keep in mind and migrate once we migrated to Vue3 and `@nextcloud/vue` 9.

mejo-: I also tested it and it doesn't work with `@nextcloud/vue` 8.

src/components/Page/PagePicker.vue
mejo-: I changed the `aria-label` to "Breadcrumb, navigate to page {page}" now as the best way I could come up with to combine the information that it's a breadcrumb *and* the button allows to navigate to the page.

src/components/PageSidebar/SharingEntryLink.vue
max-nextcloud: oh... sorry for the confusion then!

Issue #1915, opened 2025-08-02
Switching between pages may leave dangling editor


mejo-:
**Describe the bug**
After I switched to another page, the browser kept sending push requests for the old editor session. These resulted in 403 "[ERROR] text: Failed to push the steps to the server" errors.

Only push requests got sent, no sync requests.

Probably the websocket provider from last page editor didn't get cleaned up.

PR #2012, opened 2025-10-01
fix(editor): always delete editor

max-nextcloud:
Wait for the editor to be created so we can destroy it
if the component is unmounted before the editor loaded.

This avoids a race condition
when NcAppContent rendered the default template twice.


mejo- (approved):
Super nice, thanks for catching this 🙏

I wonder whether this also fixes https://github.com/nextcloud/collectives/issues/1915.

mejo-:
Tested now and indeed it also fixed #1915, very cool! The dangling editor from initial load kept making requests to the original session even after switching pages.

PR #2017, opened 2025-10-03
fix(favorites): disable drag'n'drop and sort according to page list

mejo-:
Both bugs were direct feedback from a user who got very confused when they wanted to use page favorites. 

### 🏁 Checklist


max-nextcloud (approved):
code makes sense.... did not test.

PR #2029, opened 2025-10-08
fix(build): separate files for css entry points

max-nextcloud:
Fixes https://github.com/nextcloud/text/issues/7751

Right now the single `collectives-style.css` is loaded everywhere.

I tried to only load it in collectives
but we also have some styles for files.

So we need separate entry points.
Trying to follow what text is doing here.


max-nextcloud:
Currently the css does not get loaded at all. I originally introduced the `Util::addStyle` in the `BeforeTemplateRenderedListener` to fix this during the vite migration.

But this seems to be the wrong approach.
Other apps seem to have no `addStyle` call what so ever. It looks like there's a vite mechanism to include the css that we are missing.

Issue #1700, opened 2025-04-03
"dependency apps Circles is missing" error message is out of date


fredericgrither:
**bug description**
When launching Collectives, if Teams App dependency is missing, an error message will appear mentionning **Circles** instead of **Teams**.
![Image](https://github.com/user-attachments/assets/9fb9af92-2f18-4266-a362-7426966ecdde)

**Steps to reproduce the behavior:**
The [administrator documentation](https://nextcloud.github.io/collectives/administration/) fortunately precises all the dependencies.
There are : files_versions, Viewer, Text and Teams (with Nextcloud >= 29)
With Nextcloud <= 28, the dependency was Circles, hence "Circles" in the error message. Since Circles is deprecated, I suggest swapping the name.

1. Deactivate the "Teams" App by following these steps
- click on user menu on top right. Click on "+Apps"
- filter the Apps list by clicking on in the left-hand vertical menu
- in front of the "Teams" Apps in the list, click on the "Deactivate" button.
2. Click on "Collectives" in the top menu bar
3. See the error message appear on the screen
4. Reactivate "Teams" so that the error message is no longer displayed

**Expected behavior**
Error message text: "Error: missing apps. The following dependency Apps are missing : **Teams**. Please ask the administrator to enable these apps."

**Server details**: (non relevant in this issue)
- Collectives app version: [2.16.1]
- Nextcloud version: [31.0.2]
- PHP Version: [8.3]

**Client details**: (non relevant in this issue)
- OS: [Windows/Debian]
- Browser: [Firefox]
- Browser version: [137]
- Device: [Dell laptop]

PR #2038, opened 2025-10-11
fix(StartController.php): change depricated "circles" to "teams"

janbaum:
Fixes: #1700

### 📝 Summary

* Resolves: #1700

Just changed the according string in StarterController.php

I asked myself though, whether this shouldn't be displayed as **C**ollectives? Since the other strings in this line have been and are all lower case, I left it as is.

🚧 TODO

- [ ] fix link to Nextcloud App Store, since the Teams-App's URL is still `https://apps.nextcloud.com/apps/circles`, but with this PR tries to resolve as `https://apps.nextcloud.com/apps/teams`

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image
 | image

### 🏁 Checklist

### DIFF
--- a/lib/Controller/StartController.php
+++ b/lib/Controller/StartController.php
@@ -45,6 +45,12 @@ class StartController extends Controller {
 		]);
 	}
 
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	public function indexPath(string $path): TemplateResponse {
+		return $this->index();
+	}
+
 	private function checkDependencies(): array {
 		$apps = ['circles', 'files_versions', 'text', 'viewer'];
 		$appsMissing = [];


lib/Controller/StartController.php
```diff
@@ -52,7 +52,7 @@ public function indexPath(string $path): TemplateResponse {
 	}
 
 	private function checkDependencies(): array {
-		$apps = ['circles', 'files_versions', 'text', 'viewer'];
+		$apps = ['teams', 'files_versions', 'text', 'viewer'];
```
mejo-: Unfortunately simply replacing `circles` with `teams` doesn't work. It will now always report a missing `teams` app, as there exists no app with this app ID. We have to pass the app ID (i.e. `circles`) to `$this->appManager->isEnabledForUser`.

So my suggestion would be to keep `circles` in `$apps` and instead retrieve the app name via `$appInfo = $this->appManager->getAppInfo($app)` and return `$appInfo['name']` in `checkDependencies()` below.

templates/error.php
```diff
@@ -26,7 +26,7 @@
 $i = 0;
 $len = count($_['appsMissing']);
 foreach ($_['appsMissing'] as $app) {
-	print_unescaped('<li><a href="https://apps.nextcloud.com/apps/' . $app . '" style="text-decoration: underline;">' . $app . '</a></li>');
+	print_unescaped('<li>' . $app . '</li>');
```
mejo-, line 28: I don't remember why we add the comma between missing apps, it seems like a leftover to me. If I test the code with all apps missing, it looks ugly. I guess we should just remove the four lines of code below this one (and `$i` as it's not needed then anymore neither).

lib/Controller/StartController.php
janbaum: > Unfortunately simply replacing `circles` with `teams` doesn't work. It will now always report a missing `teams` app, as there exists no app with this app ID. 

Yes, my bad, I didn't test whether it's working as intended, sorry!

templates/error.php
```diff
@@ -26,11 +26,7 @@
 $i = 0;
```
mejo-: The initialization of `$i` can also go, now that it's not used anymore.

lib/Controller/StartController.php
```diff
@@ -45,18 +45,13 @@ public function index(): TemplateResponse {
 		]);
 	}
 
-	#[NoAdminRequired]
-	#[NoCSRFRequired]
-	public function indexPath(string $path): TemplateResponse {
```
mejo-, line 50: Removing this function seems like a mistake, no? Removing this route will likely cause big trouble 😆

mejo- (changes requested):
One more thing that needs to be changed, then the PR is good to go 😊

lib/Controller/StartController.php
janbaum: Oh c'mon :disappointed:

mejo-:
Thanks so much @janbaum, I tested again and it works very well now. Also with using the app name instead of ID it's a great general improvement. The CI failures are unrelated, I'll force-merge.

Issue #2003, opened 2025-09-25
Sharing mechanism is unclear and different from standard Nextcloud UI, leading to accidental public sharing of private data.


looeee:
I've been using Collectives for several months now, and sharing links with team members (in our external chat app) using what I thought is the standard Nextcloud sharing mechanism. 

Today I noticed that the shared Collectives links don't look the same as the page URLS.  
While investigating this, I learned that the links I have been sharing are accessible to people outside the organization. 

I don't think I've shared anything too important or secret but I will have to review all the links I shared which is tedious. 

**Here's the standard Nextcloud Sharing UI.**  
Note how the section to share publicly is very clearly linked as "External Shares" and "Create public link":

<img width="504" height="1164" alt="Image" src="https://github.com/user-attachments/assets/4b790bea-dd6c-40a0-909d-74a8837be7a6" />

**Now here's the Collectives Sharing UI.** 
There's nothing there to tell me that it's public. 

<img width="513" height="288" alt="Image" src="https://github.com/user-attachments/assets/8ab002b2-1dc1-4a83-b779-80cdcb17bba1" />

I understand that it doesn't need all the features of the main Nextcloud sharing panel. However, consistent UI design would require that it matches the standard Nextcloud for external share. In other words, it should look like this: 

<img width="511" height="218" alt="Image" src="https://github.com/user-attachments/assets/d71d1117-c321-46d3-ab8d-22d7c61085fc" />

Issue #1629, opened 2025-01-09
Sharing the standard way


kaffeeringe:
**Is your feature request related to a problem? Please describe.**
I spend some minutes looking for the internal sharing option. It's not in the sidebar, where it is in any other app. There is only the link sharing in the sidebar:
![Image](https://github.com/user-attachments/assets/74f5d4d7-cfc0-4636-a71f-66946134fff8)

**Describe the solution you'd like**
Please just add internal sharing to the sidebar, so that it's like in any other app. 

**Describe alternatives you've considered**
I can get used to the way it's shared now, but it's not good, if every app tries to be cleverer about standard features.

juliusknorr:
Thanks for your feature request, currently we do not have internal sharing as a feature in collectives, rather you manage members of the whole collective, so it is slighly different from the sharing option that each page has for links.

Sharing can of course be achieved with the underlying files, but integrating this in collectives might be more complex then just sharing a single file. You would want to also integrate the incoming shares nicely on the receiving side. In addition users might expect to include subpages in the shares as well.

Can you elaborate a bit more on your use case? Are you just thinking about moving the member management to the sidebar?

kaffeeringe:
Ok. That is a different way to view collectives. To me they are something like OneNote, where I share a notebook with certain people.

But you say, that it's another organizational layer next to groups and teams - with an attached notes function.

I thought the member management was the normal sharing function but in the main part of the UI. Why is this yet another way to manage user groups? I think it is logical to have user groups that manage admin permissions and teams that manage access to resources. Why does Collectives have its' own user management?

Ornanovitch:
See also: https://github.com/nextcloud/collectives/issues/1514#issuecomment-2441212659

small1:
The use case would be if i have a quite large collective where many users has access and i want to share a specific page to a user that already has access.  Sharing the url with ?fileid and such might be filtered. So an internal collective link to a specific page would be a good thing.

PR #2045, opened 2025-10-14
Differentiate between members management and external shares in sharing tab

silverkszlo:
### 📝 Summary

* Resolves: #2003 and #1629 

The sharing tab now lists the two ways of "sharing" a collective or a page: either by adding members to your collective - similar to internal shares - or by sharing a public link to members outside your collective. Before that it was not obvious to users that they were sharing a public link by clicking on "Share link" in the sharing tab.

#### 🖼️ Screenshots

🏚️ Before
Screenshot from 2025-10-14 10-05-52

 🏡 After
Screenshot from 2025-10-14 09-49-58

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/src/components/PageSidebar/SidebarTabSharing.vue
+++ b/src/components/PageSidebar/SidebarTabSharing.vue
@@ -123,16 +123,14 @@ export default {
 				: this.sharesByPageId(this.pageId)
 		},
 
-		isCollective() {
-			return this.isLandingPage ? 'collective' : 'page'
-		},
-
 		membersManagementHelpText() {
 			return t('collectives', 'Manage members of your collective. Members have access to the whole collective.')
 		},
 
 		externalSharesHelpText() {
-			return t('collectives', 'Share this {item} with others outside your collective via public links. Anyone with the link can access the shared content.', { item: this.isCollective })
+			return this.isLandingPage
+				? t('collectives', 'Share this collective with others outside your collective via public links. Anyone with the link can access the shared content.')
+				: t('collectives', 'Share this page with others outside your collective via public links. Anyone with the link can access the shared content.')
 		},
 	},
 


src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -13,22 +13,68 @@
...
 		<!-- shares list -->
-		<ul v-else class="sharing-list">
-			<SharingEntryLink v-if="!shares.length" />
-			<SharingEntryLink
-				v-for="(share, index) in shares"
-				v-else
-				:key="share.id"
-				:index="index + 1"
-				:share="share" />
+		<ul v-else class="sharing-list__content">
```
max-nextcloud: I think the `<ul>` should only wrap the `<SharingEntryLink>`s. I'm not sure if `<section>` is valid inside a `<ul>`. Not sure what to make of

Image

[source](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ul#technical_summary)

I assume only the permitted content is allowed.

But it also makes more sense... the list is inside the section - not the other way round.

silverkszlo: Yes you are right! I got rid of the sections altogether, as I only needed an element as a styling wrapper, so I went with div: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section#usage_notes

Screenshot from 2025-10-14 13-17-16

max-nextcloud: Yes... that structure looks much better.

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -13,22 +13,70 @@
...
-				v-for="(share, index) in shares"
-				v-else
-				:key="share.id"
-				:index="index + 1"
-				:share="share" />
-		</ul>
+		<div v-else class="sharing-list__content">
+			<div class="sharing-list-item">
+				<div class="sharing-list-item__header">
+					<h4>{{ t('collectives', 'Collective members') }}</h4>
```
mejo-, line 12: I wonder whether we should display this block always, i.e. also when `loading('shares')` is truthy. It probably requires some refactoring of the template though, so feel free to skip it if you don't feel like doing it 😉

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -13,22 +13,70 @@
...
+					<NcPopover popup-role="dialog">
```
mejo-:
```suggestion
					<NcPopover popup-role="dialog" no-focus-trap>
```

This should fix the console error "Error: Your focus-trap must have at least one container with at least one tabbable node in it at all times"

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -13,22 +13,70 @@
...
+				</div>
+				<NcButton
+					@click="openCollectiveMembers">
+					{{ t('collectives', 'Manage collective members') }}
+				</NcButton>
+			</div>
+			<div class="sharing-list-item">
+				<div class="sharing-list-item__header" style="padding-top: 10px;">
+					<h4>External shares</h4>
+					<NcPopover popup-role="dialog">
```
mejo-:
```suggestion
					<NcPopover popup-role="dialog" no-focus-trap>
```

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -49,16 +100,72 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			membersManagementHelpText: t('collectives', 'Share this page with members of your collective. Members can view and edit the shared content based on their permissions.'),
+			externalSharesHelpText: t('collectives', 'Share this page or collective with others outside your collective via public links. Anyone with the link can access the shared content.'),
```
mejo-: You could use the `isLandingPage` getter from `pagesStore` and then adjust the string accordingly (mention only "Share this page" or "Share this collective").

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -49,16 +100,72 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			membersManagementHelpText: t('collectives', 'Share this page with members of your collective. Members can view and edit the shared content based on their permissions.'),
```
mejo-:
```suggestion
			membersManagementHelpText: t('collectives', 'Manage members of your collective. Members have access to the whole collective.'),
```

There's no way to add members for a specific page, so I'd rephrase it slightly to make obvious that members have access to the full collective. Also I think we should not use the term "sharing" for managing members, as it's something conceptionally different, but I'm unsure here. Would be curious what you think @silverkszlo.

src/components/PageSidebar/SharingEntryLink.vue
```diff
@@ -299,8 +299,8 @@ export default {
 
 		title() {
 			return this.index > 1
-				? t('collectives', 'Share link ({index})', { index: this.index })
-				: t('collectives', 'Share link')
+				? t('collectives', 'Share public link ({index})', { index: this.index })
```
mejo-:
```suggestion
				? t('collectives', 'Create public link ({index})', { index: this.index })
```

To be consistent with files sharing options.

silverkszlo:
> Works really well. Thanks so much @silverkszlo 🙏
> 
> I have some comments, see below.
> 
> Also I wonder whether you would like to change the "Share link" action in collective/page actions the way we discussed it. I don't remember exactly on which alternative wording we decided though. Was it "Share public link"? Or "Create public link"? 🤔

According to my notes we decided to leave it as "Share link" because it only opens the sharing sidebar, which now offers both options: managing collective members and public share links. I also think the shorter version of "Share link" is better in the small dropdown, but that's not the hill I plan to die on, so we still can overthrow that decision ;) @mejo-

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -13,22 +13,70 @@
...
+								</template>
+							</NcButton>
+						</template>
+						<p class="hint-body">
+							{{ membersManagementHelpText }}
+						</p>
+					</NcPopover>
+				</div>
+				<NcButton
+					@click="openCollectiveMembers">
```
mejo-, line 31: One more thing because we're currently working on improving UX when offline: can you add the `disabled` property to this `NcButton` if `networkOnline` from the `useNetworkState` composable is falsy? See e.g. https://github.com/nextcloud/collectives/blob/bfc49ecfc080a8c2d3a326813b52d00022772a51/src/components/PageSidebar/SharingEntryLink.vue#L277-L280 for an example how it's used.

src/components/PageSidebar/SidebarTabSharing.vue
silverkszlo: I agree!

src/components/PageSidebar/SidebarTabSharing.vue
silverkszlo: nice, thank you!

src/components/PageSidebar/SidebarTabSharing.vue
silverkszlo: thanks!

src/components/PageSidebar/SidebarTabSharing.vue
silverkszlo: done ;)

src/components/PageSidebar/SidebarTabSharing.vue
```diff
@@ -49,16 +106,86 @@ export default {
...
+		isCollective() {
+			return this.isLandingPage ? 'collective' : 'page'
+		},
+
+		membersManagementHelpText() {
+			return t('collectives', 'Manage members of your collective. Members have access to the whole collective.')
+		},
+
+		externalSharesHelpText() {
+			return t('collectives', 'Share this {item} with others outside your collective via public links. Anyone with the link can access the shared content.', { item: this.isCollective })
```
mejo-: I'd rather make it two separate strings that contain the whole sentence. It's easier for translators to digest, and "collective" and "page" become part of the translated string and are not hardcoded to the English word.

looeee:
I appreciate this, the new UI looks great!

PR #2046, opened 2025-10-14
fix(router): Detect landing page if it has a slug

mejo-:
The landing page doesn't get a slug generated, but there's a possible corner case: when another page with a slug gets moved to be the landing page via Files app, it has one.

This commit fixes the scenario in two ways:

* In `islandingPage` of pages store, URLs like `/collective-<collective-id>/Readme-<landing-page-id>` get detected as landing page URLs.
* In `pagePath` of pages store, the collective path without page path gets returned for landing pages with a slug.

### 🏁 Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -22,6 +22,7 @@ use OCA\Collectives\Listeners\BeforeTemplateRenderedListener;
 use OCA\Collectives\Listeners\CircleDestroyedListener;
 use OCA\Collectives\Listeners\CircleEditingEventListener;
 use OCA\Collectives\Listeners\CollectivesReferenceListener;
+use OCA\Collectives\Listeners\NodeRenamedListener;
 use OCA\Collectives\Listeners\ShareDeletedListener;
 use OCA\Collectives\Listeners\TextMentionListener;
 use OCA\Collectives\Middleware\PublicOCSMiddleware;
@@ -49,6 +50,7 @@ use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
 use OCP\Collaboration\Reference\RenderReferenceEvent;
 use OCP\Dashboard\IAPIWidgetV2;
 use OCP\Files\Config\IMountProviderCollection;
+use OCP\Files\Events\Node\NodeRenamedEvent;
 use OCP\Files\IMimeTypeLoader;
 use OCP\IUserManager;
 use OCP\IUserSession;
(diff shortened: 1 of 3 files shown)


src/stores/pages.js
```diff
@@ -62,12 +62,18 @@ export const usePagesStore = defineStore('pages', {
...
-				? false
-				: (!rootStore.pageId && !rootStore.pageParam) || rootStore.pageParam === INDEX_PAGE
+
+			if (collectivesStore.currentCollectiveIsPageShare) {
+				// No landing page in page shares
+				return false
+			}
+
+			return (!rootStore.pageId && !rootStore.pageParam)
+				|| (rootStore.pageSlug === INDEX_PAGE && rootStore.pageId === state.rootPage.id)
```
max-nextcloud: Couldn't the `rootStore.pageSlug` also be different from `INDEX_PAGE` ? Does the slug get updated if the page is renamed in files?

src/stores/pages.js
```diff
@@ -145,7 +151,13 @@ export const usePagesStore = defineStore('pages', {
 					? `${collectivesStore.currentCollectivePath}/${pagePath}?fileId=${id}`
 					: collectivesStore.currentCollectivePath
 			}
-			return `${collectivesStore.currentCollectivePath}/${page.slug}-${page.id}`
+
+			// The landing page should not have a slug and get handled above. But if it has a slug set
+			// (e.g. because another page got moved to landing page via Files app), let's still redirect
+			// to the collective path (without page path).
+			return page.slug === INDEX_PAGE && page.id === state.rootPage.id
```
max-nextcloud: Same question as above... would it be possible to have a different slug for the root page?

Also... Maybe this case can be handled with an early return at the top of the function together with the case for a root page that does not have a slug above.
Something like

```js
if (page.id === state.rootPage.id) {
  return collectivesStore.currentCollectivePath
}
...
```

src/stores/pages.js
mejo-: Good catch as well, you're right! I added a listener on the `NodeRenamedEvent` now, that takes care of always updating the slug.

src/stores/pages.js
mejo-: You're right. Thinking about it again, I don't see a reason why we should take `pageSlug` into account at all. Checking `page.id === state.rootPage.id` is enough. Adjusted now.

src/stores/pages.js
max-nextcloud, line 156: If `pagePath` is empty we might want to log this as it's not the `rootPage` but has no `pagePath`. That should not happen, right?

Issue #1833, opened 2025-06-28
Better error handling when creating a collective with a very long name


max-nextcloud:
**Describe the bug**
Trying to create a collective with a very long title will result in an error after adding users only. The error is also not specific:

![Image](https://github.com/user-attachments/assets/a41b0801-b397-42ac-a74c-8a3db39ee91c)

**To Reproduce**
1. Go to the collectives app.
2. Click on 'Create new collective'
3. Enter a very long name
4. Click throught the 'add members' dialog
5. See error

**Expected behavior**
A meaningfull error when entering a collective name that is too long.

**Screenshots**

![Image](https://github.com/user-attachments/assets/7b2c84bd-c2e6-418d-ad71-ad06f12fe4f2)

<details>
<summary>Nextcloud log</summary>

```json
{
  "reqId": "hG0wg2s9131XJUce2V8h",
  "level": 3,
  "time": "2025-06-28T05:52:24+00:00",
  "remoteAddr": "192.168.23.5",
  "user": "admin",
  "app": "no app in context",
  "method": "POST",
  "url": "/ocs/v2.php/apps/collectives/api/v1.0/collectives",
  "message": "An exception occurred while executing a query: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'name' at row 1",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36",
  "version": "32.0.0.1",
  "exception": {
    "Exception": "Doctrine\\DBAL\\Exception\\DriverException",
    "Message": "An exception occurred while executing a query: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'name' at row 1",
    "Code": 1406,
    "Trace": [
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1939,
        "function": "convert",
        "class": "Doctrine\\DBAL\\Driver\\API\\MySQL\\ExceptionConverter",
        "type": "->",
        "args": [
          {
            "__class__": "Doctrine\\DBAL\\Driver\\PDO\\Exception"
          },
          {
            "__class__": "Doctrine\\DBAL\\Query"
          }
        ]
      },
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1881,
        "function": "handleDriverException",
[...]

PR #2053, opened 2025-10-18
Added meaningful message in error toast for collective name too long

cyan-1:
### 📝 Summary

* Fixes: #1833

Updated the error toast to now show proper error when name of a collective is longer than 255 characters.

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image| image

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint`)

### DIFF
--- a/src/components/Nav/NewCollectiveModal.vue
+++ b/src/components/Nav/NewCollectiveModal.vue
@@ -235,7 +235,7 @@ export default {
 
 	watch: {
 		newCollectiveName(val) {
-			if (!val || this.newCollectiveName.length < 3) {
+			if (val && this.newCollectiveName.length < 3) {
 				this.setNameIsTooShortDebounced()
 				this.setNameIsTooLongDebounced.clear()
 				this.nameIsTooShort = true


max-nextcloud:
@cyan-1 That looks pretty close to what I had in mind. I'm happy to hear you could try it out.

There's one issue left with the code though:
Now `this.state` will flip-flop - initially it is `0` when the function is called. Then it becomes `1` in the first if block, just to be changed to `0` again in the second if block.
In a lot of situations this would not really matter - but here `this.state` is reactive - so changing it may trigger all kinds of side effects - in fact I think it will try to render the group selection super quickly before returning to the name entry.
I think you can avoid this by swapping the if statements - so that you check for the long name first and then if the name is too long just call `return` from with in the if statement and the second if statement will not be reached.

In addition to this code fix we need you to signoff your commit. That means you declare that this is your own work and that the code can be used under the AGPL license that nextcloud/collectives uses. You can do this for your last commit by calling `git commit --amend --signoff --no-edit`:
* `--amend` means change the last commit rather than creating a new one.
* `--signoff` sign off the commit as i just explained.
* `--no-edit` do not open an editor for the commit message.

If you add a new commit - say for the change i just explained you can use `--signoff` when calling `git commit` or if you are committing with an IDE you can use the command again for the new commit. I assume there are ways to tell your IDE to signoff the commit right away as well.

If all commits are signed off the `DCO` check will be happy. Currently it's still red:
image

max-nextcloud:
@cyan-1 Code looks good. Let's see if the tests pass. I will test this tomorrow and then approve.

You could squash all commits into one and sign that one off. If you need some help with that just let us know.

cyan-1:
okay, so i messed up squashing the my past commits. i am unsure of where to go from here now. i require your assistance @max-nextcloud

cyan-1:
@max-nextcloud I tested how many charters gives us the exception, and the limit seems to be 127, so I've added a `nameisTooLong` just like already existing `nameIsTooShort` which should now fix the issue of trying to create collectives with long names

src/components/Nav/NewCollectiveModal.vue
```diff
@@ -200,12 +202,14 @@ export default {
...
 		nameIsInvalid() {
-			return this.nameIsTooShort || this.nameIsTaken
+			return this.nameIsTooShort || this.nameIsTaken || this.nameIsTooLong
 		},
 
 		nameError() {
 			if (this.nameIsTooShort) {
 				return t('collectives', 'Name too short, requires at least three characters')
+			} else if (this.nameIsTooLong) {
+				return t('collectives', 'Name too long, name cannot be longer than 127 characters')
```
mejo-:
```suggestion
				return t('collectives', 'Name too long, must not exceed 127 characters')
```

src/components/Nav/NewCollectiveModal.vue
```diff
@@ -231,13 +235,17 @@ export default {
 
 	watch: {
 		newCollectiveName(val) {
-			if (!val || this.newCollectiveName.length > 2) {
+			if (!val || (this.newCollectiveName.length > 2 && this.newCollectiveName.length < 128)) {
```
mejo-: I think the second check need to go into a separate `else if` block. Imagine that I had a name with more than 127 chars, so `nameIsToLong` is `true`. Then I remove all characters but one in a single step (by selecting all other characters and hitting <kbd>Del</kbd>). Now `newCollectiveName` is changed, so the watcher triggers. Now this combined condition is false, since the collective name has not more than two characters. So you'll end up with both `nameIsTooShort` and `nameIsTooLong` being `true`.

src/components/Nav/NewCollectiveModal.vue
```diff
@@ -231,13 +235,17 @@ export default {
...
 				this.setNameIsTooShortDebounced.clear()
+				this.setNameIsTooLongDebounced.clear()
 				this.nameIsTooShort = false
+				this.nameIsTooLong = false
 				return
```
mejo-:
```suggestion
```

Given that we do all changes within conditional blocks now (`this.setNameIsTooShortDebounced()` happens in the `else` block now), there's no need to return early any longer. So this `return` statement can be removed now.

src/components/Nav/NewCollectiveModal.vue
```diff
@@ -270,11 +278,19 @@ export default {
...
+		setNameIsTooLong() {
+			this.nameIsTooLong = true
+		},
+
 		clearName() {
 			this.name = ''
 		},
 
 		advanceToMembers() {
+			if (this.nameIsTooLong) {
```
mejo-: This should no longer be needed, as advancing to the next step is impossible due to `nameIsInvalid` being true if `nameIsTooLong`.

mejo- (changes requested):
Thanks a lot for tackling this @cyan-1, much appreciated! I have some smaller comments that I'd like to get addressed before this can be merged.

src/components/Nav/NewCollectiveModal.vue
cyan-1: resolved, latest commit

src/components/Nav/NewCollectiveModal.vue
```diff
@@ -231,13 +235,22 @@ export default {
 
 	watch: {
 		newCollectiveName(val) {
-			if (!val || this.newCollectiveName.length > 2) {
+			if (!val || this.newCollectiveName.length < 3) {
```
mejo-:
```suggestion
			if (val && this.newCollectiveName.length < 3) {
```

We only want to set `nameIsTooShort` if the new value of `newCollectiveName` is non-empty *and* shorter than three characters. An empty value should not show the error, as we don't want to display an error from the beginning (when no name was entered yet).

Issue #346, opened 2022-06-09
Show backlink and attachment counter in page info bar


mejo-:
Would be nice to show a counter of backlinks and attachments in the subline of page read mode (where the toolbar is in edit mode).

Notion does something similar:

![2022-06-09T16_45_58_874537721+02_00](https://gitlab.com/collectivecloud/collectives/uploads/39a032494f4dc43e3e5f206d6500b2d4/2022-06-09T16_45_58_874537721+02_00.png)

We have to consider possible performance implications though. Currently, getting the list of backlinks is rather expensive and only done one the page sidebar is opened. We probably would have to cache the counter somehow in future.

max-nextcloud:
For the backlink count we'd need an index database that gets updated on `FileChanged` events. It would index links to other pages so that computing backlinks becomes cheap:
* page_id
* linked_page_id

PR #2100, opened 2025-11-05
Cache backlinks in DB and show backlink counter in page info bar

mejo-:
### 📝 Summary

* Resolves: #346 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

### 🏁 Checklist

### DIFF
--- a/lib/Fs/MarkdownHelper.php
+++ b/lib/Fs/MarkdownHelper.php
@@ -37,6 +37,8 @@ class MarkdownHelper {
 	}
 
 	/**
+	 * Extracts markdown links and returns them with link text, href and title
+	 *
 	 * @throws CommonMarkException
 	 */
 	public static function getLinksFromContent(string $content): array {
@@ -73,11 +75,9 @@ class MarkdownHelper {
 	}
 
 	/**
-	 * @throws CommonMarkException
+	 * Returns hrefs that point to given collective or are relative links (.e.g. `../Page-21`)
 	 */
-	public static function getLinkedPageIds(Collective $collective, string $content, array $trustedDomains = []): array {
-		$links = self::getLinksFromContent($content);
-
+	private static function getHrefsToCollectiveFromLinks(array $links, Collective $collective, array $trustedDomains): array {
 		// Absolute URL regex
 		$protocol = '^https?:\/\/';
 		$trustedDomainsArray = array_map(static fn (string $domain) => str_replace('\*', '\w*', preg_quote($domain, '/')), $trustedDomains);
@@ -95,13 +95,7 @@ class MarkdownHelper {
 		]) . ')(?=\/+)';
 		$rootRelativeUrlRegex = '/^' . $basePath . $appPath . $collectivePath . '(.+)$/';
 
-		// slug may only contain ascii characters
-		$slugFragmentSuffix = '(?:[?#].*)?$/';
-		$slugPagePathRegex = '/\/[\x00-\x7F]+\-([0-9]+)' . $slugFragmentSuffix;
-		$noSlugFragmentSuffix = '(?:[&#].*)?$/';
-		$noSlugPagePathRegex = '/^.*[^?]+\?fileId=([0-9]+)' . $noSlugFragmentSuffix;
-
-		$pageIds = [];
+		$hrefs = [];
 		foreach ($links as $link) {
 			$href = $link['href'];
 			if (!$href) {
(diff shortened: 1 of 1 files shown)


cypress/e2e/page-details.spec.js
max-nextcloud, line 1: You are seeding more content including a link... but are you testing it somewhere?

lib/Db/PageLinkMapper.php
```diff
@@ -0,0 +1,103 @@
...
+				continue;
+			}
+
+			if (in_array($linkedPageId, $newLinkedPageIds, true)) {
+				$this->insertPageLink($pageId, $linkedPageId);
+			} else {
+				$this->deletePageLink($pageId, $linkedPageId);
+			}
+		}
+		$this->db->commit();
```
max-nextcloud: This seems overly complex. How about this:

```suggestion
		$addedLinkedPageIds = array_diff($newLinkedPageIds, $dbLinkedPageIds);
		$removedLinkedPageIds = array_diff($dbLinkedPageIds, $newLinkedPageIds);

		$this->db->beginTransaction();
		foreach ($addedLinkedPageIds as $linkedPageId) {
			$this->insertPageLink($pageId, $linkedPageId);
		}
		foreach ($removedLinkedPageIds as $linkedPageId) {
			$this->deletePageLink($pageId, $linkedPageId);
		}
		$this->db->commit();
```

Should even be possible to do at least the removal in one db query with a fitting helper function. Something like
```sql
DELETE FROM `collectives_page_links` WHERE `page_id`=... AND `linked_page_id` IN (1,2,3,...)
```

lib/Fs/NodeHelper.php
max-nextcloud, line 1: I would not expect these common mark imports in a node helper. Maybe use a different helper for the common mark processing?

lib/Fs/NodeHelper.php
```diff
@@ -251,4 +260,122 @@ public static function folderHasSubPage(Folder $folder, string $title): int {
...
+
+		return $links;
+	}
+
+	/**
+	 * @throws CommonMarkException
+	 */
+	public static function getLinkedPageIds(Collective $collective, string $content, array $trustedDomains = []): array {
+		$links = self::getLinksFromContent($content);
+
```
max-nextcloud: I think this would benefit from being split into smaller functions. Something like


```suggestion
	public static function getLinkedPageIds(Collective $collective, string $content, array $trustedDomains = []): array {
		$links = self::getLinksFromContent($content);
		$hrefs = self::getHrefsFromLinks($links, $collective, $trustedDomains);
		$pageIds = self::getPageIdsFromHrefs($hrefs);
		return array_unique($pageIds, SORT_NUMERIC);

```

lib/Migration/MigrateBacklinks.php
```diff
@@ -0,0 +1,92 @@
...
+				$collectiveFolder = $this->collectiveFolderManager->getFolder($collective->getId());
+			} catch (InvalidPathException|FilesNotFoundException) {
+				$output->warning('Failed to migrate templates for collective with id ' . $collective->getId());
+				continue;
+			}
+
+			$pageFiles = $this->pagesFromFolder($collectiveFolder);
+
+			foreach ($pageFiles as $pageFile) {
+				$linkedPageIds = NodeHelper::getLinkedPageIds($collective, $pageFile->getContent(), $this->config->getSystemValue('trusted_domains', []));
```
max-nextcloud: It probably does not matter as the config is cached... but maybe still read the config domains just once before the outer `foreach` loop and reuse it here.

lib/Db/PageLinkMapper.php
mejo-: Nice catch! I didn't feel very happy with the code anyway and your approach is much cleaner. I also merged the delete statements by using `->andWhere($qb->expr()->in('linked_page_id', $qb->createNamedParameter($linkedPageIds)));`

lib/Migration/MigrateBacklinks.php
mejo-: Good point, fixed now.

lib/Fs/NodeHelper.php
mejo-: Agreed, moved to `lib/Fs/MarkdownHelper.php`.

lib/Fs/NodeHelper.php
mejo-: Also agreed and changed now :)

PR #2102, opened 2025-11-05
Fix: Avoid overwriting collective list when opening a shared link

Koc:
### 📝 Summary
This PR fixes next bug:
1. Create few collectives, e.g. A, B, C
2. Create public link for A
3. Open A in a separate tab by this public link

Expected result:
Collectives list in original tab has A, B, C

Actual result:
Collectives list in original tab has only A

Fixes: #2131 

### 🏁 Checklist


src/stores/pages.js
```diff
@@ -88,10 +105,6 @@ export const usePagesStore = defineStore('pages', {
 			return state.currentPage.fileName === INDEX_PAGE + PAGE_SUFFIX
 		},
 
-		rootPageForCollective(state) {
```
mejo-, line 91: This was an obsolete unused getter so I removed it.

mejo-:
> @mejo- I've just tested few of my scenarios and now it works really well 👏 . Thank you 🙏

That's great to hear. Thanks again for the great contribution and collaboration, @Koc, it's really appreciated :heart:

Issue #2050, opened 2025-10-16
feat: Use Nextcloud database for TNTSearch


benjaminfrueh:
Use the Nextcloud database for the TNTSearch used for searching in collectives.

PR #2118, opened 2025-11-11
feat: implement doctrine-based fulltext search

benjaminfrueh:
### 📝 Summary

Replaces TNTSearch with nextcloud database full-text search using doctrine.

Resolves https://github.com/nextcloud/collectives/issues/2050

### Todo

* [x] Implement language detection in stemmer
* [x] Scope `php-stemmer` dependency using `php-scoper`
* [x] Remove documentation about extra tntsearch dependencies
* [x] Migrate e2e FTS tests to Playwright
* [ ] Follow-up: add Arabic stemmer

### 🏁 Checklist


composer.json
```diff
@@ -19,7 +19,7 @@
         "league/commonmark": "^2.7",
         "symfony/string": "^6.0",
         "symfony/translation-contracts": "^3.6",
-        "teamtnt/tntsearch": "^5.0"
+        "wamania/php-stemmer": "^4.0"
```
juliusknorr, line 22: We use this one in a few other apps: https://github.com/search?q=org%3Anextcloud+wamania%2Fphp-stemmer&type=code

Maybe worth to consider scoping the dependency to avoid conflicts with different versions between apps. https://arthur-schiwon.de/isolating-nextcloud-app-dependencies-php-scoper

lib/Search/FileSearch/Db/SearchDoc.php
```diff
@@ -0,0 +1,31 @@
...
+ * @method string getCircleUniqueId()
+ * @method void setCircleUniqueId(string $value)
+ * @method string getWordId()
+ * @method void setWordId(string $value)
+ * @method int getFileId()
+ * @method void setFileId(int $value)
+ * @method int getHitCount()
+ * @method void setHitCount(int $value)
+ */
+class SearchDoc extends Entity {
```
mejo-, line 26: So far we tried to name the objects similar to the table names. So Maybe either rename the tables to `collectives_search_*` or rename the object classes to `FtsDoc` and so on? 🤔

lib/Search/FileSearch/Db/SearchDocMapper.php
```diff
@@ -0,0 +1,73 @@
...
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Collectives\Search\FileSearch\Db;
+
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+use OCP\Snowflake\IGenerator;
```
mejo-: This will not be compatible with Nextcloud 32, which we still intend to support with the main Collectives release as long as it's officially supported. Maybe we should stick with normal autoincrement IDs for now and migrate all IDs to snowflake IDs at a later point after we dropped Nextcloud 32 support, what do you think? We probably have to do this in other DB tables as well anyway.

lib/Search/FileSearch/Stemmer/Stemmer.php
```diff
@@ -0,0 +1,47 @@
...
+	private bool $stemmingEnabled = true;
+
+	public function __construct(
+		private IConfig $config,
+	) {
+	}
+
+	public function stem(string $word): string {
+		if ($this->stemmer === null && $this->stemmingEnabled) {
+			$language = $this->config->getSystemValue('default_language', 'en');
```
mejo-: I'm not sure whether it's the best option to use the instance's default language here. This effectively means that stemming only happens for instance's default language? I guess this would be much more powerful if the language of the indexed document was used here, right?

Maybe there's simple algorithms to guess language from the full document in the indexer and pass the detected language into the stemmer? Nothing that necessarily needs to happen in this PR, but still I'd be curious about your thoughts.

lib/Search/FileSearch/Stemmer/Stemmer.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Collectives\Search\FileSearch\Stemmer;
+
+use OCP\IConfig;
+use Wamania\Snowball\NotFoundException;
```
mejo-, line 13: I like that we use a third-party stemmer library. I compared the supported languages and TNTSearch supports the following languages that Wamania php-stemmer does not: Arabic, Croatian, Latvian, Polish and Ukrainian and a [PorterStemmer](https://github.com/teamtnt/tntsearch/blob/master/src/Stemmer/PorterStemmer.php) where I'm not sure about it's purpose.

Personally I think having support for Arabic would be nice, but it seems easy enough to add that to Wamania in a PR and we should consider it as a follow-up task for now.

lib/Search/FileSearch/Tokenizer/WordTokenizer.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace OCA\Collectives\Search\FileSearch\Tokenizer;
+
+class WordTokenizer {
+	private const MIN_LENGTH = 2;
+	private const MAX_LENGTH = 50;
+
+	public function tokenize(string $text): array {
+		$text = mb_strtolower($text);
+		$text = preg_replace('/([^\p{L}\p{N}@])+/u', ' ', $text);
```
mejo-: The tokenizer replace regex of TNTSearch also includes `_` and `-` as characters that should not be replaced:

```php
$pattern = '/[^\p{L}\p{N}\p{Pc}\p{Pd}@]+/u';
```

Was it on purpose that you removed these to characters? Just asking out of curiosity.

lib/Search/FileSearch/FileIndexer.php
mejo-: We only use `.md` files for pages in Collectives. Maybe it's better to use [`NodeHelper::isPage()`](https://github.com/nextcloud/collectives/blob/ee683a2d3c6a829a442b07fc2924f78f30ece0cf/lib/Fs/NodeHelper.php#L180-L182) as we already use it in other places.

lib/Search/FileSearch/Db/SearchWordMapper.php
```diff
@@ -0,0 +1,106 @@
...
+		$qb->executeStatement();
+
+		$this->deleteZeroHitWords($circleUniqueId);
+	}
+
+	private function deleteZeroHitWords(string $circleUniqueId): void {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->tableName)
+			->where($qb->expr()->eq('circle_unique_id', $qb->createNamedParameter($circleUniqueId)))
+			->andWhere($qb->expr()->lte('num_hits', $qb->createNamedParameter(0)));
```
mejo-: Maybe also remote entries where `num_files <= 0`?

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -3,126 +3,106 @@
...
+		if (empty($stems)) {
+			return [];
 		}
 
-		if (!$path) {
-			throw new FileSearchException('File searcher could not create local index.');
+		$wordIds = [];
+		foreach ($stems as $stem) {
+			$word = $this->wordMapper->findByCircleAndTerm($circleId, $stem);
+			$words = $word ? [$word] : $this->fuzzySearchWord($circleId, $stem);
```
mejo-, line 53: After reading and searching a bit about this topic, I wonder whether it's a good idea to do fuzzy search on stems instead of the original tokens. This seems to lead to much more false positives, as stems are usually rather short terms. An alternative would be to tighten `FUZZY_MAX_DISTANCE` to `1`, but I think doing fuzzy search on the original token would be the better approach:

I didn't check TNTSearch on this regard, but I realized in the past that we have lots of false positives in full-text search in Collectives. Maybe that was the reason. 🤔

lib/Search/FileSearch/FileIndexer.php
```diff
@@ -3,36 +3,94 @@
...
+		foreach ($tokens as $token) {
+			$stems[] = $this->stemmer->stem($token);
+		}
+
+		$terms = array_count_values($stems);
+		unset($content, $tokens, $stems);
+
+		foreach ($terms as $term => $hitCount) {
+			try {
+				$term = substr((string)$term, 0, 50);
```
mejo-: Maybe better use `mb_substr()` here as well to avoid truncating in the middle of a multi-byte character.

lib/Search/FileSearch/Db/SearchWordMapper.php
```diff
@@ -0,0 +1,106 @@
...
+
+		return $this->findEntities($qb);
+	}
+
+	public function decrementCounts(string $circleUniqueId, string $wordId, int $hitCount): void {
+		$qb = $this->db->getQueryBuilder();
+
+		$hitCountParam = $qb->createNamedParameter($hitCount, IQueryBuilder::PARAM_INT);
+		$qb->update($this->tableName)
+			->set('num_hits', $qb->createFunction("num_hits - $hitCountParam"))
```
mejo-: I wonder whether we want to save-guard against negative values here. Especially as the fields are unsigned integers, which means here's a risk of buffer overflows, right?

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -3,126 +3,106 @@
...
-		'ru' => 'Russian',
-		'uk' => 'Ukrainian',
-	];
-
-	private const UNSUPPORTED_LANGUAGE = 'No';
-
-	protected FileIndexer $indexer;
+class FileSearcher {
+	private const DEFAULT_LIMIT = 15;
+	private const FUZZY_PREFIX_LENGTH = 2;
```
mejo-: A prefix length of 2 seems pretty short and probably results in many candidates in many cases. How about increasing this to 3 or 4?

lib/Search/FileSearch/Tokenizer/ClauseTokenizer.php
```diff
@@ -0,0 +1,31 @@
...
+namespace OCA\Collectives\Search\FileSearch\Tokenizer;
+
+class ClauseTokenizer extends WordTokenizer {
+	public function tokenize($text): array {
+		$words = parent::tokenize($text);
+
+		$tokens = [];
+		$lastWord = '';
+		foreach ($words as $word) {
+			if (strlen($word) <= 3) {
```
mejo-:
```suggestion
			if (mb_strlen($word) <= 3) {
```

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -3,126 +3,106 @@
...
+			return $files;
 		}
-		$this->index = new PDO('sqlite:' . $indexName);
-		$this->index->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
 
-		$this->indexer->setIndex($this->index);
-		$this->indexer->setLanguage($this->language ?? self::UNSUPPORTED_LANGUAGE);
+		$scored = [];
+		foreach ($files as $file) {
+			$content = mb_strtolower($file->getContent());
```
mejo-: This looks like a potential performance problem to me, as it will re-read the contents of all matched files. I don't know though what to do about it. The ranking by bigrams seams important as it makes sure that searches for several words ranks results higher where these words are next to each other. So dropping sorting by bigrams is not an option in my opinion.

We could add a fourth DB table and compute bigram phrase counts in it at index time. But this would probably drastically increase the database size and might be a bit over-engineered.

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: We are currently only storing the stems in the database. Storing the original tokens would take significantly more space and impact performance as well. The fuzzysearch is maybe even redundant with the stemmer already handling minor variations like plurals, but maybe we should keep it with `FUZZY_MAX_DISTANCE`  `1` for smaller typos.

TNTSearch was using a fuzzy_distance of 2 per default, leading to more false positives.

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: I agree that this could be a performance issue, but as you said there is no good solution without bloating the DB by a lot and also slow the indexing. The performance impact should be limited by the `DEFAULT_LIMIT` of 15 documents which are searched ranked by hits and then ranked by bigrams.

TNTSearch offered bigram and various ngram tokenizers, neither of them were configured or previously used for collectives, they would increase the database size drastically. They stored them just as word pairs alongside their words.

I think we should test the performance impact with a large dataset first and then decide on this.

lib/Search/FileSearch/Db/SearchDoc.php
benjaminfrueh: As discussed we will prefix them with collectives_s_ for now, to be consistent with the other tables. The stable32 support currently limits the table and index length.

lib/Search/FileSearch/Stemmer/Stemmer.php
benjaminfrueh: The PorterStemmer should be the default english stemmer named after Martin Porter.
I think the Wamania stemmer library is the commonly used one and already used in other nextcloud apps, https://github.com/search?q=org%3Anextcloud+wamania%2Fphp-stemmer&type=code

We can add a PR for an arabic stemmer as a follow-up task.

lib/Search/FileSearch/Tokenizer/WordTokenizer.php
benjaminfrueh: I have seen both versions for word tokenizers, but you are correct, current TNT-Search does not split words at `-` and `_` so it will index these as single words. I will add it back to the pattern in favour for making the search results more explicit. It is individual, as there are words where splitting makes sense to find them better (e. g. `ẁell-known` can then be found by searching `known`, and words where splitting would be bad (e. g. `e-mail`, the `e` would not even get indexed)

I also added the stopword option back to the tokenizer, just in case we want to remove common and meaningless words from the index.

lib/Search/FileSearch/Db/SearchWordMapper.php
benjaminfrueh: I added the `num_files <= 0` as an `OR` condition, and renamed the function to `deleteOrphanedWords` to reflect it's new purpose.

lib/Search/FileSearch/FileIndexer.php
benjaminfrueh: Thanks for the feedback, updated it.

lib/Search/FileSearch/FileIndexer.php
benjaminfrueh: Thanks for the feedback, updated it.

lib/Search/FileSearch/Db/SearchWordMapper.php
benjaminfrueh: Thanks for the feedback, updated it.

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: Thanks for the feedback, I increased it to 3 for now.

lib/Search/FileSearch/Tokenizer/ClauseTokenizer.php
benjaminfrueh: Thanks for the feedback, updated it.

lib/Search/FileSearch/Stemmer/Stemmer.php
benjaminfrueh: Language detection per document seems like the only proper solution, it comes with a small performance downside for indexing, but that should be fine. I found that there are language detection libraries, like https://github.com/patrickschur/language-detection which could be used, what do you think?

We should then store the language in the `collectives_s_files` table, so the correct stemmer can be used for each document.

There are possible edge-cases that document language change or a document has mixed languages, but we would just have to save the first one we detect.

lib/Search/FileSearch/Db/SearchDocMapper.php
benjaminfrueh: I changed it to autoincrement IDs for now.

lib/Search/FileSearch/Db/SearchWordMapper.php
mejo-: I'm a bit unsure whether that really is enough to safeguard against negative `num_hits` and `num_files` values. If I understand it correctly, trying to put negative values into unsigned tables fields can result in buffer overflows. What was your motivation to make the integer/bigint fields unsigned in the first place? That we don't expect negative values here?

How about this to safeguard?

```php
$qb->update($this->tableName)
	->set('num_hits', $qb->func()->greatest($qb->createFunction("num_hits - $hitCountParam"), 0))
	->set('num_files', $qb->func()->greatest($qb->createFunction('num_files - 1'), 0))
	->where($qb->expr()->eq('circle_unique_id', $qb->createNamedParameter($circleUniqueId)))
	->andWhere($qb->expr()->eq('id', $qb->createNamedParameter($wordId)));
```

benjaminfrueh: Yes exactly, as `num_hits` and `num_files` are counts that semantically should never be negative, unsigned enforces that at the DB level. Of course this can lead to buffer overflows and as you say should be safeguarded.

My safeguard was preventing the decrement update if `num_hits` would underflow. Potentially I think it could leave stale data.

Wrapping it in a `greatest()` will still cause the overflow in MySQL/MariaDB, as it already happens in the subtraction, before the `greatest()` is evaluated. Can be tested with this query, which will cause a overflow error:
```sql
UPDATE oc_collectives_s_words SET num_files = GREATEST(num_files - 2, 0) WHERE num_files = 1;
```

`num_files` should always be >= 1

I like the idea of just clamping the `num_hits` to 0 and `num_files` to 1 with a `greatest()` function, to be self-healing and cover all the data inconsistencies and edge cases we didn't think about now.

So maybe we should change the columns to signed and then use a greatest function, as this is only enforced in MySQL/MariaDB anyways as far as I know. This seems more safe, what do you think?

mejo-: Sounds like a good plain :blush: Let's use signed columns for `num_hits` and `num_files` and clamp the values using the `greatest()` function :ok_hand:

benjaminfrueh: I updated this in commit https://github.com/nextcloud/collectives/pull/2118/changes/b9a3c57ef09c7a1a008217a4e180ebb0b4ef862f

All integers columns are now signed, for overflow safety and also to be consistent with the other collectives tables. The `num_hits` and `num_files` are now clamped to `0` using `greatest()`.

The `deleteOrphanedWords()` function already deletes words with `num_hits` OR `num_files` <= 0 and is called right after the decrement, so this should be good against 0 and negative values here anyways.

lib/Search/FileSearch/Stemmer/Stemmer.php
benjaminfrueh: @mejo- I updated the PR and added language detection in this commit: https://github.com/nextcloud/collectives/pull/2118/changes/7e969f26cce050f8787144b3a25e1098cc9e5e58

- added `patrickschur/language-detection` to detect the language of each document
- added a `language` column to `collectives_s_files`to store the detected language per file
- updated `Stemmer` to accept an optional `$language` parameter and cache stemmers per language
- updated `FileIndexer`to detect and save language during indexing in `collectives_s_files`
- detection during indexing is done on the first `2000` characters, defined as `LANGUAGE_DETECTION_LIMIT` to limit performance impact
- updated `FileSearcher` to search and stem all languages present in the collective, so search and stemming works with multilingual collectives

composer.json
benjaminfrueh: Added dependency scoping in commit https://github.com/nextcloud/collectives/pull/2118/changes/83f65aa6851e4bf714a010da06905f9799761916

mejo-: In the end we decided against introducing dependency scoping as it created to much problems for now, especially in combination with the dev dependency scoping via `composer-bin-plugin`.

Issue #1566, opened 2024-11-04
Public shares inherit default expiry date from file shares


mejo-:
**Describe the bug**
When a default expiry date is set for file shares, this will apply to collective shares as well. There's no way to circumvent, extend or disable this in the Collectives app.

**To Reproduce**
1. Set a default expiry date for file shares
2. Create a collective share
3. Have a look into the file share details at "Collectives/<shared_collective" and see the expiry date
4. Wait until the expiry and see that the collective share is gone

**Expected behavior**
We should either enforce file shares without expiry or make the share expiry configurable in Collectives.

PR #2126, opened 2025-11-20
feat: Add expiry date support for collective shares in backend

silverkszlo:
### 🚧 Work in progress 🚧

Add ability to view and manage expiration dates on collective shares. Shares now inherit and respect Nextcloud's default expiry date policy for file shares.

- Add expirationDate field to CollectiveShare entity (transient)
- Extract expiry dates from underlying IShare objects in service layer
- Add helper methods for expiry configuration (hasDefaultExpireDate, isExpireDateEnforced, getDefaultExpireDays)
- Update API endpoints to accept and return expiry date information
- Support setting, updating, and removing expiry dates via updateShare
- Return both share data and global expiry config in getCollectiveShares

### 📝 Summary

* Resolves: #1566 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### DIFF
--- a/lib/Service/CollectiveShareService.php
+++ b/lib/Service/CollectiveShareService.php
@@ -125,11 +125,7 @@ class CollectiveShareService {
 			$collectiveShare->setPassword($folderShare->getPassword());
 		}
 
-		if ($folderShare->getExpirationDate()) {
-			$collectiveShare->setExpirationDate($folderShare->getExpirationDate());
-		} else {
-			$collectiveShare->setExpirationDate(null);
-		}
+		$collectiveShare->setExpirationDate($folderShare->getExpirationDate());
 		return $collectiveShare;
 	}
 
@@ -157,11 +153,7 @@ class CollectiveShareService {
 			$collectiveShare->setPassword($folderShare->getPassword());
 		}
 
-		if ($folderShare->getExpirationDate()) {
-			$collectiveShare->setExpirationDate($folderShare->getExpirationDate());
-		} else {
-			$collectiveShare->setExpirationDate(null);
-		}
+		$collectiveShare->setExpirationDate($folderShare->getExpirationDate());
 		return $collectiveShare;
 	}
 
@@ -181,12 +173,7 @@ class CollectiveShareService {
 				$share->setPassword($folderShare->getPassword());
 			}
 
-			$expirationDate = $folderShare->getExpirationDate();
-			if ($expirationDate) {
-				$share->setExpirationDate($expirationDate);
-			} else {
-				$share->setExpirationDate(null);
-			}
+			$share->setExpirationDate($folderShare->getExpirationDate());
 			$shares[] = $share;
 		}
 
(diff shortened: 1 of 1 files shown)


lib/Service/CollectiveShareService.php
```diff
@@ -124,6 +124,12 @@ public function findShare(string $userId, int $collectiveId, int $pageId): ?Coll
 		if ($folderShare->getPassword()) {
 			$collectiveShare->setPassword($folderShare->getPassword());
 		}
+
+		if ($folderShare->getExpirationDate()) {
+			$collectiveShare->setExpirationDate($folderShare->getExpirationDate());
+		} else {
+			$collectiveShare->setExpirationDate(null);
+		}
```
mejo-: Seems like `$folderShare->getExpirationDate()` already returns `null` for unset expiration dates. So I guess it should be save to do `$collectiveShare->setExpirationDate($folderShare->getExpirationDate())` straight away.

lib/Service/CollectiveShareService.php
```diff
@@ -150,6 +156,12 @@ public function findShareByToken(string $token, bool $getPermissionInfo = true):
...
```
mejo-: Same here.

lib/Service/CollectiveShareService.php
```diff
@@ -168,6 +180,13 @@ public function getSharesByCollectiveAndUser(string $userId, int $collectiveId):
 			if ($folderShare->getPassword()) {
 				$share->setPassword($folderShare->getPassword());
 			}
+
+			$expirationDate = $folderShare->getExpirationDate();
+			if ($expirationDate) {
+				$share->setExpirationDate($expirationDate);
+			} else {
+				$share->setExpirationDate(null);
+			}
```
mejo-: Same here.

lib/Service/CollectiveShareService.php
```diff
@@ -250,10 +281,28 @@ public function updateShare(string $userId, Collective $collective, ?PageInfo $p
...
 
 		$share->setEditable($this->isShareEditable($folderShare));
 		$share->setPassword($folderShare->getPassword() ?? '');
+
+		if ($folderShare->getExpirationDate()) {
+			$share->setExpirationDate($folderShare->getExpirationDate());
+		} else {
+			$share->setExpirationDate(null);
+		}
+
```
mejo-, line 289: Same here.

lib/Service/CollectiveShareService.php
```diff
@@ -179,6 +198,18 @@ private function isShareEditable(IShare $folderShare): bool {
 		return ($folderShare->getPermissions() & Collective::editPermissions) === Collective::editPermissions;
 	}
 
+	public function isExpireDateEnforced(): bool {
```
mejo-, line 188: I wonder whether this would need to be taken into account as well in `createShare()` 🤔 

Ah I guess the `folderShare` gets a default expiry set automatically if it's enforced by the server configuration. But then at least we need a `$collectiveShare->setExpirationDate($folderShare->getExpirationDate())` as well at the end of `createShare()` to make sure the returned share reflects this.

mejo- (commented):
Looks really good already. I have left some remarks.

PR #2168, opened 2025-12-18
chore(deps): upgrade cypress to latest with fast visiblity

max-nextcloud:
See https://github.com/cypress-io/cypress/issues/27415#issuecomment-3662601982


mejo-:
@max-nextcloud given that we still have https://github.com/nextcloud/collectives/pull/2062 open anyway, shall we close this PR again?

PR #2196, opened 2026-01-12
Add support to import a markdown directory

mejo-:
### 📝 Summary

This PR adds a new occ command `occ collectives:import:markdown` to import markdown files and referenced attachments from a given directory path.

The import function first imports all markdown files and then tries to fix relative links and references to local attachment files.

So far it's only tested with markdown created by [Dokuwiki2Markdown](https://github.com/mm503/Dokuwiki2Markdown) as the aim is to add support for migrating a Dokuwiki instance to Collectives.

The PR contains documentation how to use the occ command.

### 🏁 Checklist

### DIFF
--- a/lib/Command/IndexCollectives.php
+++ b/lib/Command/IndexCollectives.php
@@ -57,10 +57,10 @@ class IndexCollectives extends Command {
 				$output->writeln('<info>done</info>');
 			} catch (MissingDependencyException|NotFoundException|NotPermittedException) {
 				$output->writeln("<error>Failed to find team associated with collective with ID={$collective->getId()}</error>");
-				return 1;
+				continue;
 			} catch (FileSearchException) {
 				$output->writeln('<error>Failed to save the indices to the collectives folder.</error>');
-				return 1;
+				continue;
 			}
 		}
 


mejo-:
> Just a potential papercut/confusion:
> Links in the DW source, that don't exist anymore (in DW itself) won't get parsed/translated and remain in the :namespace:page format. The target pages don't exist anyway, so there is no problem, but I thought maybe one could do something around that? Just for the notes :)

Do you have an idea what could be done about the links? I'm hesitant to touch links that cannot be resolved to an existing page as we don't know what to do with them. I'd prefer to keep them as is to be honest.

mejo-:
@max-nextcloud I did all the changes we discussed.

lib/Command/ImportMarkdownDirectory.php
```diff
@@ -0,0 +1,108 @@
...
+			->addOption('parent-id', 'p', InputOption::VALUE_REQUIRED, 'Parent page ID for the import (0 for root)', '0');
+		parent::configure();
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$collectiveId = (int)$input->getOption('collective-id');
+		$directory = $input->getArgument('directory');
+		$userId = $input->getOption('user-id');
+		$parentId = (int)$input->getOption('parent-id');
+		$verbose = (bool)$input->getOption('verbose');
```
benjaminfrueh: Nitpick: `$verbose` seems redundant here as the Symfony OutputInterface already tracks verbosity.
In the `ProgressReporter` this could be used either from `$output->isVerbose()` or using the `writeln()` option directly like `$this->output->writeln('<info>' . $message . '</info>', OutputInterface::VERBOSITY_VERBOSE);`

mejo-: Ah nice, didn't know that. I'll keep $verbose in IProgressReporter for now, as the idea is to maybe add support for importing via API later on and then we will need a second implementation of ProgressReporter that also has to pass in verbosity.

PR #2198, opened 2026-01-12
fix(PageActionMenu): Fix download link in page list

mejo-:
### 🏁 Checklist


max-nextcloud (approved):
Makes sense.

Issue #2189, opened 2026-01-08
Collectives cannot be shared with users with zero quota


bentuna:
**Describe the bug**
When a user has zero or negative quota, they cannot create any files or folders. Thus the "Collectives" folder cannot be created in their root directory and the sharing of a collective with them fails.

**To Reproduce**
1. Create a new user and give 0 B quota to them
2. Log in to another user and share a Collective with the newly created user
3. Log in to the new user for the first time and try to open the shared collective
4. See error "Could not fetch collecitve pages"

**Expected behavior**
The collective should be able to share with users that have no quota (left) because they don't need any of their quota to view or edit the collective when the collective is owned by another user.

**Screenshots**
<img width="1295" height="430" alt="Image" src="https://github.com/user-attachments/assets/407035de-9f0f-4dc7-89cb-9e9b20a96d09" />

**Server details:**
- Collectives app version: 3.4.0
- Nextcloud version: 32.0.3

mejo-:
We fixed it for guest users with commit a0b4b6038a7f3e9ae05e90bfd7e06b4e662884ca. So in theory it should work for all users where `$user->getQuota()` returns "0 B".

@bentuna could you check whether that's the case for your user? Maybe we need to adjust that and be more relaxed with the check. I remember that Nextcloud user quota string can vary quite a bit. Probably it can also be just `0` or `0b` or `0 b` or `0B` 🤔

bentuna:
Thank you @mejo- ! We will patch it and test it. But does it also account for users that used all their quota?

> Also, we need to account for the case that a user has a quota above zero but used all of it. So we need to check the free quota instead of $user->getQuota() right?

PR #2200, opened 2026-01-13
feat(mount): Make collectives user folder a mount point

mejo-:
Turns the collectives user folder into a mount point that uses a custom empty storage. The storage is read-only.

This fixes several issues:

* No more issues with 0 or fully used quota.
* No longer possible to put unrelated nodes in the collectives user folder.

Existing nodes with the name of the mountpoint are renamed before mounting collectives folder.

Fixes: #2201
Fixes: #2189

### 🏁 Checklist

### DIFF
--- a/tests/Integration/features/bootstrap/FeatureContext.php
+++ b/tests/Integration/features/bootstrap/FeatureContext.php
@@ -731,6 +731,11 @@ class FeatureContext implements Context {
 		} else {
 			$this->assertStatusCode(400);
 		}
+
+		if ($key === 'user_folder') {
+			// Workaround some weird caching/timing issue by resetting user session after changing user_folder setting
+			unset($this->cookieJars[$user], $this->requestTokens[$user]);
+		}
 	}
 
 	/**


mejo-:
@provokateurin especially `lib/Mount/CollectivesUserStorage.php` would be great to get a thorough review.

I mostly copied [`lib/private/Lockdown/Filesystem/NullStorage.php` from server](https://github.com/nextcloud/server/blob/master/lib/private/Lockdown/Filesystem/NullStorage.php) and adjusted it to my needs. Main differences are in the functions `opendir()`, `isReadable()`, `getPermissions()`, `filemtime()`, `getMimeType()`, `getETag()`, `getOwner()` and that it's using `getCache()` from its parent instead of returning `NullCache` there.

I'm particularly unsure about `opendir()` and the caching, but it works as expected in my tests :grimacing:

lib/Mount/CollectivesUserStorage.php
```diff
@@ -0,0 +1,180 @@
...
+	public function mkdir(string $path): never {
+		throw new \OC\ForbiddenException('This request is not allowed to access the filesystem');
+	}
+
+	public function rmdir(string $path): never {
+		throw new \OC\ForbiddenException('This request is not allowed to access the filesystem');
+	}
+
+	public function opendir(string $path): IteratorDirectory {
+		return new IteratorDirectory();
```
provokateurin, line 43: Shouldn't this return the collectives? Then you could get rid of the per-collective mount points. Depending on complexity and amount of refactoring required, this might be a better solution in the long term.

mejo-: Thanks, valid point. I'd like to postpone this though, as it brings more complexity. E.g. permissions can be different per collective, but currently permissions are handled via the `CollectiveMountPoint`. So if all this is merged into one Storage, all relevant functions (like `isCreatable()`, `isShareable()`, ...) would need to check depending on the given path.

PR #2216, opened 2026-01-22
fix(MountProvider): Don't setup filesystem for conflict resolution

mejo-:
Follow-up for https://github.com/nextcloud/collectives/pull/2200.

Trying to access a node with the mountpoint name results in a circular situation. This commit changes the implementation to only use the cache.

Implementation inspired by how groupfolders does it.

### 🏁 Checklist

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -152,12 +152,12 @@ class MountProvider implements IMountProvider {
 		$filesPathHash = md5($filesPath);
 
 		$query = $this->connection->getQueryBuilder();
-		$query->select('path')
+		$query->select($query->func()->count('path'))
 			->from('filecache')
 			->where($query->expr()->eq('storage', $query->createNamedParameter($userHome->getNumericStorageId(), IQueryBuilder::PARAM_INT)))
 			->andWhere($query->expr()->eq('path_hash', $query->createNamedParameter($filesPathHash, IQueryBuilder::PARAM_STR)));
 
-		if ($query->executeQuery()->rowCount() === 0) {
+		if ($query->executeQuery()->fetchOne() === 0) {
 			// No conflicting entry found
 			return;
 		};
@@ -171,11 +171,10 @@ class MountProvider implements IMountProvider {
 		$isEmptyDir = false;
 		if ($userFolderCacheEntry !== false && $userFolderCacheEntry->getMimeType() === 'httpd/unix-directory') {
 			$query = $this->connection->getQueryBuilder();
-			$query->select('path')
+			$query->select($query->func()->count('path'))
 				->from('filecache')
-				->where($query->expr()->eq('parent', $query->createNamedParameter($userFolderCacheEntry->getId(), IQueryBuilder::PARAM_INT)))
-				->setMaxResults(1);
-			if ($query->executeQuery()->rowCount() === 0) {
+				->where($query->expr()->eq('parent', $query->createNamedParameter($userFolderCacheEntry->getId(), IQueryBuilder::PARAM_INT)));
+			if ($query->executeQuery()->fetchOne() === 0) {
 				$isEmptyDir = true;
 			}
 		}


lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+		$filesPath = 'files/' . $path;
+		$filesPathHash = md5($filesPath);
+
+		$query = $this->connection->getQueryBuilder();
+		$query->select('path')
+			->from('filecache')
+			->where($query->expr()->eq('storage', $query->createNamedParameter($userHome->getNumericStorageId(), IQueryBuilder::PARAM_INT)))
+			->andWhere($query->expr()->eq('path_hash', $query->createNamedParameter($filesPathHash, IQueryBuilder::PARAM_STR)));
+
+		if ($query->executeQuery()->fetchOne() === false) {
```
provokateurin: I personally prefer fetching the row count and checking that

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
 		);
 	}
+
+	private function resolveNameConflict(IUser $user, string $path): void {
+		$userHome = $this->mountProviderCollection->getHomeMountForUser($user);
+		$filesPath = 'files/' . $path;
+		$filesPathHash = md5($filesPath);
+
+		$query = $this->connection->getQueryBuilder();
+		$query->select('path')
```
provokateurin: You could also make this a single query. Since this method will be called every filesystem setup, it might be a good idea for performance.

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+		};
+
+		/** @var IStorage $userHomeStorage */
+		$userHomeStorage = $userHome->getStorage();
+		$userHomeCache = $userHomeStorage->getCache();
+		$userFolderCache = $userHomeCache->get($filesPath);
+
+		// Check if node is a folder and empty
+		$isEmptyDir = false;
+		if ($userFolderCache && $userFolderCache['mimetype'] === 'httpd/unix-directory') {
```
provokateurin:
```suggestion
		if ($userFolderCache !== false && $userFolderCache->getMimeType() === 'httpd/unix-directory') {
```

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+			$query = $this->connection->getQueryBuilder();
+			$query->select('path')
+				->from('filecache')
+				->where($query->expr()->eq('storage', $query->createNamedParameter($userHome->getNumericStorageId(), IQueryBuilder::PARAM_INT)))
+				->andWhere($query->expr()->like('path', $query->createNamedParameter($filesPath . '/%', IQueryBuilder::PARAM_STR)))
+				->setMaxResults(1);
+			if ($query->executeQuery()->fetchOne() === false) {
```
provokateurin: You should rather check where a row has the parent set to the folder. That will perform much better than this.

provokateurin: Again, I'd prefer counting the rows.

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+				$isEmptyDir = true;
+			}
+		}
+
+		if ($isEmptyDir) {
+			// Delete empty folder
+			$userHomeStorage->unlink($filesPath);
```
provokateurin:
```suggestion
			$userHomeStorage->rmdir($filesPath);
```

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+			$userHomeCache->remove($filesPath);
+			$userHomeStorage->getPropagator()->propagateChange($filesPath, time());
+		} else {
+			// Rename node
+			$i = 1;
+			$folderName = $path . ' (' . $i++ . ')';
+			while ($userHomeCache->inCache('files/' . $folderName)) {
+				$folderName = $path . ' (' . $i++ . ')';
```
provokateurin:
```suggestion
			$folderName = $path . ' (' . $i . ')';
			while ($userHomeCache->inCache('files/' . $folderName)) {
				$folderName = $path . ' (' . ++$i . ')';
```
I was already confused by this when reviewing. Maybe you should even do the `$i++` before the string concatenation.

tests/Integration/features/mountpoint.feature
```diff
@@ -17,6 +17,24 @@ Feature: mountpoint
     And user "alice" has webdav access to "BehatMountPoint" with permissions "RMG"
     And user "bob" has webdav access to "BehatMountPoint" with permissions "MG"
 
+  Scenario: Delete empty folder when in conflict with mountpoint
+    When user "jane" creates folder "MyCollectives"
+    And user "jane" sees webdav node "MyCollectives"
+    Then user "jane" sets setting "user_folder" to value "/MyCollectives"
+    And user "jane" fails to see webdav node "MyCollectives (1)"
+    Then user "jane" sets setting "user_folder" to value "/Collectives"
+    And user "jane" fails to see webdav node "MyCollectives"
```
provokateurin, line 31:
```suggestion
    And user "jane" fails to see webdav node "MyCollectives"
    And user "jane" sees webdav node "Collectives"
```

tests/Integration/features/mountpoint.feature
```diff
@@ -17,6 +17,24 @@ Feature: mountpoint
     And user "alice" has webdav access to "BehatMountPoint" with permissions "RMG"
     And user "bob" has webdav access to "BehatMountPoint" with permissions "MG"
 
+  Scenario: Delete empty folder when in conflict with mountpoint
+    When user "jane" creates folder "MyCollectives"
+    And user "jane" sees webdav node "MyCollectives"
+    Then user "jane" sets setting "user_folder" to value "/MyCollectives"
```
provokateurin, line 25: Maybe also create a collective here, just to make sure you can see that as well in the correct folder.

tests/Integration/features/mountpoint.feature
```diff
@@ -17,6 +17,24 @@ Feature: mountpoint
...
+    And user "jane" fails to see webdav node "MyCollectives (1)"
+    Then user "jane" sets setting "user_folder" to value "/Collectives"
+    And user "jane" fails to see webdav node "MyCollectives"
+
+  Scenario: Rename non-empty node when in conflict with mountpoint
+    When user "jane" creates folder "MyCollectives"
+    And user "jane" creates folder "MyCollectives/subfolder"
+    And user "jane" sees webdav node "MyCollectives/subfolder"
+    Then user "jane" sets setting "user_folder" to value "/MyCollectives"
```
provokateurin, line 42: Maybe also create a collective here, just to make sure you can see that as well in the correct folder.

lib/Mount/MountProvider.php
mejo-: Well, the second query further down below should only happen if the first one returns a result and that one is a directory (before setting up the mountpoint). So my hope would be that the second one only happens a single time when renaming/deleting a conflicting folder would happen and not in subsequent calls. So my thought was to keep the query that happens every time cheap and simple. What do you think?

provokateurin: I get what you mean, but in the referenced code you're already doing two queries that could be one. It might not matter though, because the getHomeMountForUser call might return data that was already fetched and cached anyway. Maybe this isn't important enough to bother with it.

lib/Mount/MountProvider.php
provokateurin: Or you could make it a do-while-loop, that would be even cleaner :sunglasses:

tests/Integration/features/mountpoint.feature
mejo-: This is already checked in another Scenario ("Change collectives folder for user") further down :wink:

tests/Integration/features/mountpoint.feature
mejo-: Same here.

tests/Integration/features/mountpoint.feature
provokateurin: But with a custom folder after doing the conflict resolution? I think it's a good idea to test this specifically.

mejo-: Ok agreed :blush: Added now.

mejo-:
Thanks for the thorough review @provokateurin, much appreciated. I hope I addressed all your comments, see the pushed commits.

lib/Fs/UserFolderHelper.php
```diff
@@ -48,57 +36,20 @@ public function getUserFolderSetting(string $userId): string {
...
-		try {
-			$userCollectivesFolder = $userRootFolder->get($userCollectivesPath);
-			if (!$userCollectivesFolder instanceof Folder) {
-				throw new NotFoundException('Collectives path exists but is not a folder');
-			}
-		} catch (FilesNotFoundException $e) {
-			throw new NotFoundException('Collectives folder mount point not found', 0, $e);
+		$userCollectivesFolder = $userHomeFolder->get($userCollectivesPath);
+		if (!$userCollectivesFolder instanceof Folder) {
+			throw new FilesNotFoundException('Collectives path exists but is not a folder');
```
provokateurin: Given that you're renaming conflicts now, couldn't you handle this properly now or drop the check since you know that it must be a folder at all times?

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+
+		if ($query->executeQuery()->rowCount() === 0) {
+			// No conflicting entry found
+			return;
+		};
+
+		/** @var IStorage $userHomeStorage */
+		$userHomeStorage = $userHome->getStorage();
+		$userHomeCache = $userHomeStorage->getCache();
+		$userFolderCache = $userHomeCache->get($filesPath);
```
provokateurin:
```suggestion
		$userFolderCacheEntry = $userHomeCache->get($filesPath);
```
This is a bit confusing otherwise.

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+
+		// Check if node is a folder and empty
+		$isEmptyDir = false;
+		if ($userFolderCache !== false && $userFolderCache->getMimeType() === 'httpd/unix-directory') {
+			$query = $this->connection->getQueryBuilder();
+			$query->select('path')
+				->from('filecache')
+				->where($query->expr()->eq('storage', $query->createNamedParameter($userHome->getNumericStorageId(), IQueryBuilder::PARAM_INT)))
+				->andWhere($query->expr()->eq('parent', $query->createNamedParameter($userFolderCache->getId(), IQueryBuilder::PARAM_INT)))
```
provokateurin:
```suggestion
				->where($query->expr()->eq('parent', $query->createNamedParameter($userFolderCache->getId(), IQueryBuilder::PARAM_INT)))
```
This is enough and probably a tiny bit cheaper.

lib/Mount/MountProvider.php
```diff
@@ -142,14 +135,68 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
...
+			$userHomeStorage->rmdir($filesPath);
+			$userHomeCache->remove($filesPath);
+			$userHomeStorage->getPropagator()->propagateChange($filesPath, time());
+		} else {
+			// Rename node
+			$i = 1;
+			$folderName = $path . ' (' . $i . ')';
+			while ($userHomeCache->inCache('files/' . $folderName)) {
+				$folderName = $path . ' (' . ++$i . ')';
+			}
```
provokateurin: No do-while-loop? :smiling_face_with_tear:

lib/Fs/UserFolderHelper.php
mejo-: I added it to make psalm happy. Replaced with a var annotation now :blush:

lib/Mount/MountProvider.php
mejo-: Ok, changed now :wink:

mejo-:
Thanks again @provokateurin, all additional comments addressed :dizzy:

Issue #1773, opened 2025-05-21
Enhanced file management in collectives


marcoambrosini:
## The problem 

Working with files in collective is complicated. Users share files with each other within the collectives app, but there's no way to have an overview of them or an easy way to browse them. For this reason, they usually have to reshare the files in the files app and create a maintain shared folder there. This can lead to mistakes and duplication.

## Possible solutions

- overview of the files inside a collective, maybe in the collective homepage
- Searchable files list
- Include files to the "recently edited" feed

cc @juliusknorr @max-nextcloud @mejo-

mejo-:
@max-nextcloud and me discussed the issue with several Collectives users and drafted an idea how to tackle this problem.

The basic idea is that in future each page in Collectives has three views:
* Text: the default with text editor as it's now
* Subpages: a list of all direct subpages in tiles, as known from the recent pages tiles on the landingpage
* Files: a list of all files in the page folder (and in the attachments folder?)

The default view is configurable for each page. That's particularly useful for pages that serve as parent for subpages but don't have an own content, which is often the case for category/team pages like "Meeting notes" or "Engineering team". For these pages, the subpage view would be more appropriate as default.

Here's dirty mockups of how the views would look like:

| Text view | Subpages view | Files view |
| --- | --- | --- |
|  <img width="2041" height="1430" alt="Image" src="https://github.com/user-attachments/assets/556080df-60c6-41ac-8653-4cde1dab9619" /> | <img width="2041" height="1430" alt="Image" src="https://github.com/user-attachments/assets/52971da4-f11f-4ec3-a9a1-022681aa3142" /> | <img width="2041" height="1429" alt="Image" src="https://github.com/user-attachments/assets/de6ffa47-5739-4961-ac19-9c19c5becd95" /> |

Some details:
* Navigating between views adds entry in browser history, so it's possible to navigate back and forth
  * We can use URL hashs for that (e.g. `tab-text`, `tab-subpages`, `tab-files`)
* Maybe add possibility to add subpages and upload files when page is empty (to advertise the other views in empty pages)
* How to link a file from another page inside a page?
  * Ideally "Link to File or Folder" from editor opens a modal where you can browse non-markdown files in the collective and select a file. And a smart picker action to select non-markdown files from current collective.

mejo-:
[...]

PR #2235, opened 2026-02-04
Enhanced attachments support

mejo-:
### 📝 Summary

This PR enhances the attachment sidebar tab.

* Upload/rename/delete attachments
* Upload attachments via drag & drop into sidebar
* Insert attachments into page content via drag & drop
* List files that live in the folder directory as "Files next to page"
* Contributes to #1773

Requires ~~https://github.com/nextcloud/text/pull/8238 and https://github.com/nextcloud-libraries/nextcloud-vue/pull/8169~~ (released).

#### 🖼️ Screenshots

Attachments sidebar | With drop area (on dragover) | Attachment actions | Rename modal
---|---|---|---
image | image | image | image

Extra files actions | Extra files info box
--- | ---
image | image

### 🚧 TODO

- [x] Highlight attachments that are embedded in content.
- [x] Allow to undo deleting an attachment.
- [x] Behat tests
- [x] Cypress tests

### 🏁 Checklist

### DIFF
--- a/src/components/PageSidebar/AttachmentItem.vue
+++ b/src/components/PageSidebar/AttachmentItem.vue
@@ -13,13 +13,16 @@
 		@dragstart="onDragstart"
 		@click="onClick">
 		<template #icon>
-			<img
-				lazy="true"
-				:src="previewUrl"
-				alt=""
-				height="256"
-				width="256"
-				class="attachment__image">
+			<div class="icon-container">
+				<img
+					lazy="true"
+					:src="previewUrl"
+					alt=""
+					height="256"
+					width="256"
+					class="attachment__image">
+				<DeleteIcon v-if="isDeleted" :size="36" class="attachment__image-overlay" />
+			</div>
 		</template>
 		<template #subname>
 			<div class="attachment__info">
@@ -307,10 +310,29 @@ export default {
 		}
 	}
 
+	.icon-container {
+		position: relative;
+		display: flex;
+
+	}
+
 	&__image {
 		width: 3rem;
 		height: 3rem;
 		border: 1px solid var(--color-border);
+		border-radius: var(--border-radius-large);
+	}
+
+	&__image-overlay {
+		position: absolute;
+		top: 50%;
+		left: 50%;
+		transform: translate(-50%, -50%);
+		color: var(--color-text-maxcontrast-background-blur);
+		background: var(--color-main-background-blur);
+
+		width: 100%;
+		height: 100%;
 		border-radius: var(--border-radius-large);
 	}
 


max-nextcloud:
Thanks a lot for tackling this! Looks great already.

Just based on the screenshots I have a few remarks.

### Drop zone

I'd say always indicate the possibility to drag files in, even though less prominent. Maybe just `[+ Upload] or drag files here`.

### `Files next to page`

If I have a `Protocols` page and a `Today` subpage that is a plain `Today.md` file... would the files in the `Protocols` folder also show in the `Today` subpage? I assume not - but the wording sounds like they do.

How about we rename the tab to `Files` and list all files there without separation. Files that have been attached to the page get an extra button next to the three dot menu that shows them in the page. This button could be a `pin` or an `attachment` symbol.
We could still sort the attached files to the top. But maybe it's even nicer to have a newly uploaded file on the top so it's easy to find.

src/components/PageSidebar/SidebarTabAttachments.vue
```diff
@@ -181,173 +184,203 @@ export default {
...
-				ev.preventDefault()
-				window.OCA.Viewer.open({ path: attachment.path })
+			await this.uploadAttachments(files)
+		},
+
+		async uploadAttachments(files) {
+			for (const file of files) {
+				try {
+					await this.uploadAttachment(file)
+					this.$refs.fileInput.value = ''
```
max-nextcloud: This is not really relevant, right? There's no use of `fileInput.value` anywhere I can see.

Otherwise we might have to think about scenarios where the upload is not triggered by the file input but by drag and drop. Then you could have a bunch of files uploaded which clear the value even while using the `fileInput` to trigger another upload.

mejo-: Good catch @max-nextcloud. I tested and with drag and drop `fileInput.value` is always an empty string. With using the upload button, the value always contains a fake path to the first selected file (e.g. `C:\fakepath\stegosaurus.png`). So I think it's still a good thing to clear the value at the end of the upload, but not within the for loop for sure. Changed this now.

lib/Service/AttachmentService.php
```diff
@@ -191,7 +191,13 @@ public function restoreAttachment(int $collectiveId, File $pageFile, int $attach
 
 		$trashItem = $this->trashBackend->getTrashItemByCollectiveAndId($this->userManager->get($userId), $collectiveId, $attachmentId);
 		if ($trashItem === null) {
-			throw new NotFoundException('Failed to restore attachment ' . $attachmentId . '. Not found in trash.');
+			throw new NotFoundException('Failed to restore attachment ' . $attachmentId . '. Not found in trash for page attachments.');
+		}
+
+		$attachmentsDirectory = $this->getAttachmentDirectory($pageFile);
+		$originalLocation = $trashItem->getOriginalLocation();
+		if ($attachmentsDirectory->getInternalPath() !== dirname($originalLocation)) {
```
max-nextcloud, line 199: This will fail with renames / moves, right?

Maybe we just compare the `attachments.1234` part of the path? This would imply that if I move a folder into a share, people with write access to the share can restore files that I deleted before the move. I'm not sure about that. I'd rather be strict with restore policy than risking data leaks.

The main purpose of allowing recovery through share links from my point of view is to allow restoring files one accidentally deleted. This would already be possible with your version.

Or we check:
* the original path starts with the path to the share
* the original path ends with the `.attachments.1234` folder.

That way moves / renames within the share would not be a problem. Renames / moves above the share (i.e. Collectives folder, Collective name etc)  would still not work. But those are outside of the reach of people working on the share at least.

mejo-: > Maybe we just compare the attachments.1234 part of the path? This would imply that if I move a folder into a share, people with write access to the share can restore files that I deleted before the move. I'm not sure about that. I'd rather be strict with restore policy than risking data leaks.

That's actually a good catch and a strong reason why we should not limit the comparison to the `.attachments.<fileId>` part of the path.

> The main purpose of allowing recovery through share links from my point of view is to allow restoring files one accidentally deleted. This would already be possible with your version.

Currently we only offer restore until the user reloaded or navigated away anyway, so I think the practical impact of the limitation with your scenario is very small.

src/components/PageSidebar/SidebarTabAttachments.vue
```diff
@@ -22,162 +27,159 @@
...
+
+				<input
+					ref="fileInput"
+					type="file"
+					multiple
+					style="display: none"
+					@change="onFilesSelected">
+
+				<!-- drag and drop notice -->
+				<div v-show="!hasAttachments || isDragover" class="upload-drop-area">
```
max-nextcloud: This will always show if there are no attachments yet. I don't think that's what you want according to the conversation so far.

Image

mejo-: Indeed, good catch. Will fix tomorrow.

mejo-: Fixed now:

image

src/components/PageSidebar/AttachmentItem.vue
```diff
@@ -0,0 +1,370 @@
...
+		border-radius: var(--border-radius-large);
+	}
+
+	&__image-overlay {
+		position: absolute;
+		top: 50%;
+		left: 50%;
+		transform: translate(-50%, -50%);
+		color: var(--color-text-maxcontrast);
+		background: rgba(var(--color-main-background-rgb), 0.5);
```
max-nextcloud: I think the contrast is too low here for dark images on a light mode and vice versa:

Image

I'm afraid you'll have to use:

```suggestion
		color: var(--color-text-maxcontrast-background-blur);
		background: var(--color-main-background-blur);
```

mejo-: Agreed, good idea. Changed, tested and pushed :relaxed:

max-nextcloud: Which looks like this:
grafik

mejo-: I think that's good enough for now :wink:

max-nextcloud: Not quite... Sorry.. :stuck_out_tongue_winking_eye: 

it's
```
		color: var(--color-text-maxcontrast-background-blur);
```

With the `background-` in there. Messed that up at first, then corrected my proposal - but you were faster.

PR #2236, opened 2026-02-05
fix(controllers): Add User Trait to handle null sessions

solracsf:
### 📝 Summary

> TypeError: OCA\Collectives\Controller\PageController::__construct(): Argument #9 ($userId) must be of type string, null given

This Pull Request addresses a critical `TypeError` occurring when the user session is missing or expired. By making the `$userId` dependency nullable and implementing a centralized session validation, we ensure the application remains stable during anonymous or unauthenticated OCS API calls.

Previously, the constructor strictly required a string for the `$userId` argument. This caused a fatal error when Nextcloud's dependency injection container passed null (e.g., for expired sessions or guest access).

- Updated `$userId` type hint to `?string` in the constructor to allow safe instantiation.
- Throws an `OCSForbiddenException` if the user is not authenticated, providing a clean API response instead of a crash.

### 🏁 Checklist

### DIFF
--- a/lib/Controller/CollectiveController.php
+++ b/lib/Controller/CollectiveController.php
@@ -35,6 +35,7 @@ use Psr\Log\LoggerInterface;
  */
 class CollectiveController extends OCSController {
 	use OCSExceptionHelper;
+	use UserTrait;
 
 	public function __construct(
 		string $AppName,
@@ -44,7 +45,7 @@ class CollectiveController extends OCSController {
 		private IFactory $l10nFactory,
 		private LoggerInterface $logger,
 		private NodeHelper $nodeHelper,
-		private string $userId,
+		private ?string $userId,
 	) {
 		parent::__construct($AppName, $request);
 	}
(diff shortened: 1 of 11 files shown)


mejo- (changes requested):
Thanks for the contribution @solracsf! The code change itself looks good to me, but I'd be curious what exactly you want to fix here. It's not about broken functionality, but about fixing the TypeErrors being logged, right?

In any case, the fix would need to be applied to all other non-public controllers as well. Could you take care of this?

lib/AppInfo/Application.php
```diff
@@ -84,6 +85,7 @@ public function register(IRegistrationContext $context): void {
 		$context->registerEventListener(MentionEvent::class, TextMentionListener::class);
 
 		$context->registerMiddleware(PublicOCSMiddleware::class);
+		$context->registerMiddleware(SessionMiddleware::class);
```
mejo-: Is the middleware even required? If I understand the code correctly, the UserTrait already takes care of checking whether the userID is null, so it seem to me like the SessionMiddleware duplicates this check. Or am I missing something here?

solracsf: Sorry, this was a leftover of my 1st approach before moving to the Trait...

PR #2256, opened 2026-02-12
fix(navigation): delay closing only when switching collectives

mejo-:
We don't want delayed closing when opening a collective in a new tab or in public collectives.

### 🏁 Checklist

### DIFF
--- a/src/components/CollectiveContainer.vue
+++ b/src/components/CollectiveContainer.vue
@@ -113,7 +113,7 @@ export default {
 		'currentCollective.id': function(val) {
 			this.clearFilterTags()
 			if (val) {
-				this.initCollective()
+				this.initCollective({ closeNavDelay: true })
 			}
 		},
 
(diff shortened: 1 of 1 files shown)


src/components/CollectiveContainer.vue
```diff
@@ -146,8 +146,13 @@ export default {
 		...mapActions(useTagsStore, ['clearFilterTags']),
 		...mapActions(useVersionsStore, ['selectVersion']),
 
-		async initCollective() {
-			setTimeout(() => this.closeNav(), 1000)
+		async initCollective({ closeNavDelay = false }) {
+			if (closeNavDelay && !this.isPublic) {
+				setTimeout(() => this.closeNav(), 1000)
+			} else {
+				console.debug('closeNav')
```
blizzz: is it supposed to stay?

mejo-: Nope, good catch!

PR #2268, opened 2026-02-13
chore(test): introduce playwright tests

mejo-:
### 🏁 Checklist


playwright/support/fixtures/Collective.ts
max-nextcloud, line 1: I suggest to define a type for all the data inside the collective and then construct it with `new Collective({data, page})`. That way you'd only define the structure once.

In the function you can then deconstruct the data for convenience:
```js
const { slug, name, id } = this.data
...
```

mejo-: Good point, thank you. Adjusted now.

PR #2274, opened 2026-02-17
fix(CollectiveStore): remove cache instance variable

mejo-:
It's already defined in OC\Files\Storage\Wrapper\Wrapper which our CollectiveStorage class extends. Wrapper got strong typing added in server master branch recently, which broke Collectives due to incompatible typing.

### 🏁 Checklist

### DIFF
--- a/src/components/Nav/NewCollectiveModal.vue
+++ b/src/components/Nav/NewCollectiveModal.vue
@@ -42,6 +42,7 @@
 				<NcButton
 					v-if="anyCircle && !pickCircle"
 					:title="t('collectives', 'Select an existing team')"
+					:aria-label="t('collectives', 'Select an existing team')"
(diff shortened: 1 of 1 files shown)


src/components/Nav/NewCollectiveModal.vue
```diff
@@ -42,6 +42,7 @@
 				<NcButton
 					v-if="anyCircle && !pickCircle"
 					:title="t('collectives', 'Select an existing team')"
+					:aria-label="t('collectives', 'Select an existing team')"
```
CarlSchwan, line 45: Unrelated change?

mejo-: Hehe, unrelated but still good to get this in as well :zany_face:

Issue #1742, opened 2025-05-06
Missing emoji icons after copying subpages to new Collective


viktorix:
**Describe the bug**
When you copy pages with subpages to a new Collective, the subpages lose their emoji icon.

**To Reproduce**
1. Go to 'Collectives'
2. Find a page with subpages that have emoji icons.
3. Copy that page to a new Collective.
4. Check the copied subpages and their emoji icons.

**Expected behavior**
All emoji icons for subpages are copied too.

**Screenshots**

![Image](https://github.com/user-attachments/assets/db62ad72-b0ca-4c79-9e19-98cb15f910bb)

**Server details:**
c.nc.c

PR #2275, opened 2026-02-17
fix(PageService): copy metadata of subpages when copying an index page

mejo-:
Fixes: #1742 

### 🏁 Checklist

### DIFF
--- a/tests/Integration/features/bootstrap/FeatureContext.php
+++ b/tests/Integration/features/bootstrap/FeatureContext.php
@@ -643,6 +643,20 @@ class FeatureContext implements Context {
 	}
 
 	/**
+	 * @When user :user sees emoji :emoji for page :page in :collective
+	 *
+	 * @throws GuzzleException
+	 */
+	public function userSeesPageEmoji(string $user, string $page, string $emoji, string $collective): void {
+		$this->setCurrentUser($user);
+		$collectiveId = $this->collectiveIdByName($collective);
+		$pageId = $this->pageIdByName($collectiveId, $page);
+		$this->sendOcsCollectivesRequest('GET', 'collectives/' . $collectiveId . '/pages/' . $pageId);
+		$this->assertStatusCode(200);
+		$this->assertPageKeyValue($pageId, 'emoji', $emoji);
+	}
+
+	/**
 	 * @When user :user sets emoji for template :template to :emoji in :collective
 	 * @When user :user :fails to set emoji for template :template to :emoji in :collective
 	 *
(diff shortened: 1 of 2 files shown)


lib/Service/PageService.php
```diff
@@ -737,6 +737,43 @@ private function isAncestorOf(Folder $collectiveFolder, int $pageId, int $target
...
+			} elseif ($sourceNode instanceof File && NodeHelper::isPage($sourceNode)) {
+				try {
+					$targetNode = $targetFolder->get($sourceNode->getName());
+					if ($targetNode instanceof File && NodeHelper::isPage($targetNode)) {
+						$sourcePageInfo = $this->getPageByFile($sourceNode);
+						if ($sameCollective) {
+							$this->updatePage($collectiveId, $targetNode->getId(), $userId, $sourcePageInfo->getEmoji(), $sourcePageInfo->isFullWidth(), $sourcePageInfo->getSlug(), $sourcePageInfo->getTags());
+						} else {
+							$this->updatePage($collectiveId, $targetNode->getId(), $userId, $sourcePageInfo->getEmoji(), $sourcePageInfo->isFullWidth(), $sourcePageInfo->getSlug(), '[]');
+						}
```
max-nextcloud: I think it's a little easier to understand like this:
```suggestion
						$tags = $sameCollective ? $sourcePageInfo->getTags() : '[]';
						$this->updatePage($collectiveId, $targetNode->getId(), $userId, $sourcePageInfo->getEmoji(), $sourcePageInfo->isFullWidth(), $sourcePageInfo->getSlug(), $tags);
```

lib/Service/PageService.php
```diff
@@ -737,6 +737,43 @@ private function isAncestorOf(Folder $collectiveFolder, int $pageId, int $target
 		return $this->isAncestorOf($collectiveFolder, $pageId, $targetParentPageId);
 	}
 
+	/**
+	 * @throws NotFoundException
+	 */
+	private function copySubpagesMetadata(int $collectiveId, Folder $sourceFolder, Folder $targetFolder, bool $sameCollective, string $userId): void {
```
max-nextcloud: Maybe rename `$sameCollective` to `$copyTags` as that's what it really means in the context of this function.

lib/Service/PageService.php
mejo-: Good idea, changed.

lib/Service/PageService.php
mejo-: same.

Issue #1399, opened 2024-08-01
Migrate to Vue3


mejo-:
**Describe the bug**
At one point we have to migrate the app to Vue3.

Preliminaries known so far:
* [ ] Migrate from vuex to pinia (#1398)
* [ ] Migrate from webpack to vite
* [ ] Find a vue3 wrapper lib for sortable.js. (https://github.com/Alfred-Skyblue/vue-draggable-plus, https://github.com/MaxLeiter/sortablejs-vue3)

PR #2279, opened 2026-02-18
feat: migrate to vue3

mejo-:
### 📝 Summary

Resolves: #1399 

### 🚧 TODO

- [x] Subpages don't get rendered.
- [x] PageInfoBar component is not rendered in reader, probably because it's a Vue3 component now.
- [x] Files app header is not displayed.
- [x] Test drag and drop in page list intensively as we migrated to a new wrapper library. ~~`revertOnSpill` is not supported by it~~
- [x] Test opening anchor links with headings. vue-router uses history for hash links in v4/v5, no idea if this is a problem.

### 🏁 Checklist

### DIFF
--- a/cypress/e2e/collective-settings.spec.js
+++ b/cypress/e2e/collective-settings.spec.js
@@ -39,7 +39,7 @@ describe('Collective settings', function() {
 	it('Allows to rename the collective', function() {
 		cy.openCollectiveMenu('Change me')
 		cy.clickMenuButton('Settings')
-		cy.get('div.collective-name input[type="text"]').type(' now{enter}')
+		cy.get('div.collective-name-and-emoji input[type="text"]').type(' now{enter}')
 		cy.reload()
 		cy.get('.collectives_list_item')
 			.should('contain', 'Change me now')
@@ -48,7 +48,7 @@ describe('Collective settings', function() {
 	it('Allows to change editing permissions', function() {
 		cy.openCollectiveMenu('Change me')
 		cy.clickMenuButton('Settings')
-		cy.get('.permissions-input-edit').contains('Admins only')
+		cy.get('.edit-permissions').contains('Admins only')
 			.click()
 		cy.get('div.toast-success').should('contain', 'Editing permissions updated')
 	})
@@ -56,7 +56,7 @@ describe('Collective settings', function() {
 	it('Allows to change sharing permissions', function() {
 		cy.openCollectiveMenu('Change me')
 		cy.clickMenuButton('Settings')
-		cy.get('.permissions-input-share').contains('Admins only')
+		cy.get('.share-permissions').contains('Admins only')
 			.click()
 		cy.get('div.toast-success').should('contain', 'Sharing permissions updated')
 	})
@@ -64,10 +64,7 @@ describe('Collective settings', function() {
 	it('Allows to change page mode', function() {
 		cy.openCollectiveMenu('Change me')
 		cy.clickMenuButton('Settings')
-		cy.get('a.navigation-list__link')
-			.contains('Page settings')
-			.click()
-		cy.get('.edit-mode').contains('Edit')
+		cy.get('.page-mode').contains('Edit')
 			.click()
 		cy.get('div.toast-success').should('contain', 'Default page mode updated')
 	})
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -13,5 +13,8 @@
 		"moduleResolution": "Bundler",
 		"lib": ["webworker"],
 		"allowImportingTsExtensions": true
-	}
+	},
+	"vueCompilerOptions": {
+		"target": 3.5
+	},
```
ShGKme: This is not needed. It was only necessary for 2.7.

src/components/Member/MemberItem.vue
```diff
@@ -17,11 +17,11 @@
...
-			:is-no-user="isNoUser"
-			:icon-class="iconClass"
-			:disable-menu="true"
-			:disable-tooltip="true"
-			:hide-status="true"
+			:isNoUser="isNoUser"
+			:iconClass="iconClass"
+			:disableMenu="true"
+			:disableTooltip="true"
+			:hideStatus="true"
```
ShGKme: **note:**
- Boolean props can be passed without value, like in HTML (also in Vue 2)
- In Vue 3.5 there is a new shortcut like in JS objects when key and value are the same
```suggestion
			:isNoUser
			:iconClass
			disableMenu
			disableTooltip
			hideStatus
```

src/components/Member/MemberSearchResults.vue
```diff
@@ -119,19 +119,19 @@ export default {
 		},
 
 		circleUserType() {
-			return function(source) {
+			return (source) => {
 				return circlesMemberTypes[autocompleteSourcesToCircleMemberTypes[source]]
 			}
 		},
```
ShGKme, line 125: Not a part of the question. But this this is a `computed` and not a `method`?
There is no computing here, it is a static function. And the same in many places.

src/components/CollectiveContainer.vue
```diff
@@ -187,8 +187,12 @@ export default {
...
 				this.$router.replace({ path: this.currentCollectivePath, hash: document.location.hash })
 			} else if (this.currentPage
-				&& this.$route.path !== this.currentPagePath) {
-				this.$router.replace({ path: this.currentPagePath, hash: document.location.hash })
+				&& this.$route.fullPath !== this.currentPagePath) {
+				this.$router.replace({
+					path: this.currentPagePath,
+					query: this.fileIdQuery ? { fileId: this.fileIdQuery } : {},
+					hash: document.location.hash,
+				})
```
ShGKme, line 194: It is different from the previous replace. Is it intentional?

src/components/Page/PageTitle.vue
```diff
@@ -81,6 +88,12 @@ export default {
 				event.preventDefault()
 			}
 		},
+
+		onBlur() {
+			if (this.$.isMounted) {
+				this.$emit('blur')
+			}
+		},
```
ShGKme, line 96: This looks like a Private API. Why was it needed?

ShGKme (approved):
Looks good on a brief check in the Collectives.
I had 500 on `ocs/v2.php/apps/collectives/api/v1.0/collectives/3/pages/templates` on opening a newly created page in a new collective.

One small issue:
In `@nextcloud/vue` v9 you need `legacy` prop on `<NcAppSettingsDialog>` to have old paddings with old settings design.

Or migrate to the new design :D 

Image

src/components/Member/MemberItem.vue
mejo-: Good point, adjusted now.

src/components/Member/MemberSearchResults.vue
mejo-: Yeah, that's something we do a lot and we need to address this separately. We have a lot of computed properties that in fact return a static function and do any caching. I think that's one area where we can improve the frontend performance a lot. But out of scope for this PR I'd say.

src/components/CollectiveContainer.vue
mejo-: Well, apparently before (with vue-router 3) it was possible to pass the query as part of `path`, which no longer works. That's why I added the `query` parameter explicitly now.

src/components/Page/PageTitle.vue
mejo-: There is a race condition when renaming the page by pressing <Enter>. The page gets re-routed to the new page URL via vue-router which triggers the `blur` event which in turn tries to rename again. That was my best bet to prevent the race condition. Do you have an idea how to otherwise fix the race condition?

PR #2280, opened 2026-02-18
upgrade: cypress to v15, drop NC 31

max-nextcloud:
- Upgrade cypress to latest with fast visiblity.
- Include url in visit options to match cypress API.
- Add clipboard commands to cypress.
- drop support for Nextcloud 31.

### DIFF
--- a/tests/Integration/features/bootstrap/FeatureContext.php
+++ b/tests/Integration/features/bootstrap/FeatureContext.php
@@ -1943,11 +1943,6 @@ class FeatureContext implements Context {
 		// simplexml_load_string() would be better than preg_replace
 		$folderPermissions = preg_replace('/.*<oc:permissions>(.*)<\/oc:permissions>.*/sm', '\1', $this->response->getBody()->getContents());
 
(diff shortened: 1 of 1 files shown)


tests/Integration/features/bootstrap/FeatureContext.php
```diff
@@ -1943,11 +1943,6 @@ public function hasCollectivePublicWebdavAccess(string $collective, string $user
 		// simplexml_load_string() would be better than preg_replace
 		$folderPermissions = preg_replace('/.*<oc:permissions>(.*)<\/oc:permissions>.*/sm', '\1', $this->response->getBody()->getContents());
 
-		// TODO: Remove after we only support Nextcloud 33
-		if ((int)$this->getNextcloudVersion(true) <= 32) {
-			$permissions = str_replace('RG', 'RMG', $permissions);
-		}
-
```
max-nextcloud, line 1950: This was introduced due to permission changes in 33 ( 4155461d0 ). I suspect they were also backported to 32 afterwards. At least tests were failing with this special handling in 32.

PR #2294, opened 2026-02-24
fix: scope trash node lookup to requested collective

benjaminfrueh:
When requesting a file from trash, ensure the resolved node belongs to the requested collective's trash folder.

### DIFF
--- a/lib/Trash/PageTrashBackend.php
+++ b/lib/Trash/PageTrashBackend.php
@@ -480,6 +480,12 @@ class PageTrashBackend implements ITrashBackend {
 			// Build the TrashItem object
 			$trashFolder = $this->getTrashFolder($collectiveId);
 			$trashNode = $this->getTrashNodeById($user, $fileId);
+
+			// Ensure the resolved node belongs to the requested trash
+			if ($trashNode === null || !str_starts_with($trashNode->getPath(), $trashFolder->getPath() . '/')) {
+				return null;
+			}
+
 			// Get parent folder for index pages
 			if ($trashNode instanceof File && NodeHelper::isIndexPage($trashNode)) {
 				// The extra `get()` is required to resolve the lazy folder from getParent()


lib/Trash/PageTrashBackend.php
mejo-, line 446:
```suggestion
	public function getTrashNodeById(IUser $user, int $fileId, ?Folder $trashFolder = null): ?Node {
```

How about adding an optional parameter `$trashFolder` to `getTrashNodeById()` and passing the trash folder for the collective in `getTrashItemByCollectiveAndId()` instead?

mejo-, line 449:
```suggestion
			$trashFolder = $trashFolder ?? $this->getAppFolder()->get('trash');
```

lib/Trash/PageTrashBackend.php
```diff
@@ -480,6 +480,12 @@ public function getTrashItemByCollectiveAndId(IUser $user, int $collectiveId, in
 			// Build the TrashItem object
 			$trashFolder = $this->getTrashFolder($collectiveId);
 			$trashNode = $this->getTrashNodeById($user, $fileId);
```
mejo-, line 482:
```suggestion
			$trashNode = $this->getTrashNodeById($user, $fileId, $trashFolder);
```

lib/Trash/PageTrashBackend.php
```diff
@@ -480,6 +480,12 @@ public function getTrashItemByCollectiveAndId(IUser $user, int $collectiveId, in
...
+
+			// Ensure the resolved node belongs to the requested trash
+			if ($trashNode === null || !str_starts_with($trashNode->getPath(), $trashFolder->getPath() . '/')) {
```
mejo-, line 485:
```suggestion
			if ($trashNode === null) {
```

lib/Trash/PageTrashBackend.php
benjaminfrueh: @mejo- thanks for the feedback, I thought about this initially too, but I didn't want to change the method signature here as `getTrashNodeById()` is implemented from the interface `ITrashBackend`

But technically we could still add the optional parameter to `getTrashNodeById()`, do you think it would be okay to do so?

mejo-: Given that PHP allows extra option parameters to an interface method, I'd be in favor of doing this. What do you think @benjaminfrueh?

benjaminfrueh: Sure, we can do that, I think it is a reasonable approach in favour to make the check scoped in case `getTrashNodeById` is called in the future without going through `getTrashItemByCollectiveAndId`.

I always try to stick with interfaces if possible,  but it seems better than a string comparison. Contra is that this is invisible at the interface level but it is optional and only called from within `PageTrashBackend` so that is very minor.

So let's make this change, I will test and push a commit with the suggestions! Thanks again @mejo-

PR #2304, opened 2026-02-28
feat: Bring back team overview

pymnh:
### 📝 Summary

Initially I wanted to open a ticket to remove the Teams overview from the collectives landing page and move it into the members modal, because I too think that it causes too much confusion there and belongs in the members modal. My wish came partially true in 8c9e829f165ff5ae03e0c5c324d20ff0ab4201be but now there is no direct link from the collective to its corresponding team overview, which I personally found to be useful.

This PR brings back the link to the team overview that was removed in 8c9e829f165ff5ae03e0c5c324d20ff0ab4201be , but places it in the members modal next to the name of the collective. This is more intuitive then having it on the landing page: I am already in the members overview and am offered to go "deeper" into the team configuration.  The link icon is generally understood as leading to a new page/modal. 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image | image

#### Notes

The usage of the OpenInNewIcon follows the design of [the Teams dashboard widget](https://github.com/nextcloud/circles/blob/master/src/views/DashboardTeamsWidget.vue):

image

### 🏁 Checklist

- [x] Documentation ([README](https://github.com/nextcloud/collectives/blob/main/README.md) or [documentation](https://github.com/nextcloud/collectives/blob/main/docs/)) <del>has been updated or</del> is not required

### DIFF
--- a/src/components/Nav/CollectiveMembersModal.vue
+++ b/src/components/Nav/CollectiveMembersModal.vue
@@ -10,18 +10,15 @@
 		@close="onClose">
 		<div class="modal-collective-members">
 			<h2 class="modal-collective-members__name">
+				{{ t('collectives', 'Members of collective {name}', { name: collective.name }, { escape: false }) }}
 				<a
 					v-if="showTeamLink"
 					class="team-link"
 					:href="teamUrl"
 					:title="t('collectives', 'Go to team overview')"
 					target="_blank">
-					{{ t('collectives', 'Members of collective {name}', { name: collective.name }, { escape: false }) }}
 					<OpenInNewIcon :size="20" />
 				</a>
-				<template v-else>
-					{{ t('collectives', 'Members of collective {name}', { name: collective.name }, { escape: false }) }}
-				</template>
 			</h2>
 			<MemberPicker
 				:show-current="true"
@@ -131,6 +128,7 @@ export default {
 	height: 550px;
 	max-height: 80vh;
 
+	// Rules are mostly copied from NcDialog:
 	&__name {
 		font-size: 21px;
 		text-align: center;
@@ -138,7 +136,7 @@ export default {
 		min-height: var(--default-clickable-area);
 		overflow-wrap: break-word;
 		margin-block: 0 12px;
-		margin-inline: 0;
+		padding-inline: var(--default-clickable-area);
 
 		&:hover .team-link,
 		&:hover .material-design-icon {
@@ -150,13 +148,11 @@ export default {
 
 .team-link {
 	color: var(--color-main-text);
-	text-decoration: none;
 
 	.material-design-icon {
 		display: inline;
 		vertical-align: middle;
 		color: var(--color-text-maxcontrast);
-		white-space: nowrap;
 	}
 
 	&:focus-visible {


src/components/Nav/CollectiveMembersModal.vue
```diff
@@ -82,9 +107,35 @@ export default {
...
+}
+
 .modal-collective-members {
 	height: 550px;
 	max-height: 80vh;
 	padding-bottom: 12px;
 }
+
+.team-link {
+	position: absolute;
```
mejo-: Unfortunately this doesn't work well. The button is hidden below the "close dialog" button:

Image

I think a better approach would be to migrate `CollectiveMembersModal` back from [NcDialog](https://stable8--nextcloud-vue-components.netlify.app/#/Components/NcDialog) to [NcModal](https://stable8--nextcloud-vue-components.netlify.app/#/Components/NcModal) which allows to customize all the contents of the modal, including its header.

See [the code of NcDialog](https://github.com/nextcloud-libraries/nextcloud-vue/blob/stable8/src/components/NcDialog/NcDialog.vue#L214) for a suggestion how to rebuild the component using NcModal but keeping it similar to how it looks now with NcDialog (except for the added button in the header).

pymnh: Thanks for the feedback. I will look into that :)

pymnh: I migrated the the CollectiveMembersModal to NcModal as you suggested. 

I also used the freedom that the migration gives to make the whole title a hyperlink to the team overview, instead of only the icon, but I can revert that if you think that is too opinionated (Making only the name a link such as `Members of collective<a>{name}[^]</a>` does not seem possible as it would break the positioning due to different translations).

Full disclosure: The css is mostly AI generated with a bit of manual adjustment and some sanity-checking if it matches with the NcDialog layout.

Also seems to work on mobile:

image

image

src/components/Nav/CollectiveMembersModal.vue
```diff
@@ -82,9 +113,55 @@ export default {
...
-	padding-bottom: 12px;
+
+	&__name {
+		font-size: 21px;
+		text-align: center;
+		line-height: var(--default-clickable-area);
+		min-height: var(--default-clickable-area);
+		overflow-wrap: break-word;
+		margin-block: 0 12px;
+		margin-inline: 0;
```
mejo-: This seems superfluous and also doesn't exist in NcDialog sources. AI artifact? 😉

src/components/Nav/CollectiveMembersModal.vue
```diff
@@ -82,9 +113,55 @@ export default {
...
+	}
+}
+
 .modal-collective-members {
 	height: 550px;
 	max-height: 80vh;
-	padding-bottom: 12px;
+
+	&__name {
+		font-size: 21px;
```
mejo-, line 133: A short comment that these rules are copied over from NcDialog would be good I think.

src/components/Nav/CollectiveMembersModal.vue
```diff
@@ -82,9 +113,55 @@ export default {
...
+
+.team-link {
+	color: var(--color-main-text);
+	text-decoration: none;
+
+	.material-design-icon {
+		display: inline;
+		vertical-align: middle;
+		color: var(--color-text-maxcontrast);
+		white-space: nowrap;
```
mejo-: I don't think setting `white-space` has an impact on an icon-only element.

src/components/Nav/CollectiveMembersModal.vue
```diff
@@ -82,9 +113,55 @@ export default {
...
+		&:hover .material-design-icon {
+			color: var(--color-primary-element);
+			transition: color var(--animation-quick) ease;
+		}
+	}
+}
+
+.team-link {
+	color: var(--color-main-text);
+	text-decoration: none;
```
mejo-: `text-decoration` seems superfluous as there's already none without setting it.

src/components/Nav/CollectiveMembersModal.vue
```diff
@@ -4,36 +4,54 @@
...
-	<NcDialog
-		:name="t('collectives', 'Members of collective {name}', { name: collective.name }, { escape: false })"
+	<NcModal
 		size="normal"
-		@closing="onClose">
+		class="collective-members-modal"
+		@close="onClose">
 		<div class="modal-collective-members">
+			<h2 class="modal-collective-members__name">
+				<a
```
mejo-, line 14: To be honest I don't think it's a good idea to turn the whole heading into a link. People will hover over it accidentally and don't get where this link leads. I'd prefer to just have the button next to it being a link.

src/components/Nav/CollectiveMembersModal.vue
pymnh: :see_no_evil: removed and replaced with `padding-inline` so that title/icon and close button do not overlap

Issue #2169, opened 2025-12-20
Last edited information should not be clickable in link shares


y3n4:
**Describe the bug**

Last edited information is a clickable link even if I am not allowed to see the history (i.e. when vistiting a link share of a page or collective)

**To Reproduce**
1. Create a share link for a collectives page or whole collective
2. Open the link in another session
3. Hover mouse over the last edited information "<username> days ago" under the page header 
4. Click on it
5. Sidebar opens, but only "Attachments" and "Backlinks" tabs are available

**Expected behavior**

If I am not allowed to see the history in the sidebar of the page, the last edited information should not be a clickable link to the sidebar.

**Server details:**
- Collectives app version: 3.4.0
- Nextcloud version: 31.0.12

**Client details:**
- Firefox

PR #2305, opened 2026-02-28
fix(PageInfoBar): Do not show versionsLink for public shares

pymnh:
### 📝 Summary

* Completes fix for #2169  

- Last edited information should not be a link if I can not open the versions sidebar
- Beside checking for canEdit, should also check if !isPublic (public shares have no versions sidebar)

### 🏁 Checklist

- [x] Tests (unit, integration and/or end-to-end) passing <del>and the changes are covered with tests</del>
- [x] Documentation ([README](https://github.com/nextcloud/collectives/blob/main/README.md) or [documentation](https://github.com/nextcloud/collectives/blob/main/docs/)) <del>has been updated or</del> is not required

### DIFF
--- a/src/components/Page/PageInfoBar.vue
+++ b/src/components/Page/PageInfoBar.vue
@@ -6,11 +6,11 @@
 <template>
 	<div class="text-menubar">
 		<component
-			:is="canEdit ? 'a' : 'div'"
+			:is="versionsLink ? 'a' : 'div'"
 			v-if="currentPage.lastUserId"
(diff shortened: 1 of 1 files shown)


src/components/Page/PageInfoBar.vue
```diff
@@ -85,6 +87,11 @@ export default {
 			type: Number,
 			required: true,
 		},
+
+		isPublic: {
```
mejo-: This seems superfluous to me.

PR #2312, opened 2026-03-03
Perf: Optimize trashed pages

Koc:
### 📝 Summary
In our organization we are actively using Collectives. And currently we have 2 performance issues with trashed pages:
- queries in the loop (each trashed page loaded using separate query) -> load all pages using single `WHERE IN` query
- trashed pages loaded immediately on each Collective opening -> delay loading until popup with trashed pages opened

This PR aims to fix both of this issues

## :framed_picture: Screenshots
image image

## :construction: TODO
- I can implement lazy loading for deleted collectives as well, just confirm that you ok with that
image

### 🏁 Checklist

### DIFF
--- a/src/stores/pages.js
+++ b/src/stores/pages.js
@@ -914,7 +914,9 @@ export const usePagesStore = defineStore('pages', {
 			const response = await api.trashPage(this.context, pageId)
 			const trashPage = response.data.ocs.data.page
 			removeFrom(this.allPages[this.collectiveIndex], trashPage)
-			updateOrAddTo(this.allTrashPages[this.collectiveIndex], trashPage)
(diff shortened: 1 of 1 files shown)


mejo-:
> Imagine that we have few collecties and each of them has their own trashed pages. We load trashes pages for collective 1, set flag as true, then open collective 2, try to open trashed pages popup but flag is already set to true

That's why the flag is reset each time the collective is switched. And as the flag is not persisted in browser storage, its scope is limited to the current browser tab. The idea was to make sure that each time a collective gets opened (either by browsing there or by switching the collective from within the app), opening the page trash for the first time will fetch the trashed pages.

PR #2229, opened 2026-02-03
debug/circle initiator

juliusknorr:
- **debug: Log when getting null as initiator**
- **fix: Catch generic exceptions during mountpoint setup**

### 📝 Summary

* Resolves: # 

May help to debug further cases like https://github.com/nextcloud/circles/issues/1284 and avoid them breaking the whole mountpoint setup

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

- [ ] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [ ] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [ ] Tests (unit, integration and/or end-to-end) passing and the changes are covered with tests
- [ ] Documentation ([README](https://github.com/nextcloud/collectives/blob/main/README.md) or [documentation](https://github.com/nextcloud/collectives/blob/main/docs/)) has been updated or is not required

PR #2314, opened 2026-03-05
fix: Catch exceptions at Circle::getInitiator during mountpoint setup

mejo-:
Extracted from https://github.com/nextcloud/collectives/pull/2229 and adapted

### 🏁 Checklist

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -12,6 +12,7 @@ namespace OCA\Collectives\Mount;
 use OC\Files\Cache\Cache;
 use OCA\Collectives\Db\Collective;
 use OCA\Collectives\Fs\UserFolderHelper;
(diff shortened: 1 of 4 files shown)


mejo-:
Thanks for your thoughts @max-nextcloud

> Assuming that's your goal I agree with it. 

Yes, you nailed it :laughing:

> I'm not a fan of simply logging unexpected errors - as they may remain unnoticed for quite a while.

Indeed, I agree.

> Instead I'd prefer to keep the scope of generic error handling as small as possible. I.e. catching generic exceptions thrown by circles, rethrowing them as our own Exception - for example CircleException and only adding that to the exceptions caught in the MountProvider.

This sounds like a good way forward to me :+1:

mejo-:
@max-nextcloud I refactored the change now to only catch generic exceptions at `Circle::getInitiator()` for now as this was the original problem.

PR #2343, opened 2026-03-15
feat(PageReference): Preview "Page not found" if page is not accessible

pymnh:
#### Change

* Let the backend return empty Reference with `accessible = false` if current user cannot resolve
* In the refernce widget: If `accessible = false`, show an informative "Page not found" preview and explain that the page does not seem to exist or that the permissions are missing to view it.
* <del>On click: open a popover to confirm if user really wants to follow the link</del>

### Reasoning:

* Currently if I have not the permission to view a collectives page, the ReferenceProvider just falls back to opengraph, even if the URL matches a collective page.
* This is IMO not correct: The collectives preview widget should still be responsible, even if it does not preview the content
* One consideration: The way I did it, neither the name of the collective nor the title of the page are shown to the user. Technically these are not secret information, as they are included in the link. IMO one could also just render the title and change the description to include that the page was not found. I have no strong opinion on this.
 

##### 🏚️ Before

image

##### 🏡 After

image

### 🚧 TODO

- [ ] A fallback preview might also make sense for other reference providers and there should be a common way of doing this, so it would be nice to gather consensus if...
  - The `accessible` property is the right place to indicate that a "resource not found" message should be displayed
  - If the Reference provider should return a link or a completely empty richObject or can sometimes carry some more information? (-> Decks have only numeric ids in their URL, while collectives include Collective name and Page name in the URL, so for collectives, passing on these infos might not be a privacy issue?)
- [ ] Add some unit tests 

### 🏁 Checklist

### DIFF
--- a/src/views/PageReferenceWidget.vue
+++ b/src/views/PageReferenceWidget.vue
@@ -9,14 +9,13 @@
 		v-if="richObject && notFound"
 		:href="richObject.link"
 		target="_blank"
-		class="collective-page not-found"
-		@click="clickLink">
+		class="collective-page not-found">
 		<div class="collective-page--image">
 			<PageIcon
 				:size="50" />
 		</div>
 		<div class="collective-page--info">
-			<div>
+			<div class="title">
 				<strong>
 					{{ t('collectives', 'Page not found') }}
 				</strong>
@@ -44,7 +43,7 @@
 				:size="50" />
 		</div>
 		<div class="collective-page--info">
-			<div>
+			<div class="title">
 				<strong>
 					{{ richObject.page.title }}
 				</strong>
@@ -65,10 +64,8 @@
 <script>
 import { t } from '@nextcloud/l10n'
 import { generateUrl } from '@nextcloud/router'
-import { mapState } from 'pinia'
 import NcUserBubble from '@nextcloud/vue/components/NcUserBubble'
 import PageIcon from '../components/Icon/PageIcon.vue'
-import { useRootStore } from '../stores/root.js'
 
 export default {
 	name: 'PageReferenceWidget',
@@ -98,8 +95,6 @@ export default {
 	},
 
 	computed: {
-		...mapState(useRootStore, ['isPublic']),
-
 		emoji() {
 			return this.richObject.page?.emoji
 		},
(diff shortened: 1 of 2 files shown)


src/views/PageReferenceWidget.vue
```diff
@@ -4,8 +4,31 @@
...
+				:size="50" />
+		</div>
+		<div class="collective-page--info">
+			<div>
+				<strong>
+					{{ t('collectives', 'Cannot preview page') }}
+				</strong>
+			</div>
+			<div class="description">
+				{{ t('collectives', 'This link points to a Collectives page, but the page either does not exist or you are not allowed to view it.') }}
```
mejo-:
```suggestion
				{{ t('collectives', 'The page does not exist or you are not allowed to view it.') }}
```

src/views/PageReferenceWidget.vue
```diff
@@ -4,8 +4,31 @@
...
+		class="collective-page not-found"
+		@click="clickLink">
+		<div class="collective-page--image">
+			<PageIcon
+				:size="50" />
+		</div>
+		<div class="collective-page--info">
+			<div>
+				<strong>
+					{{ t('collectives', 'Cannot preview page') }}
```
mejo-:
```suggestion
					{{ t('collectives', 'Page not found') }}
```

src/views/PageReferenceWidget.vue
```diff
@@ -72,17 +98,27 @@ export default {
...
+		t,
+
 		clickLink(event) {
 			const appUrl = '/apps/collectives'
 			const linkUrl = new URL(this.richObject.link, window.location)
-			// Only consider rerouting if we're inside the collectives app and for links to collectives app
+			// Only consider rerouting if we're inside the collectives app excluding public shares) and for links to
+			// collectives app.
 			if (window.OCA.Collectives?.vueRouter
+				&& !this.isPublic
```
mejo-: I don't understand the rationale behind this change. In public shares I still want links to other pages in this collectives to be opened by re-routing them, not in a new tab, no?

src/views/PageReferenceWidget.vue
```diff
@@ -4,8 +4,31 @@
...
 
 <template>
+	<!-- handle inaccessible pages -->
 	<a
-		v-if="richObject"
+		v-if="richObject && notFound"
+		:href="richObject.link"
+		target="_blank"
+		class="collective-page not-found"
+		@click="clickLink">
```
mejo-: I'd remove the click handler here as I'd expected links to unknown pages to be opened in a new tab, instead of trying to reroute them via vue-router.

src/views/PageReferenceWidget.vue
pymnh: Maybe I misunderstand the the rerouting logic here. What I wanted to achieve here was to not reroute within the collectives page if I am an unauthenticated user accessing a public share, since I would be prompted with a useless left sidebar and a greyed out settings button in the left corner.

src/views/PageReferenceWidget.vue
pymnh: > I'll go ahead and push the to changes I suggested to still get this into the next release. I hope that's ok for you.

Sure thing!

> I'd remove the click handler here as I'd expected links to unknown pages to be opened in a new tab, instead of trying to reroute them via vue-router.

See my other comment

src/views/PageReferenceWidget.vue
pymnh: Duh, now I get it! 🤦‍♀️ Of course, you are right! Simply removing the link handler is the correct way.

Issue #2345, opened 2026-03-16
Exception type mismatch in error handling


DorraJaouad:
When anonymous users search through a public page share link, the search scans the entire Collective but attempts to filter unauthorized results using checkPageShareAccess(). However, this function throws `OCSForbiddenException` instead of the expected `NotPermittedException`, causing the exception to bypass the catch block and crash the API with HTTP 403 Forbidden.

PR #2346, opened 2026-03-16
fix: catch OCSForbiddenException in public page search

saschabuehrle:
## Bug
[Issue #2345](https://github.com/nextcloud/collectives/issues/2345) — When anonymous users search through a public page share link, the search scans the entire Collective but attempts to filter unauthorized results using checkPageShareAccess(). However, this function throws `OCSForbiddenException` instead of the expected `NotPermittedException`, causing the exception to bypass the catch block and crash the API with HTTP 403 Forbidden.

## Fix
This PR updates the exception handling in the `contentSearch` method to catch both `NotPermittedException` and `OCSForbiddenException`. This ensures that unauthorized search results are properly filtered out instead of causing the entire API call to fail.

## Testing
The fix ensures that anonymous users can search through public page shares without encountering HTTP 403 errors when some results are outside their access permissions.

Greetings, saschabuehrle


mejo-:
Superseeded by #2592

PR #2360, opened 2026-03-21
fix(ImportDirectory): really prevent path traversal in processAttachment

mejo-:
Also add a unit test for ImportService happy path and one for detecting path traversal in attachments path.

### 🏁 Checklist

### DIFF
--- a/tests/Unit/Service/ImportServiceTest.php
+++ b/tests/Unit/Service/ImportServiceTest.php
@@ -1,0 +1,146 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace Unit\Service;
+
+use OCA\Collectives\Db\Collective;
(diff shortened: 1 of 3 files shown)


tests/Unit/Service/ImportServiceTest.php
```diff
@@ -0,0 +1,139 @@
...
+				return $pageInfo;
+			}
+		);
+		// Only used for rewriting links to pages, which we don't test here
+		$pageService->method('findByPath')
+			->willThrowException(new NotFoundException('Not found'));
+		$pageService->method('getPageFile')->willReturnCallback(
+			function (int $collectiveId, int $pageId, string $userId): File {
+				$file = $this->createMock(File::class);
+				$file->method('getContent')->willReturn($this->pageContentById[$pageId] ?? '');
```
max-nextcloud: I'd suggest throwing `NotFoundException` if `$this->pageContentById[$pageId]` does not exist - as that's what the page service does when asked about a page with an id that does not exist.
You may need to make `getOrCreate` create an entry.

Or am I missing something and there's a legit reason for accessing pages that have not been created before?

mejo-: Agreed, that's cleaner. Changed now.

PR #2369, opened 2026-03-24
fix(PageList): fix sorting via drag and drop

mejo-:
### 🏁 Checklist

### DIFF
--- a/src/stores/pages.js
+++ b/src/stores/pages.js
@@ -377,10 +377,10 @@ export const usePagesStore = defineStore('pages', {
 		},
 
 		pageTitle(state) {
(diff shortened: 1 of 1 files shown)


src/components/PageList/DraggableElement.vue
max-nextcloud, line 1: Are these changes necessary for the fix?

mejo-: Yes, that's the changes that actually fix the sorting via drag and drop. It seems like the short form to pass parameters in Vue 3.5 is not supported for VueDraggable.

PR #2376, opened 2026-03-25
fix(pages): guard against undefined pages in `favoritePages()`

mejo-:
Fixes "TypeError: (destructured parameter) is undefined"

### 🏁 Checklist

### DIFF
--- a/src/stores/pages.js
+++ b/src/stores/pages.js
@@ -268,6 +268,7 @@ export const usePagesStore = defineStore('pages', {
 			const collectivesStore = useCollectivesStore()
 			const pages = collectivesStore.currentCollective.userFavoritePages
 				.map((id) => state.pageById(id))
+				.filter(Boolean) // filter out undefined
 			return state.sortPages(pages)
 		},
 


src/stores/pages.js
```diff
@@ -253,6 +253,9 @@ export const usePagesStore = defineStore('pages', {
 					return [] // root page
 				}
 				const parent = state.pages.find((p) => (p.id === parentId))
+				if (!parent) {
+					return []
+				}
```
max-nextcloud: This will effectively treat the page as another root page. I don't think that's a good idea. I would return `undefined` instead and treat that separately. It's basically an error situation, right? A parent page could not be found for a page that still is somewhere in the state.

src/stores/pages.js
max-nextcloud, line 278: Here we could either drop all pages with undefined indices or sort them to the end in `sortOrder.byIndices`.

mejo-: Agreed, and it is not necessary for the fix. Dropped completely.

PR #2385, opened 2026-03-30
fix(store): don't persist page list display state

mejo-:
On mobile, we use `showing('details')` to define whether the page list or page details are displayed. It's not useful to have this persisted in browser storage, the default should always be to display page details.

Additionally, use browser session storage instead of local storage to persist sidebar display state. We want the sidebar display state to be persistent accross page reloads and when switching between pages, but we don't want it to be persistent accross different tabs. I.e. when opening to collectives in different tabs of the browser, toggleing the sidebar in one tab should not have an effect on the other tab.

### 🏁 Checklist


src/stores/root.js
```diff
@@ -3,18 +3,19 @@
...
-import { useLocalStorage } from '@vueuse/core'
+import { useSessionStorage } from '@vueuse/core'
 import { defineStore } from 'pinia'
 import { editorApiAttachments, editorApiReaderFileId, editorApiUpdateReadonlyBarProps } from '../constants.js'
 
 const STORE_PREFIX = 'collectives/pinia/root/'
 
 export const useRootStore = defineStore('root', {
 	state: () => ({
-		showings: useLocalStorage(STORE_PREFIX + 'showings', {}),
```
silverkszlo, line 14: in the localStorage of my browser I can still find the following keys:
```
collectives/pinia/root/activeSidebarTab:"attachments"
collectives/pinia/root/showings:"{"details":true}"
```
I guess that the code won't write to them anymore, but they will also linger there unless cleaned up explicitly?

mejo-: You're right, but I'd rather accept the small footprint instead of adding complex cleanup code. Anytime the local storage is cleared, this will go away.

Issue #2382, opened 2026-03-27
Allow to change default Collectives folder via admin webinterface


frederikfaye:
During the upgrade to v4.1.0 (or possibly v4.0.0) you seem to changed the default directory of the Collectives files to `/.Collectives` instead of the usual `/Collectives`, making it hidden in the Files app by default. 

This breaks my organization's setup for a couple of reasons: 
1. The individual collectives wants to also share non-markdown files (e.g., pdfs), and have been doing so through a `/Collectives/<the collective name>/files/`, which is extremely convenient since it inherits the user read/write rights as set through the user management in the Collectives app. 
2. After the update that changed the default location to a hidden one, people can no longer find their folders, and any shortcuts on desktop set up to these folder locations are broken.
3. For some reason, it seems that Windows users report that sync is no longer possible due to `:` in filenames in the `/.Collectives` folder, which wasn't a problem up until now. 

- How can I (as the admin) **globally** reset the default Collectives folder location to `/Collectives`?
- Please do not do this in the future. 

Thank you kindly.

frederikfaye:
@mejo-, thanks for the reply :)
Unfortunately, it's only a limited set of occ commands that are available, and this one is not included. 

Having an admin option through the web interface to set the default folder location on disk would be ideal. But I also just wanted to stress how disruptive a change such as this is, and that I think it's important to handle it as such, which doesn't seem to be the case right now.

mejo-:
> I have a similar problem: Since the folder was changed the users can not open the files any more. The server returns 503 "Cannot decrypt this file, probably this is a shared file. Please ask the file owner to reshare the file with you." .

@tobexyz that's a separate problem and it's discussed in https://github.com/nextcloud/collectives/issues/2326.

frederikfaye:
Thanks for the understanding, the quick responses, and for all your awesome work! :)

PR #2391, opened 2026-03-31
feat(settings): allow to set default_user_folder via admin settings

mejo-:
Fixes: #2382

#### 🖼️ Screenshots

| 🏡 After |
|---|
| image |

### 🏁 Checklist


mejo-:
> Didn't test thoroughly but how is it handled if the collectives folder already has contents (its moved to foldername (1) apparently) or if the folder is a receiving share etc.

If the mountpoint target already exists and is an empty folder, it's removed. In all other cases, it's renamed to `name (1)`.

> Also is it expected that when a default user folder is set the users can't change it anymore? In collectives it keeps saying create a "collective first" when trying to change the folder

Nope, the default folder just defines what is used per default. Users still can overwrite it with a custom user setting.

PR #2411, opened 2026-04-09
fix(build): clean stale CSS artifacts before each build

### LINKED ISSUE #2408: Collectives is loading many css files, all replicated
File here attached with many replicated content: [main-knJ3eTTo.chunk.css](https://github.com/user-attachments/files/26577681/main-knJ3eTTo.chunk.css)

- Collectives 4.2.0 on Nextcloud 32.0.8

`css` folder on disk contains 687 files and weights 158 MiB !

From the Webserver:

Image

All these files are identical (content).

Image

solracsf:
### 📝 Summary

* Resolves: https://github.com/nextcloud/collectives/issues/2408

Generated CSS build artifacts (`*.chunk.css`, entry CSS files) were never cleaned between builds. Each run of `npm run build` added new hashed files to `css/` without removing old ones, causing them to accumulate silently (gitignored files don't fail `git status --porcelain`), so the check in `make build` never caught it. The `rsync` step then copied the entire bloated `css/` directory into the release tarball, and users ended up with hundreds of stale CSS files (and ~160 MB of CSS) after extracting the app.

### 🏁 Checklist


max-nextcloud:
Hi @solracsf 

Thanks a lot for finding this and tackling it yourself. :heart: 

Approach looks good to me but I would like to hear from @mejo- what he thinks and if this fixes and actual build.

We'll get back to you next week.

mejo- (changes requested):
Thanks so much @solracsf, indeed a good finding!

I did some research and came to the conclusion that it would be cleaner to move `css/collectives.css` into `src`, make `css` and an asset-only directory that can be completely cleaned both in Makefile and by vite. Are you ok with me taking over the PR and pushing the required changes?

solracsf:
@mejo- sure, go ahead 🚀

Issue #2326, opened 2026-03-10
Some pages in the collectives-directory are unaccessible when inside the collectives-folder


dennmuel:
**Describe the bug**
Some pages and the related .md-files cannot be opened with the collectives app or via the markdown editor in the files app, as long as they reside somewhere inside the directory that is designated to be the directory for the collectives app.

The problematic files/pages are shown empty in the collectives app and trigger the following error when opening it with collectives or from the file app...
```
DecryptionFailedException Cannot decrypt this file, probably this is a shared file. Please ask the file owner to reshare the file with you.
Uncaught exception 
```
...even if server-side encryption is disabled and while I am the owner and sole user of the collective.

However, viewing the files offline, they are absolutely fine.

Interestingly, the problematic behavior disappears (meaning md-files can be opened via the files app) as soon as they do not reside in the collectives-folder (e.g. by copying/renaming the collectives-folder or changing the collectives-folder in the settings of the collectives app).

**To Reproduce**
Sorry, I have not yet figured out, why some files (in the same collective) are affected and others are not.

**Expected behavior**
All files in the collectives-folder can be opened via the collectives app and the markdown editor from the files app.

**Screenshots**
Exact copy of collectives folder:

<img width="3692" height="178" alt="Image" src="https://github.com/user-attachments/assets/159beae9-1789-41ee-a34d-d80d5bbcf6ea" />

Files do have content:

<img width="2525" height="1321" alt="Image" src="https://github.com/user-attachments/assets/2a98e431-47b4-44ae-8932-d34f8d1288c3" />

Files behave differently when opened via files app:

<img width="5050" height="2650" alt="Image" src="https://github.com/user-attachments/assets/42ae9873-1bda-4787-9a24-1baef10c9a9b" />

Page displays nothing when opened in collectives app and seems to trigger error:

<img width="10100" height="5100" alt="Image" src="https://github.com/user-attachments/assets/43e2e251-c7a7-46ca-9a2a-2865869f0220" />

**Server details:**
- Collectives app version: 4.0.0
- Nextcloud version: 32.0.6 (Docker)
- PHP Version: [e.g. 8.3]
- Database: [e.g. MariaDB 10.6]

**Client details:**
 - OS: Debian 13
 - Browser: Firefox
 - Browser version: 140.8.0esr
 - Device: Notebook

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
[...]

PR #2422, opened 2026-04-15
fix: Repair command to copy encryption keys to new path

enjeck:
workaround for https://github.com/nextcloud/collectives/issues/2326

 It has a repair command for the hidden/non-hidden Collectives mount regression when server-side encryption is enabled. It scans the effective Collectives user folder, checks the corresponding alias path, and copies only missing encryption keys into the currently used path


enjeck:
too many changes, could be easily work around with rsync copying for keys

PR #2352, opened 2026-03-19
feat(editor): pass cached team members to Text on mention

mejo-:
Fixes: nextcloud/text#6881

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits
- [x] Tests (unit, integration and/or end-to-end) passing and the changes are covered with tests

PR #2443, opened 2026-04-22
Perf: Optimize member list for collective with large number of members

Koc:
### 📝 Summary
This PR improves FE performance for a Collectives with large number of members by adding `RecycleScroller` (like in other places in the app) to the members list in popup.

:gift: Also it fixes calculation of members to display on the landing page. Right now we have 3 members on initial page load. But after window resize even for 1px smaller we're recalculating members to show. With my adjustment we have same members list for more or less similar screen sizes.

#### 🖼️ Screenshots

image

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 TODO
- [x] Add limit 15 for members loading query on the landing page (as a depends on https://github.com/nextcloud/circles/pull/2112)

### 🏁 Checklist

### DIFF
--- a/src/views/CollectiveView.vue
+++ b/src/views/CollectiveView.vue
@@ -221,7 +221,7 @@ export default {
 					promises.push(this.getTemplates(setLoading))
 				}
 				if (!this.isPublic) {
-					promises.push(this.getCircleMembers(this.currentCollective.circleId))
+					promises.push(this.getCircleMembers(this.currentCollective.circleId, 15))
 				}
 			}
 
--- a/src/views/CollectiveView.vue
+++ b/src/views/CollectiveView.vue
@@ -221,7 +221,7 @@ export default {
 					promises.push(this.getTemplates(setLoading))
 				}
 				if (!this.isPublic) {
-					promises.push(this.getCircleMembers(this.currentCollective.circleId, 15))
+					promises.push(this.getCircleMembers(this.currentCollective.circleId))
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


src/views/CollectiveView.vue
```diff
@@ -221,7 +221,7 @@ export default {
 					promises.push(this.getTemplates(setLoading))
 				}
 				if (!this.isPublic) {
-					promises.push(this.getCircleMembers(this.currentCollective.circleId))
+					promises.push(this.getCircleMembers(this.currentCollective.circleId, 15))
```
mejo-: We recently added logic to always provide the members of the collective first as suggestions when mentioning a user with `@...` in PR #2352.

One downside I see with this your changes is that this will be limited to the first 15 members of the collective in future.

To be honest, I wonder whether it would be a better approach to add a lightweight API endpoint in circles/teams that gets only the user names + display names of all circle members of type user (also indirect ones) and use that one. We don't need the full member graph with all the extra metadata anyway. What do you think?

Koc: in general I like your idea. But it requires additional development in another module which can delay merge a lot. So, let's separate things.

Current PR mostly focused on a virtual scroll and this part is done.

I've reverted `limit` for now and will continue work on a lightweight endpoint later in follow-up PRs.

mejo-: Agreed.

mejo- (approved):
Thanks again for the contribution @Koc, much appreciated. I adjusted the height CSS property to make sure the whole modal height is used.

PR #2448, opened 2026-04-26
fix(share): don't expose share password via JSON serialization

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `lib/Db/CollectiveShare.php:L57`

The `CollectiveShare` entity includes the `password` field in `jsonSerialize()`. If this object is returned by API endpoints or logged, plaintext share passwords may be exposed to clients, browser tooling, logs, or intermediaries. Share passwords should be treated as secrets and never returned after creation/update.

## Solution

Remove `password` from serialized output entirely. If needed for UI state, return only a boolean flag (e.g., `hasPassword`) and handle password updates through dedicated write-only endpoints.

## Changes

- `lib/Db/CollectiveShare.php` (modified)
- `lib/Db/Session.php` (modified)


mejo- (approved):
Thanks @tomaioo, nice finding. I agree that we should not expose the share password.

Your fix was incomplete though as we indeed need to know in the frontend is password protected. I pushed another commit to the PR that fixes this (and another bug along the way).

The session token on the other hand needs to be exposed in the response to the init session API endpoint as that's what the client needs to authenticate in the session afterwards. So that's not a security problem but by design. I therefore removed this commit from the PR.

I hope it's fine that I took over your PR. Please let me know if not 😉

benjaminfrueh (approved):

Issue #2299, opened 2026-02-26
Allow to filter pages by mentioned user


mejo-:
**Is your feature request related to a problem? Please describe.**
Currently it's not possible to search for pages where I have been mentioned. In order to get a quick overview of pages where I got mentioned (e.g. because tasks got assigned to me in meeting notes) it would be helpful to do that.

Unfortunately even full-text search currently doesn't find pages where the username is mentioned.

**Describe the solution you'd like**
We should allow to search and filter for pages where my user got mentioned.

In the page list filter field, typing `@...` should offer the possibility to filter by mentioned user and offer my own user first, other members of the collective next.

**Describe alternatives you've considered**
We probably need support for searching by mentioned user in the full text search implementation. So this should probably be a follow-up of #2118.

PR #2452, opened 2026-04-28
enh: improve full text search

benjaminfrueh:
This PR improves the existing full text search implementation with several enhancements.

### Changes

- Store original terms alongside their stems to enable exact match ranking over stem matches
- Search now uses a three-step ranking: 1. exact match, 2. prefix match, 3. stem match with fuzzy fallback
- Added stop words for `de` and `en` to reduce search index size
- Skip `.attachments` and `.templates` directories when indexing
- Use `collective_id` (integer) instead of `circle_unique_id` (string) to reduce search index size
- Normalize mentions (e.g. `@[name](mention://user/name)` becomes `@name`) to make them searchable https://github.com/nextcloud/collectives/issues/2299
- Strip HTML tags before indexing to avoid noise
- Use the actual matched term to calculate the snippet position in search results

### Migration

On migration, all existing index data is truncated before schema change, to ensure the index includes both terms and stems after the next indexing.

### DIFF
--- a/lib/Migration/Version040301Date20260427000000.php
+++ b/lib/Migration/Version040301Date20260427000000.php
@@ -1,0 +1,133 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Collectives\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version040301Date20260427000000 extends SimpleMigrationStep {
+
+	public function __construct(
+		private readonly IDBConnection $connection,
+	) {
+	}
+
+	public function preSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('collectives_s_docs')) {
+			$this->connection->truncateTable('collectives_s_docs', false);
+		}
+		if ($schema->hasTable('collectives_s_words')) {
+			$this->connection->truncateTable('collectives_s_words', false);
+		}
+		if ($schema->hasTable('collectives_s_files')) {
+			$this->connection->truncateTable('collectives_s_files', false);
+		}
+	}
+
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$this->migrateWordsTable($schema);
+		$this->migrateDocsTable($schema);
+		$this->migrateFilesTable($schema);
+
+		return $schema;
+	}
+
+	private function migrateWordsTable(ISchemaWrapper $schema): void {
+		if (!$schema->hasTable('collectives_s_words')) {
+			return;
+		}
+
+		$table = $schema->getTable('collectives_s_words');
+
+		if (!$table->hasColumn('collective_id')) {
+			$table->addColumn('collective_id', Types::BIGINT, ['notnull' => true, 'default' => 0]);
+		}
+
+		if ($table->hasColumn('circle_unique_id')) {
+			$table->dropColumn('circle_unique_id');
+		}
+
+		if ($table->hasIndex('s_words_circle_term')) {
+			$table->dropIndex('s_words_circle_term');
+			$table->addUniqueIndex(['collective_id', 'term'], 's_words_collective_term');
+		}
+
+		if ($table->hasIndex('s_words_circle')) {
+			$table->dropIndex('s_words_circle');
+			$table->addIndex(['collective_id'], 's_words_collective');
+		}
+
+		if (!$table->hasColumn('stem')) {
+			$table->addColumn('stem', Types::STRING, [
+				'notnull' => false,
+				'length' => 50,
+				'default' => null,
+			]);
+			$table->addIndex(['collective_id', 'stem'], 's_words_collective_stem');
+		}
+	}
+
+	private function migrateDocsTable(ISchemaWrapper $schema): void {
+		if (!$schema->hasTable('collectives_s_docs')) {
+			return;
+		}
+
+		$table = $schema->getTable('collectives_s_docs');
+
+		if (!$table->hasColumn('collective_id')) {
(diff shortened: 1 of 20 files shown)


lib/Search/FileSearch/FileIndexer.php
```diff
@@ -117,20 +119,25 @@ private function getDirectoryFiles(Folder $folder, bool $recursive = false): arr
 				continue;
 			}
 
+			if (str_starts_with($node->getParent()->getName(), '.attachments')
```
mejo-:
```suggestion
			if (str_starts_with($node->getParent()->getName(), '.attachments.')
```

We can add the extra dot for more exact matching.

lib/Search/FileSearch/FileIndexer.php
```diff
@@ -117,20 +119,25 @@ private function getDirectoryFiles(Folder $folder, bool $recursive = false): arr
...
+				|| str_starts_with($node->getParent()->getName(), '.templates')) {
```
mejo-:
```suggestion
				|| $node->getParent()->getName() === '.templates') {
```

For `.templates` we can do an exact string match.

lib/Migration/Version040301Date20260427000000.php
```diff
@@ -0,0 +1,124 @@
...
+
+class Version040301Date20260427000000 extends SimpleMigrationStep {
+
+	public function __construct(
+		private readonly IDBConnection $connection,
+	) {
+	}
+
+	public function preSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		$this->connection->executeStatement('TRUNCATE TABLE `*PREFIX*collectives_s_docs`');
```
mejo-: Apparently SQLite doesn't support 'TRUNCATE TABLE'. `IDBConnection` seems to have `truncateTable(string $table, bool $cascade)` though, so maybe better to use that.

mejo-: Maybe better to also check for existence of the tables here like you do in the `migrate*Table()` functions.

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -31,39 +30,57 @@ public function __construct(
 	) {
 	}
 
-	public function search(string $circleId, string $query, int $limit = self::DEFAULT_LIMIT): array {
+	public function search(int $collectiveId, string $query, int $limit): array {
 		$tokens = $this->tokenizer->tokenize($query);
```
mejo-, line 45: If I understand correctly, not passing a language here will result in stop words not being excluded. So searching for "die Katze" will search for both "die" and "Katze", despite "die" being a stopword. Not sure whether it's a problem though.

lib/Service/SearchService.php
```diff
@@ -36,11 +36,11 @@ public function indexCollective(Collective $collective, $incremental = false): v
 			throw new FileSearchException('Collectives search service could not find folder for collective.', 0, $e);
 		}
 
-		$this->indexer->indexFolder($collectiveFolder, $collective->getCircleId(), $incremental);
+		$this->indexer->indexFolder($collectiveFolder, $collective->getId(), $incremental);
 	}
 
-	public function searchCollective(Collective $collective, string $query, int $maxResults = 15): array {
-		return $this->searcher->search($collective->getCircleId(), $query, $maxResults);
+	public function searchCollective(Collective $collective, string $query, int $maxResults = 50): array {
```
mejo-: I wonder whether it's better to keep the default lower now that I realized that we had it at 15 before. How about 25?

mejo- (changes requested):
Looks really nice, thanks. I have some comments.

lib/Search/FileSearch/Tokenizer/AbstractTokenizer.php
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Collectives\Search\FileSearch\Tokenizer;
+
+abstract class AbstractTokenizer {
+	public static function normalize(string $text): string {
+		// strip HTML tags
+		$text = preg_replace('/<[a-zA-Z\/][^>]*>/', '', $text) ?? $text;
```
mejo-, line 19: I wonder whether it would be better to use [strip_tags()](https://www.php.net/manual/en/function.strip-tags.php).

benjaminfrueh: I did have `strip_tags` here initially, and then it caused many weird issue with my test collective, with some markdown documents. It was stripping too much and some documents ended up incomplete or even empty, something like content between `<!-- -->` get stripped and more weird cases 🙈 So I changed it to the preg_replace to make sure only <> and </> tags are replaced.

mejo-: Ok, good to know. Maybe we should add a short comment that `strip_tags()` is too greedy then to not forget about this in the future :laughing:

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: In this case `die` will not be found because it is not the index, but `Katze` will be found, so it should not be an issue. I think for searching the problem is that we can not reliably detect the language from just the search term, so we stem in every language of the collective.

mejo-: Yes, this makes sense. Thanks for thinking it through again. Agreed that it's not an issue then.

lib/Search/FileSearch/Tokenizer/AbstractTokenizer.php
benjaminfrueh: I added a more detailed comment for both regex to give some context here.

lib/Search/FileSearch/FileIndexer.php
benjaminfrueh: Thank you, I updated this now!

lib/Search/FileSearch/FileIndexer.php
benjaminfrueh: Thank you, I updated this now!

lib/Migration/Version040301Date20260427000000.php
benjaminfrueh: Thank you, I updated this now!

benjaminfrueh: Thank you, I updated this now!

lib/Service/SearchService.php
benjaminfrueh: Thank you, I updated this now, also I saw that the `PublicPageController` was still passing `100`, so I removed the argument to fallback to `25`.

mejo- (approved):
Once again really nice work 😍

PR #2457, opened 2026-04-29
Authorize hidden default folder

Jerome-Herbinet:
Please check commit. 

For years, my “Collectives” storage folder has been hidden (/.Collectives). I'd like to allow this as the default folder.

**⚠️ I have no test environment. Test from reviewers is required before merging.**

### DIFF
--- a/src/views/AdminSettings.vue
+++ b/src/views/AdminSettings.vue
@@ -33,7 +33,7 @@ const defaultUserFolder = ref(adminSettings.default_user_folder)
 
 const defaultUserFolderError = computed(() => {
 	return defaultUserFolder.value !== ''
-		&& /^\/[a-zA-Z0-9-_.\/]+$/.test(defaultUserFolder.value)
+		&& !/^\/[a-zA-Z0-9-_./]+$/.test(defaultUserFolder.value)
 })
 
 const defaultUserFolderHint = computed(() => {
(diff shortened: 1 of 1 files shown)


src/views/AdminSettings.vue
```diff
@@ -33,7 +33,7 @@ const defaultUserFolder = ref(adminSettings.default_user_folder)
 
 const defaultUserFolderError = computed(() => {
 	return defaultUserFolder.value !== ''
-		&& !/^\/[a-zA-Z0-9-_/]+$/.test(defaultUserFolder.value)
+		&& /^\/[a-zA-Z0-9-_.\/]+$/.test(defaultUserFolder.value)
```
max-nextcloud: You are also inverting the condition, right? There was an `!` in front of the regexp and now there's none.

Did you intend this?

Also eslint is complaining about an unnecessary escape character: `\/`. I assume the `/` could stay as it is.

max-nextcloud (commented):
My understanding is that `.Collectives` or a translation thereof is the default collectives folder right now. But you are right in that it would be good to allow adjusting it to other hidden folder names.

However I think the code changes do not behave as you intended.

src/views/AdminSettings.vue
Jerome-Herbinet: OK ; please check the new commit.

Jerome-Herbinet:
> My understanding is that `.Collectives` or a translation thereof is the default collectives folder right now. But you are right in that it would be good to allow adjusting it to other hidden folder names.
> 
> However I think the code changes do not behave as you intended.

Fixed.
Now, is the last error related ?

Jerome-Herbinet:
As well @max-nextcloud IMO Cypress errors are not related ; this PR is just making non-visible changes...

mejo- (approved):
Thanks @Jerome-Herbinet, much appreciated 🙏

PR #2463, opened 2026-05-03
Fix clear stale pageId when navigating by slug

### LINKED ISSUE #2438: three dots Share links is not opening the side bar
**Describe the bug**

In 4.2.0 click on the three dots => Share links => should open the side bar. It only works in the landing page

**To Reproduce**
Steps to reproduce the behaviour:
1. In a subpage (not on landing page) 
2. Click on '....' => Share link
3.  Side bar should open

**Expected behavior**

**Screenshots**

**Server details:**
- Collectives app version: 4.2.0
- Nextcloud version: 32.0.6
- PHP Version: [e.g. 8.3]
- Database: [e.g. MariaDB 10.6]

**Client details:**
 - OS: [e.g. Windows/macOS/Ubuntu]
 - Browser: [e.g. Firefox, Chrome]
 - Browser version: [e.g. 112]
 - Device: [e.g. iPhone13, desktop]

<details>
<summary>Logs</summary>

```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

taiebot:
clear stale pageId when navigating by slug

### 📝 Summary

* Could Resolves: #2438 
### This is completely untested ( Collectives is not available on https://github.com/szaimen/nextcloud-easy-test), This has been worked around with vibe coding suggestion with AI. As it is a one liner it is simple to review the logic and dismiss this code if it is found not working.

Here the reasonning of the AI.

# 🚨 The real problem is *before* this code runs

`currentPageIds` depends entirely on:

- `rootStore.pageId`
- `rootStore.pageParam`

Those values are set in your `$route` watcher:

```js
this.rootStore.pageParam = val.params.page
this.rootStore.pageId = val.params.pageId ? parseInt(val.params.pageId) : null
```

Here’s the key flaw:

### When navigating via slug (`pageParam`), `pageId` does NOT get cleared  
So `pageId` stays stuck on the previous page.

Then this condition in `currentPageIds` becomes wrong:

```js
if (!rootStore.pageId && !rootStore.pageParam)
```

Because:

- `pageParam` *does* change
- but `pageId` stays stale
- so the condition never triggers
- and the resolver thinks you’re still on the old page

This is why the sharing menu breaks on non‑landing pages.

---

The correct fix is to ensure that `pageId` is cleared when the route does not include a numeric `pageId`.

That’s why the one‑liner goes here:

```js
this.rootStore.pageParam = val.params.page

// Fix: clear stale pageId when navigating by slug
if (!val.params.pageId) this.rootStore.pageId = null

this.rootStore.pageId = val.params.pageId ? parseInt(val.params.pageId) : null
```

This ensures:

- `currentPageIds` receives correct values
- `currentPage` resolves correctly
- `isLandingPage` becomes accurate
- `shareUrl` is computed correctly
- the sharing menu works everywhere

---

### 🏁 Checklist

### DIFF
--- a/src/CollectivesApp.vue
+++ b/src/CollectivesApp.vue
@@ -85,6 +85,7 @@ export default {
 				this.rootStore.collectiveParam = val.params.collective
 				this.rootStore.collectiveId = val.params.collectiveId ? parseInt(val.params.collectiveId) : null
 				this.rootStore.pageParam = val.params.page
+				if (!val.params.pageId) this.rootStore.pageId = null
 				this.rootStore.pageId = val.params.pageId ? parseInt(val.params.pageId) : null
 				this.rootStore.shareTokenParam = val.params.token
 				this.rootStore.fileIdQuery = val.query.fileId ? parseInt(val.query.fileId) : ''


src/CollectivesApp.vue
```diff
@@ -85,6 +85,7 @@ export default {
 				this.rootStore.collectiveParam = val.params.collective
 				this.rootStore.collectiveId = val.params.collectiveId ? parseInt(val.params.collectiveId) : null
 				this.rootStore.pageParam = val.params.page
+				if (!val.params.pageId) this.rootStore.pageId = null
```
mejo-, line 88: This doesn't make any sense. The line after the added one already contains the same logic: if `val.params.pageId` is falsy, `this.rootStore.pageId` is set to null.

mejo- (commented):
I tried to follow the reasoning of the PR description, but it doesn't make any sense. When the page URL has the `<slug>-<id>` structure, `pageParam` is not used at all. Actually it's the other way around.

I also tried to manually reproduce #2438 by navigating any combination of pages without and with slug, but failed so far.

PR #2497, opened 2026-05-11
chore: migrate composables to typescript

mejo-:
### 🏁 Checklist

### 🤖 AI (if applicable)

- [x] The content of this PR was partly generated using AI tools
- [x] The AI-generated content was reviewed, comprehended and tested by me


benjaminfrueh (approved):

src/composables/useColor.ts
```diff
@@ -3,16 +3,17 @@
...
 
 /**
  * Get luminance for a color
  *
- * @param {number[]} rgbColor color in RGB format
- * @return {number}
+ * @param rgbColor color in RGB format
  */
-function getLuminance(rgbColor) {
+function getLuminance(rgbColor: number[]): number {
```
max-nextcloud: I think you could be more strict with [tuple types](https://www.typescriptlang.org/docs/handbook/variable-declarations.html#tuple-destructuring)
```suggestion
function getLuminance(rgbColor: [number, number, number]): number {
```
You could even use a type alias as this is used a number of times here:

```ts
type Rgb = [number, number, number]
```

src/composables/useReader.ts
```diff
@@ -39,7 +42,7 @@ export function useReader(content) {
 	})
 
 	const backlinkCount = computed(() => {
-		return pagesStore.backlinks(defaultReaderPage.id).length
+		return pagesStore.backlinks(defaultReaderPage.value?.id).length
```
max-nextcloud, line 56: uhh. You found a bug there, right?

src/composables/useReader.ts
```diff
@@ -144,7 +147,13 @@ export function useReader(content) {
...
 	 */
 	function updateReadonlyBarProps() {
-		const el = readerEl.value?.querySelector('page-info-bar')
+		interface PageInfoBarElement extends Element {
+			currentPage: PageInfo | null | undefined
+			canEdit: boolean
+			attachmentCount: number
+			backlinkCount: number
+		}
+		const el = readerEl.value?.querySelector<PageInfoBarElement>('page-info-bar')
```
max-nextcloud: This looks somewhat odd to me. It's an `HTML` `Element`, right? And then it has these additional properties of currentPage, canEdit, etc. Seems like we are mixing two things up here.

max-nextcloud (commented):
Thanks a lot for looking into this.

I already commented outside the review as I was not sure if I was gonna finish it. Mostly minor things. One type annotation looks off to me though..

src/composables/useReader.ts
mejo-: PageInfoBar is a [custom element](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements) and yes, it has these props. At least that's my understanding.

Though it seems like `currentPage` is not visible in dom structure, only the others are. When you look at the dom structure, it looks like this:

image

src/composables/useColor.ts
mejo-: Nice, indeed that's much better!

src/composables/useReader.ts
mejo-: I did a bit further research on this.

Vue custom elements expose their component props as JavaScript properties on the DOM element. So e.g. using `el.attachmentCount = 3` is the correct API usage.

Though only props with simple types (string, number, boolean) get exposed as HTML attributes in the DOM, complex attributes are only available as JavaScript properties. That's why `currentPage` is not visible in the DOM.

See also https://vuejs.org/guide/extras/web-components.html#passing-dom-properties and https://web.dev/articles/custom-elements-best-practices.

mejo-: I moved the interface to the top of the file and changed `extends Element` to `extends HTMLElement` as that's indeed more accurate.

src/composables/useReader.ts
mejo-: Yes, thanks to Typescript :blush:

src/composables/useReader.ts
max-nextcloud: Ah... I see. Could we move the `defineCustomElement` to the top of the composable then and use `PageInfoBarCE` to get the type?

```typescript
const PageInfoBarCE = defineCustomElement(PageInfoBar, { shadowRoot: false })
type PageInfoBarType = InstanceType<typeof PageInfoBarCE>
```

max-nextcloud: (Based on https://vuejs.org/guide/typescript/composition-api.html#typing-component-template-refs for typing of refs which also kind of applies here i think.)

mejo-: Already merged, but maybe something we can improve later on? I gave your approach a quick test and it didn't work out of the box:

With `type PageInfoBarType = InstanceType<typeof PageInfoBarCE>` and `const el = readerEl.value?.querySelector<PageInfoBarType>('page-info-bar')`, `el.attachmentCount = attachmentCount.value` results in

> TS2339: Property `attachmentCount` does not exist on type `VueElement`

PR #2500, opened 2026-05-12
chore(types): type custom element with instance type

max-nextcloud:
Follow up for https://github.com/nextcloud/collectives/pull/2497#discussion_r3224791556

### DIFF
--- a/src/composables/useReader.ts
+++ b/src/composables/useReader.ts
@@ -19,12 +19,8 @@ import { useSearch } from './useSearch.ts'
  * Vue custom element interface for <page-info-bar>.
  * Props are exposed as JavaScript properties on the DOM element.
  */
-interface PageInfoBarElement extends HTMLElement {
-	currentPage: PageInfo | null | undefined
-	canEdit: boolean
-	attachmentCount: number
-	backlinkCount: number
-}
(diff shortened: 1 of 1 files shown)


src/composables/useReader.ts
```diff
@@ -19,12 +19,8 @@ import { useSearch } from './useSearch.ts'
  * Vue custom element interface for <page-info-bar>.
  * Props are exposed as JavaScript properties on the DOM element.
  */
-interface PageInfoBarElement extends HTMLElement {
-	currentPage: PageInfo | null | undefined
-	canEdit: boolean
-	attachmentCount: number
-	backlinkCount: number
-}
+const PageInfoBarCE = defineCustomElement(PageInfoBar, { shadowRoot: false })
```
mejo-, line 22: I remember that somehow the destructured element with explicit `styles` attribute was necessary to not break the CSS styles for PageInfoBar when I originally migrated Collectives to Vue 3. Did you test whether the PageInfoBar still looks clean now in browser?

max-nextcloud: ah. I see. I was wondering why that was there in the first place. Will look into it.

max-nextcloud: image

I checked and it still works fine.

mejo- (approved):
Very nice ❤️

PR #2518, opened 2026-05-17
Perf: Optimize member handling in collectives with a large number of members

Koc:
## 📝 Summary
This is continuation of #2443. 

I've generated 1200 users using `./occ user:add` and assigned them to a single collective using `./occ circles:members:add`. Here are my findings:

### 1. Members list endpoint works really slow
I had an attempt to implement lightweight API endpoint in circles module but stuck with it due to federated users. This is a really loong path which I'd prefer to avoid. Also keep in mind that we're reloading this members list every 1 minute due to [background fetcher](https://github.com/nextcloud/collectives/blob/8c56442bbc828760c6a06c75a70efe0ba00d920a/src/views/CollectiveView.vue#L186-L191) and endpoint latency goes up to 10s in our organization.

So, instead of that we're loading only 15 users for a landing page. And then re-loading all users once we toggle editor mode from preview to edit. This improves performance of BE without degradation of the users autocomplete functionality in the editor.

Feel the difference :smiley: 
image

### 2. vDOMException: The quota has been exceeded
This happens because we store `circlesMembers` in the local storage. I've changed it to a local variable. As an option - we can store same 15 users but this would add even more complexity.

At the same time we're still store tons of keys like  `nextcloud_per_bmV4dGNsb3Vk_user-has-avatar.admin` but this is a separate topic, let's handle it in the follow-up PRs.

image

### 3. There is no skeleton loading in the member picker popup
We have bad UX as endpoint can respond up to 10s in our case. I've added it

image

### 4. Full list reloading for an every member change
Every member change like add/remove/change level causes full list reloading. As we already mentioned - this is slow operation. When we managing large circle it becomes a nightmare. So I've added in-place update to state instead of full re-fetching.

## :thought_balloon: Future improvements
From my PoV we should move users management into a single place, e.g. to Contacts module. Because right now we are doing more or less same things, see https://github.com/nextcloud/contacts/pull/5330

## 🏁 Checklist

## 🤖 AI (if applicable)


src/components/Page/TextEditor.vue
```diff
@@ -121,9 +122,15 @@ export default {
...
 		})
 
-		this.textEditWatcher = this.$watch('isTextEdit', (val) => {
+		this.textEditWatcher = this.$watch('isTextEdit', async (val) => {
 			if (val === false) {
 				this.stopEdit()
+			} else if (val === true) {
+				// Load full circle members for autocomplete when entering edit mode
+				const circlesStore = useCirclesStore()
+				if (!circlesStore.currentCircleMembersFullyLoaded && !this.isPublic && circlesStore.currentCircleMembers?.length > 0) {
```
mejo-: What's the reason for the `circlesStore.currentCircleMembers?.length > 0` check here? Do you only want to call `getCircleMembers()` here if it already got called before? If so, why?

src/stores/circles.js
```diff
@@ -108,10 +115,14 @@ export const useCirclesStore = defineStore('circles', {
 		 * Get members of a team
 		 *
 		 * @param {string} circleId ID of the team
+		 * @param {number} limit Limit of members to fetch, 0 for all members
 		 */
-		async getCircleMembers(circleId) {
-			const response = await axios.get(generateOcsUrl(`apps/circles/circles/${circleId}/members?fullDetails=true`))
+		async getCircleMembers(circleId, limit = 0) {
+			const response = await axios.get(generateOcsUrl(`apps/circles/circles/${circleId}/members?fullDetails=true&limit=${limit}`))
 			this.circlesMembers[circleId] = response.data.ocs.data
```
mejo-: Unfortunately this introduces a race condition. Whenever `getCircleMembers()` gets called with `limit=15` later on, `circlesMembersFullyLoaded` stays untouched but `circlesMembers` gets reset to only hold the first 15 members.

mejo- (dismissed):

Koc:
@mejo- heh, it seems like that topic is much more complex. I've updated the PR description and pushed more changes. Please, re-review

src/stores/circles.js
```diff
@@ -108,10 +116,23 @@ export const useCirclesStore = defineStore('circles', {
...
-			const response = await axios.get(generateOcsUrl(`apps/circles/circles/${circleId}/members?fullDetails=true`))
-			this.circlesMembers[circleId] = response.data.ocs.data
+		async getCircleMembers(circleId, limit = 0) {
+			// Skip if already fully loaded or request is pending
+			if (this.circlesMembersFullyLoaded[circleId] || this.circlesMembersPending[circleId]) {
+				return
+			}
+			this.circlesMembersPending[circleId] = true
+			try {
+				const response = await axios.get(generateOcsUrl(`apps/circles/circles/${circleId}/members?fullDetails=true&limit=${limit}`))
```
Koc, line 130: `fullDetails=true` slowdowns endpoint by x4. And this needed only for detection is member is list current user or not

```js
// CurrentMembers.vue
		isCurrentUser() {
			return (item) => {
				return item.userId === this.currentUser
					|| item.singleId === item.circle.initiator.singleId
			}
		},
```

We can think about improving that in follow-up PR

Koc:
@max-nextcloud please review and help me with Cypress tests :pray:  (if you have time for that)

max-nextcloud:
@Koc It's on my list. Can't promise anything unfortunately.

Koc:
@max-nextcloud @mejo- mates, I've fixed Cypress tests in the end of June as well and we have completely green pipeline. Any chance to merge this one?

src/stores/circles.js
max-nextcloud: That sounds like a good idea. It should also reduce the usage of the local storage drastically so we can bring that back.

max-nextcloud (approved):

src/stores/circles.js
Koc: opened follow-up: https://github.com/nextcloud/collectives/pull/2685

PR #2522, opened 2026-05-18
fix(PagePicker): pass title from smart picker to text

### LINKED ISSUE #1583: Use page name instead of URL when using the Smart-Picker and selecting a page from current collective

When you use "/" and select a page from the current collective (or using the smart-picker from the top bar menu without selecting any text beforehand), it creates a link with the page URL as the link text. Users may almost never need a full URL of a page from the current collective in their pages, I think people would usually want the page's name as the link text. Changing the link's text is a tedious, and when you often link pages to one another, it can become an obstacle.

The behavior could be the same as when you drag a page to the text field. It inserts the page's name as the link text and it links to the page.

**Describe alternatives you've considered**

I made a user-script. I first write the text I want to link, select it, hit a keyboard shortcut, then the script open the smart picker, selects collectives, and enter the selected text into the search input (I have to press at least one key to trigger the search, I wasn't able to trigger the search with the user-script). This makes it possible to first select the text to search and then add a link to it with a keyboard shortcut.

**Additional context**

I use links between pages from the same collective really often and dragging the pages is the smoothest way to do that right now, it's an amazing feature, but it requires you to use the mouse. And to find the page in the list of pages that can get really big you end up needing to use the search field before. That means one mouse use to click on the search field, then back to the keyboard to search, then back to the mouse to drag the page, then back to the keyboard. The proposed solution would need zero mouse use.

mejo-:
Requires: nextcloud-libraries/nextcloud-vue#8532
Requires: nextcloud/text#8615

Fixes: #1583

We should only release this after nextcloud/text#8615 got released.

### 🏁 Checklist

### DIFF
--- a/playwright/e2e/page-picker.spec.ts
+++ b/playwright/e2e/page-picker.spec.ts
@@ -43,6 +43,17 @@ test.describe('Custom page picker - local search', () => {
 		await expect(pageWidget).toHaveAttribute('href', origin + targetPage.getPageUrl())
 	})
 
+	test('link to local page has page title as link text', async ({ page, editor }) => {
+		const targetPageItem = editor.pagePicker.locator('.page-preview-item').filter({ hasText: 'Target page' })
+		await targetPageItem.click()
+
+		const targetPagePreview = editor.getContent().locator('.preview').filter({ hasText: 'Target page' })
+		await targetPagePreview.getByLabel('Actions').click()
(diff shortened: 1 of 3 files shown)


max-nextcloud:
waiting for a release of nextcloud vue.

max-nextcloud (commented):
assuming this is still in development

mejo-:
> assuming this is still in development

This is finally ready for review now.

PR #2541, opened 2026-05-27
feat: allow users to get notified about changes in a collective

### LINKED ISSUE #317: Notifications for the members of a collective
In GitLab by @robshnup on Apr 14, 2022, 08:25

There is no possibiltity to inform the members of a collective, that there was a new article/entry or contribution.
Neither by push nor by email.
It would be a big step up for working together being informed about the progress of our joint actions.

a feature update would be highly appreciated. Thanks

mejo-:
Sends notifications if pages got updated or deleted. Doesn't send notifications when a page got created on purpose as usually pages get renamed straight after they got created.

Also allows to turn notifications off completely for a collective, which also silences notifications on being mentioned in a page.

Fixes: #317

#### 🖼️ Screenshots

Collectives Action Menu 1 | Collectives Action Menu 2
---|---
image | image

| Notification |
| --- |
| image |

### 🚧 TODO

So far it only sends notifications on page updates (content updated, page renamed/moved/deleted, emoji changed). I wonder whether adding/removing members shall trigger a notification as well. And how about tagging/untagging pages etc.?

Also, so far there's only three states: "all changes", "@-mention only" (the default) and "off" (no notifications at all). I wonder whether we want to make this more fine-grained, but maybe that's something for a follow-up later on.

- [x] Decide if collective actions is the right place for the subscribe toggle in UI
- [x] Playwright tests
- [ ] Consider activity app integration

### 🏁 Checklist

### 🤖 AI (if applicable)

- [x] The content of this PR was partly generated using AI tools
- [x] The AI-generated content was reviewed, comprehended and tested by my

### DIFF
--- a/lib/Service/NotificationService.php
+++ b/lib/Service/NotificationService.php
@@ -1,0 +1,122 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Collectives\Service;
+
+use OCA\Circles\Model\Member;
+use OCA\Collectives\AppInfo\Application;
+use OCA\Collectives\Db\Collective;
+use OCA\Collectives\Db\CollectiveUserSettingsMapper;
+use OCA\Collectives\Model\PageInfo;
+use OCA\Collectives\Notification\Notifier;
+use OCP\IURLGenerator;
+use OCP\Notification\IManager as INotificationManager;
+use Psr\Log\LoggerInterface;
+
+class NotificationService {
+	public function __construct(
+		private readonly INotificationManager $notificationManager,
+		private readonly CircleHelper $circleHelper,
+		private readonly CollectiveUserSettingsMapper $settingsMapper,
+		private readonly IURLGenerator $urlGenerator,
+		private readonly LoggerInterface $logger,
+	) {
+	}
+
+	/**
+	 * Notify all members of a collective if notifications are enabled.
+	 */
+	public function notifyMembers(
+		Collective $collective,
+		PageInfo $pageInfo,
+		string $subject,
+		string $actingUserId,
+		string $pageRelativePath,
+	): void {
+		try {
+			$circle = $this->circleHelper->getCircle($collective->getCircleUniqueId(), null, true);
+		} catch (MissingDependencyException|NotFoundException|NotPermittedException $e) {
+			$this->logger->warning('Could not fetch circle members for notification: ' . $e->getMessage(), ['exception' => $e]);
+			return;
+		}
+
+		// Collect direct user member IDs
+		$memberUserIds = [];
+		foreach ($circle->getMembers() as $member) {
+			// TODO: also notify indirect circle/team members
+			if ($member->getUserType() === Member::TYPE_USER) {
+				$memberUserIds[] = $member->getUserId();
+			}
+		}
+
+		if (empty($memberUserIds)) {
+			return;
+		}
+
+		// Fetch all user settings for this collective in one query
+		$allSettings = $this->settingsMapper->findByCollectiveId($collective->getId());
+		$notifyUserIds = [];
+		foreach ($allSettings as $setting) {
+			if ($setting->getSetting('notify') === true) {
+				$notifyUserIds[] = $setting->getUserId();
+			}
+		}
+
+		// Build URLs
+		$baseUrl = $this->urlGenerator->linkToRouteAbsolute('collectives.start.index');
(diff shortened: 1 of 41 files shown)


marcoambrosini (commented):
Hi Jonas, I think that the location you've chosen makes sense, I would maybe do it like in Talk though, using a sub-action-menu, so we have more space to be clear with labels and icons.

https://github.com/user-attachments/assets/dfcacfc8-b700-4219-a679-8eaba662f1e2

On top of that, this could be indeed put in the collective settings dialog. Just like in Talk, the dialog could be exposed to all users, and have the conditional display logic (owner/user) on settings section instead. Templates could also go in here.

mejo-:
> Hi Jonas, I think that the location you've chosen makes sense, I would maybe do it like in Talk though, using a sub-action-menu, so we have more space to be clear with labels and icons.

Great idea @marcoambrosini. Implemented now, see the updated screenshots.

marcoambrosini (approved):
🥳 really nice!

blizzz:
> I wonder whether adding/removing members shall trigger a notification as well. And how about tagging/untagging pages etc.?
> Also, so far there's only three states: "all changes", "@-mention only" (the default) and "off" (no notifications at all). I wonder whether we want to make this more fine-grained, but maybe that's something for a follow-up later on.

That might be worthy activity information, but I am doubtful whether they are worth to create a notification. I would keep it as is for now. (Tagging might probably also need to discriminate between tag types and user privileges if it was implemented :thinking:)

src/apis/collectives/userSettings.js
```diff
@@ -37,7 +37,7 @@ export function setCollectiveUserSettingPageOrder(collectiveId, pageOrder) {
  */
 export function setCollectiveUserSettingShowMembers(collectiveId, showMembers) {
 	return axios.put(
-		collectiveUserSettingsApiUrl(collectiveId, 'showMembers'),
+		collectiveUserSettingsApiUrl(collectiveId, ['showMembers']),
```
max-nextcloud: The `'showMembers'` part should not need to be an array.

I think `collectiveUserSettingsApiUrl` is missing `...parts` in the call to `apiUrl`. Also the type anotation for `apiUrl` might be off.

lib/Service/NotificationService.php
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		// Fetch all user settings for this collective in one query
+		$allSettings = $this->settingsMapper->findByCollectiveId($collective->getId());
+		$notifyUserIds = [];
+		foreach ($allSettings as $setting) {
+			if ($setting->getSetting('notify') === CollectiveUserSettings::NOTIFY_ALL) {
+				$notifyUserIds[] = $setting->getUserId();
+			}
+		}
```
max-nextcloud, line 87: I think we could address the indirect members right away:

```suggestion
		// Collect direct user member IDs in one query
		$memberUserIds = [];
		foreach ($circle->getMembers() as $member) {
			if ($member->getUserType() === Member::TYPE_USER) {
				$memberUserIds[] = $member->getUserId();
			}
		}

		// Fetch all user settings for this collective in one query
		$allSettings = $this->settingsMapper->findByCollectiveId($collective->getId());
		$notifyUserIds = [];
		foreach ($allSettings as $setting) {
			if ($setting->getSetting('notify') === CollectiveUserSettings::NOTIFY_ALL) {
				$userId = $setting->getUserId();
				// Skip acting user
				if ($userId === $actingUserId) {
					continue;
				}
				if (in_array($userId, $memberUserIds, true)) {
					$notifyUserIds[] = $setting->getUserId();
					continue;
				}
				// Check for indirect membership
				$circles = $this->circleHelper->getCircles($userId);
				$circleIds = array_map(static fn ($circle) => $circle->getSingleId(), $circles);
				if (in_array($collective->getCircleId(), $circleIds, true)) {
				  	$notifyUserIds[] = $setting->getUserId();
				}	
			}
		}
```

lib/Service/NotificationService.php
```diff
@@ -0,0 +1,123 @@
...
+
+		foreach ($memberUserIds as $userId) {
+			// Skip acting and non-notify users
+			if ($userId === $actingUserId) {
+				continue;
+			}
+			if (!in_array($userId, $notifyUserIds, true)) {
+				continue;
+			}
+
```
max-nextcloud: With the change proposed above we only need to iterate over `$notifyUserIds` as group membership is already accounted for.
```suggestion
		foreach ($notifyUserIds as $userId) {

```

max-nextcloud (approved):
All in all looks good. Just a few minor comments.
Have not tested it yet though.

lib/Service/NotificationService.php
mejo-: Nice! Took over this solution!

src/apis/collectives/userSettings.js
mejo-: Good catch, fixed.

PR #2390, opened 2026-03-31
Perf: Optimize pages loading (Listeners approach)

Koc:
### 📝 Summary

This PR improves pages loading by introduction of the new column `collectives_pages.collective_id` and using pages table for loading almost without touching Filesystem api. No recursion anymore, just load all pages using single `WHERE IN` query :rocket: .

#### 🖼️ Screenshots

image

🏚️ Before | 🏡 After
---|---
image | image

### 🚧 TODO
- Migration :heavy_check_mark: :question: 
  - Do we want create composite index `collective_id, trash_timestamp` instead of just `collective_id`  :question: 
  - Create new column :heavy_check_mark: 
  - Populate new column for already existent pages :heavy_check_mark: 
 - Operations via Collective :heavy_check_mark: 
   - Create new collective -> collectiveId set for root page :heavy_check_mark:
   - Create new page -> collectiveId set for created page :heavy_check_mark: 
   - Copy page to another collective -> new collectiveId set for copied page :heavy_check_mark: 
   - Move page to another collective -> new collectiveId set for moved page :heavy_check_mark: 
 - Operations via Files :construction: 
   - Create new file inside collective folder -> a new page with collectiveId is created :heavy_check_mark:
   - Copy file to another collective -> a new page with new collectiveId and slug is created :heavy_check_mark:
   - Copy file outside of collectives folder to a collective folder -> a new page with collectiveId is created :construction: :
   - Copy folder outside of collectives folder to a collective folder -> a new pages with collectiveId are recursively created :construction: 
   - Move file outside of collectives folder to a collective folder -> a new page with collectiveId is created :heavy_check_mark: 
   - Move folder outside of collectives folder to a collective folder -> a new pages with collectiveId are recursively created :construction: 
   - Move file to another collective -> collectiveId is updated :heavy_check_mark:
   - Move file to another collective with file renaming -> collectiveId and slug are updated :heavy_check_mark: 
   - Move file to non-collective directory -> page should be trashed :heavy_check_mark:
   - Delete file -> page should be trashed :heavy_check_mark: 
   - Restore file -> page should be restored :heavy_check_mark: 
   - Directory creation -> create a new page with Readme.md -> not needed :x: 
 - Collectives loading :construction: :question:
   - Load collective by shared link  `\OCA\Collectives\Controller\PublicPageController::index` :heavy_check_mark: 
   - Load collective by shared link filtered by page  `\OCA\Collectives\Controller\PublicPageController::index` :construction: 
[...]

PR #2549, opened 2026-05-31
Perf: Optimize pages loading (Filecache path like approach)

### LINKED ISSUE #2380: Optimize performance of the pages loading by stopping using filesystem always
I profiled some endpoints, e.g. `PublicPageController::index()` and realized that we have a lot of more or less repeatable queries to database, mostly `filecache` table. For some of our collectives we have up to 500 queries per collective view. This happens because [`PageService::getPagesFromFolder()`](https://github.com/nextcloud/collectives/blob/v4.2.0/lib/Service/PageService.php#L423) recursively loads all files from a folder and then loads pages for each file one by one. The more pages we have with deeper nesting level -> the bigger quantity of queries we got.

As we already have `collectives_pages` table - let's use it to load all pages for collective in a single query. We need just add columns like `collective_id, parent_id, etc`.

Of course we need also somehow maintain this table for cases when users creating files via filesystem instead of using Collectives app. But this can be a background process.

WDYT guys?

Koc:
### 📝 Summary

This is alternative approach to #2390 that fixes same performance issue (closes #2380).

Benefits comparing to previous implementation:
- no extra columns
- no migration to re-process already existent page
- no listeners
- much simpler implementation

So, we're just load all necessary pages via simple query `SELECT * FROM filecache WHERE storage_id = <storageId> AND path LIKE 'appdata_<instanceId>/collectives/<collectiveId>/%'`

#### 🖼️ Screenshots

image

Collective with 390 pages with various nesting level

🏚️ Before | 🏡 After
---|---
**6.31s** | **5.20s**
image | image
**747** queries - depends on pages count and nesting level | **50** queries - more or less constant
image | image

### 🚧 TODO
- [x] As a future improvement we can consider to add index to `filecache` table to `storage_id, path` columns (but this requires extra PR to nextcloud/server) - see comment https://github.com/nextcloud/collectives/pull/2549#issuecomment-4765551936

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/lib/Model/PageInfo.php
+++ b/lib/Model/PageInfo.php
@@ -230,6 +230,70 @@ class PageInfo implements JsonSerializable {
 	}
 
 	/**
+	 * Build the page info from a lightweight filecache entry (see PageService::getPagesFromFolder).
+	 */
+	public function fromFileInfo(
+		CollectiveFileInfo $fileInfo,
+		int $parentId,
+		?string $collectivePath = null,
+		?string $lastUserId = null,
+		?string $lastUserDisplayName = null,
+		?string $emoji = null,
+		?string $subpageOrder = null,
+		?bool $fullWidth = false,
+		?string $slug = null,
+		?string $tags = null,
+		?array $linkedPageIds = null,
+	): void {
+		$this->setId($fileInfo->fileId);
+		$dirName = dirname($fileInfo->path);
+		$dirName = $dirName === '.' ? '' : $dirName;
+		if ($fileInfo->isIndexPage()) {
+			if ($parentId === 0) {
+				// Landing page
+				$this->setTitle(Server::get(IFactory::class)->get('collectives')->t('Landing page'));
+			} else {
+				// Index page
+				$this->setTitle(basename($dirName));
+			}
+		} else {
+			$this->setTitle(basename($fileInfo->name, self::SUFFIX));
+		}
+		$this->setFilePath($dirName);
+		$this->setTimestamp($fileInfo->mtime);
+		$this->setSize($fileInfo->size);
+		$this->setFileName($fileInfo->name);
+		if ($collectivePath !== null) {
+			$this->setCollectivePath($collectivePath);
+		}
+		if ($lastUserId !== null) {
+			$this->setLastUserId($lastUserId);
+		}
+		if ($lastUserDisplayName !== null) {
+			$this->setLastUserDisplayName($lastUserDisplayName);
+		}
+		if ($emoji !== null) {
+			$this->setEmoji($emoji);
+		}
+		if ($fullWidth !== null) {
+			$this->setFullWidth($fullWidth);
+		}
+		if ($subpageOrder !== null) {
+			$this->setSubpageOrder($subpageOrder);
+		}
+		if ($slug !== null) {
+			$this->setSlug($slug);
+		}
+		if ($tags !== null) {
+			$this->setTags($tags);
+		}
+		if ($linkedPageIds !== null) {
+			$this->setLinkedPageIds($linkedPageIds);
+		}
+		$this->setParentId($parentId);
+	}
+
+	/**
 	 * @throws InvalidPathException
 	 * @throws NotFoundException
 	 */
--- a/lib/Model/PageInfo.php
+++ b/lib/Model/PageInfo.php
@@ -245,52 +245,26 @@ class PageInfo implements JsonSerializable {
 		?string $tags = null,
 		?array $linkedPageIds = null,
 	): void {
-		$this->setId($fileInfo->fileId);
 		$dirName = dirname($fileInfo->path);
 		$dirName = $dirName === '.' ? '' : $dirName;
-		if ($fileInfo->isIndexPage()) {
-			if ($parentId === 0) {
-				// Landing page
-				$this->setTitle(Server::get(IFactory::class)->get('collectives')->t('Landing page'));
-			} else {
-				// Index page
-				$this->setTitle(basename($dirName));
-			}
-		} else {
-			$this->setTitle(basename($fileInfo->name, self::SUFFIX));
-		}
-		$this->setFilePath($dirName);
-		$this->setTimestamp($fileInfo->mtime);
-		$this->setSize($fileInfo->size);
-		$this->setFileName($fileInfo->name);
-		if ($collectivePath !== null) {
-			$this->setCollectivePath($collectivePath);
-		}
-		if ($lastUserId !== null) {
-			$this->setLastUserId($lastUserId);
-		}
-		if ($lastUserDisplayName !== null) {
-			$this->setLastUserDisplayName($lastUserDisplayName);
-		}
-		if ($emoji !== null) {
-			$this->setEmoji($emoji);
-		}
-		if ($fullWidth !== null) {
-			$this->setFullWidth($fullWidth);
-		}
-		if ($subpageOrder !== null) {
-			$this->setSubpageOrder($subpageOrder);
-		}
-		if ($slug !== null) {
-			$this->setSlug($slug);
-		}
-		if ($tags !== null) {
-			$this->setTags($tags);
-		}
-		if ($linkedPageIds !== null) {
-			$this->setLinkedPageIds($linkedPageIds);
-		}
-		$this->setParentId($parentId);
+		$this->fromData(
+			$fileInfo->fileId,
+			$dirName,
+			$fileInfo->isIndexPage(),
+			$parentId,
+			$fileInfo->name,
+			$fileInfo->mtime,
+			$fileInfo->size,
+			$collectivePath,
+			$lastUserId,
+			$lastUserDisplayName,
+			$emoji,
+			$subpageOrder,
+			$fullWidth,
+			$slug,
+			$tags,
+			$linkedPageIds,
+		);
 	}
 
 	/**
@@ -309,11 +283,51 @@ class PageInfo implements JsonSerializable {
 		?string $tags = null,
 		?array $linkedPageIds = null,
 	): void {
-		$this->setId($file->getId());
-		// Set folder name as title for all index pages except the collective landing page
 		$dirName = dirname($file->getInternalPath());
 		$dirName = $dirName === '.' ? '' : $dirName;
-		if (strcmp($file->getName(), self::INDEX_PAGE_TITLE . self::SUFFIX) === 0) {
+		$isIndexPage = strcmp($file->getName(), self::INDEX_PAGE_TITLE . self::SUFFIX) === 0;
+		$mountPoint = explode('/', $file->getMountPoint()->getMountPoint(), 4);
+		$collectivePath = count($mountPoint) >= 4 ? rtrim($mountPoint[3], '/') : null;
+		$this->fromData(
+			$file->getId(),
+			$dirName,
+			$isIndexPage,
+			$parentId,
+			$file->getName(),
+			$file->getMTime(),
+			(int)$file->getSize(),
+			$collectivePath,
+			$lastUserId,
+			$lastUserDisplayName,
+			$emoji,
+			$subpageOrder,
+			$fullWidth,
+			$slug,
+			$tags,
+			$linkedPageIds,
+		);
+	}
+
+	private function fromData(
+		int $id,
+		string $dirName,
+		bool $isIndexPage,
+		int $parentId,
+		string $fileName,
+		int $timestamp,
+		int $size,
+		?string $collectivePath = null,
+		?string $lastUserId = null,
+		?string $lastUserDisplayName = null,
+		?string $emoji = null,
+		?string $subpageOrder = null,
+		?bool $fullWidth = false,
+		?string $slug = null,
+		?string $tags = null,
+		?array $linkedPageIds = null,
+	): void {
+		$this->setId($id);
+		if ($isIndexPage) {
 			if ($parentId === 0) {
 				// Landing page
 				$this->setTitle(Server::get(IFactory::class)->get('collectives')->t('Landing page'));
@@ -322,15 +336,14 @@ class PageInfo implements JsonSerializable {
 				$this->setTitle(basename($dirName));
 			}
 		} else {
-			$this->setTitle(basename($file->getName(), self::SUFFIX));
+			$this->setTitle(basename($fileName, self::SUFFIX));
 		}
 		$this->setFilePath($dirName);
-		$this->setTimestamp($file->getMTime());
-		$this->setSize((int)$file->getSize());
-		$this->setFileName($file->getName());
-		$mountPoint = explode('/', $file->getMountPoint()->getMountPoint(), 4);
-		if (count($mountPoint) >= 4) {
-			$this->setCollectivePath(rtrim($mountPoint[3], '/'));
+		$this->setTimestamp($timestamp);
+		$this->setSize($size);
+		$this->setFileName($fileName);
+		if ($collectivePath !== null) {
+			$this->setCollectivePath($collectivePath);
 		}
 		if ($lastUserId !== null) {
 			$this->setLastUserId($lastUserId);
--- a/lib/Mount/CollectiveFolderManager.php
+++ b/lib/Mount/CollectiveFolderManager.php
@@ -15,6 +15,7 @@ use OC\Files\Node\LazyFolder;
 use OC\Files\Storage\Wrapper\Jail;
 use OC\Files\Storage\Wrapper\PermissionsMask;
 use OCA\Collectives\ACL\ACLStorageWrapper;
+use OCA\Collectives\Model\CollectiveFileInfo;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Files\Cache\ICacheEntry;
 use OCP\Files\Folder;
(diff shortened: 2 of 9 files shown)


lib/Service/PageService.php
```diff
@@ -511,6 +515,228 @@ public function getPagesFromFolder(int $collectiveId, Folder $folder, string $us
...
+	 * @param bool $recurse Descend into subfolders; if false, only their index page is added
+	 * @param bool $forceIndex Create the entry folder's index page even if it has no pages
+	 *
+	 * @return PageInfo[]
+	 *
+	 * @throws MissingDependencyException
+	 * @throws NotFoundException
+	 * @throws NotPermittedException
+	 */
+	public function getPagesFromFolderV2(int $collectiveId, Folder $folder, string $userId, bool $recurse = false, bool $forceIndex = false): array {
```
mejo-: Is there a reason to introduce this as a new function instead of replacing `getPagesFromFolder()` altogether? There's still one call left to the old function (`lib/Service/PageService.php:984`), was this left out on purpose?

I'd prefer to replace the old function and update the Unit tests along, as it would be great to not loose the existing unit testing for `getPagesFromFolder()` for the new function.

lib/Model/FileInfo.php
```diff
@@ -0,0 +1,44 @@
...
+namespace OCA\Collectives\Model;
+
+/**
+ * Lightweight representation of a file entry from the `filecache` table.
+ *
+ * `path` is relative to the collective root folder (e.g. `Readme.md` or
+ * `subfolder/page.md`), so it matches the semantics of `File::getInternalPath()`
+ * within the jailed collective storage.
+ */
+class FileInfo {
```
mejo-: Maybe better to call this `CollectiveFileInfo` to not confuse it with the public `FileInfo` API from server.

lib/Mount/CollectiveFolderManager.php
```diff
@@ -250,6 +251,53 @@ public function getFolderFileCachePerCollectiveId(array $ids): array {
...
+	 *
+	 * @throws NotFoundException
+	 * @throws \OCP\DB\Exception
+	 */
+	public function getFileCacheForCollective(int $collectiveId): array {
+		$jailPath = $this->getJailPath($collectiveId);
+		$storageId = $this->getRootFolderStorageId();
+
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select('fileid', 'storage', 'path', 'parent', 'name', 'mimetype', 'mimepart',
```
mejo-, line 273: I wonder whether this needs a guard to not fetch more than 1000 entries, which would be possible in theory on a large collective. IIRC, both SQLite and Oracle DB backends don't support more than 1000 rows per query.

Probably it's pre-emptive optimization and can be postponed as Collectives doesn't really work well with a 1000 pages. But on the other hand we're in the process to improve performance for large collectives anyway, so maybe it's a good moment to think about collectives with 1000+ pages.

We could add a limit to `999` and an optional offset parameter here and then check in `getPagesFromFolderV2()` whether the query result holds `999` rows and do follow-up queries until we have one with less rows. What do you think?

lib/Mount/CollectiveFolderManager.php
```diff
@@ -250,6 +251,53 @@ public function getFolderFileCachePerCollectiveId(array $ids): array {
...
+			'size', 'mtime', 'storage_mtime', 'encrypted', 'etag', 'permissions')
+			->from('filecache')
+			->where($qb->expr()->eq('storage', $qb->createNamedParameter($storageId, IQueryBuilder::PARAM_INT)))
```
mejo-, line 276: Any reason to not further filter on `mimetype` here and only get folders and Markdown files? If I understand the code further down correctly, we only process folders and Markdown files anyway, right?

lib/Service/PageService.php
```diff
@@ -511,6 +515,228 @@ public function getPagesFromFolder(int $collectiveId, Folder $folder, string $us
...
+		}
+
+		// Batch load page metadata, links and display names
+		$pageFileIds = [];
+		foreach ($fileInfos as $fileInfo) {
+			if ($fileInfo->isPage()) {
+				$pageFileIds[] = $fileInfo->fileId;
+			}
+		}
+		$pagesByFileId = $this->pageMapper->findByFileIds($pageFileIds);
```
mejo-: I wonder whether this could become expensive if a few hundred fileIds get passed here. Not sure how well DB backends work with a large list of parameters to `IN()`. Did you test it on a collective with 500+ pages?

lib/Service/PageService.php
```diff
@@ -511,6 +515,228 @@ public function getPagesFromFolder(int $collectiveId, Folder $folder, string $us
...
+		array $displayNames,
+		string $collectivePath,
+		string $userId,
+		array &$pageInfos,
+	): void {
+		$indexFileInfo = null;
+		$pageFileInfos = [];
+		$subFolderIds = [];
+		foreach ($childrenByParent[$folderId] ?? [] as $child) {
+			if (str_starts_with($child->name, '.')) {
```
mejo-: If we don't process hidden files anyway, maybe we can filter them out in the SQL query straight away with a `name NOT LIKE '.%'`? Or would that have negative performance impact on the query?

lib/Service/PageService.php
```diff
@@ -511,6 +515,228 @@ public function getPagesFromFolder(int $collectiveId, Folder $folder, string $us
...
+		}
+
+		if ($indexFileInfo === null) {
+			if (!$forceIndex && !$this->folderHasPages($folderId, $childrenByParent)) {
+				// Ignore folders without an index page and without any (sub)pages
+				return;
+			}
+
+			// Create missing index page if folder or subfolders have page files (or forceIndex)
+			$subFolder = $collectiveFolder->getFirstNodeById($folderId);
```
mejo-: I found the variable name `subFolder` missleading, as it's the folder object of the current folder. Why not call it `folder` instead?

lib/Mount/CollectiveFolderManager.php
```diff
@@ -250,6 +251,53 @@ public function getFolderFileCachePerCollectiveId(array $ids): array {
...
 
+	/**
+	 * Load all filecache entries (files and folders) of a collective folder in a single query.
+	 *
+	 * @return FileInfo[] Indexed by file id, with paths relative to the collective root folder
+	 *
+	 * @throws NotFoundException
+	 * @throws \OCP\DB\Exception
+	 */
+	public function getFileCacheForCollective(int $collectiveId): array {
```
mejo-: Wouldn't it be an option to pass the folder name here as well and further narrow down the query to only contain files in this subfolder? This would make the query less heavy when `getPagesFromFolderV2()` is called with a subfolder and not for the whole collective folder.

lib/Service/PageService.php
```diff
@@ -511,6 +515,228 @@ public function getPagesFromFolder(int $collectiveId, Folder $folder, string $us
...
+				$pageInfos[] = $this->buildPageInfo($subIndexFileInfo, $indexPageId, $pagesByFileId, $linkedPageIdsByFileId, $displayNames, $collectivePath);
+			}
+		}
+	}
+
+	/**
+	 * @param array<int, FileInfo[]> $childrenByParent
+	 */
+	private function findIndexPageInfo(int $folderId, array $childrenByParent): ?FileInfo {
+		foreach ($childrenByParent[$folderId] ?? [] as $child) {
```
mejo-: Note to myself: is there a more performant way to get the index page?

lib/Model/PageInfo.php
```diff
@@ -229,6 +229,70 @@ public function jsonSerialize(): array {
...
+				$this->setTitle(basename($dirName));
+			}
+		} else {
+			$this->setTitle(basename($fileInfo->name, self::SUFFIX));
+		}
+		$this->setFilePath($dirName);
+		$this->setTimestamp($fileInfo->mtime);
+		$this->setSize($fileInfo->size);
+		$this->setFileName($fileInfo->name);
+		if ($collectivePath !== null) {
```
mejo-: I guess these checks can be omitted as the default is `null` anyway, so setting them to `null` again doesn't hurt.

lib/Model/PageInfo.php
```diff
@@ -229,6 +229,70 @@ public function jsonSerialize(): array {
 		];
 	}
 
+	/**
+	 * Build the page info from a lightweight filecache entry (see PageService::getPagesFromFolderV2).
+	 */
+	public function fromFileInfo(
```
mejo-, line 235: This copies much of the logic of `fromFile()`. Maybe worth refactoring to have less code duplication?

mejo- (dismissed):
Thanks a lot @Koc, this looks really promising 🤩

I have some comments, but I'm genuinely curious what you think about the comments.

If you do further changes to the PR, could you do them in separate fixup commits (and don't force-push changes to the existing commit for now) so it's easier to review your changes?

lib/Mount/CollectiveFolderManager.php
mejo-: This would be particularly helpful for building the templateFolder page tree, as that one doesn't need to process all the files from the collective rood folder.

lib/Mount/CollectiveFolderManager.php
mejo-: Sorry, I was wrong about this. SQLite/Oracle's 1000 limit is on bound parameters (? placeholders), not on result rows.

lib/Service/PageService.php
mejo-: Well, here the limit of max `999` (SQLite) or `1000` (Oracle) hits. So we'd have to chunk these queries. Probably best done inside PageMapper::findByFileIds and PageLinkMapper::findByPageIds so callers don't need to know. But can be done in a follow-up later on.

lib/Service/PageService.php
mejo-: Ok, this seems to be a bad idea as `NOT IN` will slow down the query significantly (cannot use indexes). So let's keep the filtering of hidden files in PHP as it is already.

lib/Service/PageService.php
mejo-: Disregard, as this is only hit for non-recusive `getPagesFromFolderV2()` calls, which is only called from ImportService, so not worth the optimization.

lib/Service/PageService.php
Koc: I would say that this is temporary POC method to have possibility quickly switch between old an new implementation. So it's easier to add `V2` prefix to call.

But, anyway, I guess that old implementation should be preserved for background jobs/console commands to re-create tree.

lib/Mount/CollectiveFolderManager.php
Koc: My goal was to introduce a new index that will cover columns used in filter. So, in this case we should extend it with mime type then

lib/Service/PageService.php
Koc: Yeah, it should be fine from performance perspective. I will add workaround for 1000+ items (this is a known issue from other modules :smile: )

lib/Service/PageService.php
Koc: done

lib/Model/FileInfo.php
Koc: done

lib/Service/PageService.php
Koc: I did 1000 like we already have in [tables](https://github.com/nextcloud/tables/blob/0d9d53bf7514d32212aaa67791483f6a6e6a65fd/lib/Db/TableMapper.php#L74) and [circles](https://github.com/nextcloud/circles/blob/66d50497e9d3c11e34b1d31fe53fe7dbd53e85c3/lib/Db/CoreQueryBuilder.php#L1589)

lib/Model/PageInfo.php
Koc: yeah, but in this case we should update setter to accept null. And we should do the same for many other setters. I'd prefer to keep same code style like we have in `PageInfo::fromFile()` method.

lib/Model/PageInfo.php
Koc: I had an attempt to do that but with this approach we will have large method with all possible parameters. So we will have even more lines of code comparing to what we have now.

lib/Mount/CollectiveFolderManager.php
Koc: yeah, but as far I understand we should load all possible descendants (not only direct children).  Filecache is not nested tree/closure tree structure, so we can't efficiently load all descendants

max-nextcloud: @Koc I don't quite understand your reply. It seems we're reading @mejo- s comment differently.

I understand it as proposing something like this:
```suggestion
	public function getFileCacheForCollective(int $collectiveId, string $subdirectory = ''): array {
```

and then in line 273:

```
			->andWhere($qb->expr()->like('path', $qb->createNamedParameter($this->connection->escapeLikeParameter($jailPath . '/' . $subdirectory) . '%')));
```

So as a way to call this function to retrieve all descendants inside a subdirectory of the collective. So limiting the query not to the upper levels (direct children) but rather to the lower levels of the tree.

Koc: aah, now I got the idea! I will try, thanx for explanation

Koc: Done

lib/Service/PageService.php
Koc: Now everything stabilized, I've removed old implementation

Koc:
We have green pipeline :tada:

lib/Service/PageService.php
```diff
@@ -434,81 +438,225 @@ public static function getIndexPageFile(Folder $folder): File {
...
+		bool $forceIndex,
+		bool $recurse,
+		array $childrenByParent,
+		array $pagesByFileId,
+		array $linkedPageIdsByFileId,
+		array $displayNames,
+		string $collectivePath,
+		string $userId,
+		array &$pageInfos,
+	): void {
```
max-nextcloud: This function has a very long list of parameters. Many of these are initialized once before the function is called and then never change afterwards. I propose to move this into a separate class instead.

That way the class could be instantiated with the parameters given to the `getPagesFromFolder` function. They would then become instance variables.
Variables that are computed once and then reused such as `$childrenByParent` would become cached functions or getters.
Recursion would happen inside the class. The recursion function would only receive the parameters that actually change between different function calls.

max-nextcloud (commented):
Proposing a refactoring of the `getPagesFromFolder` / `buildPageInfoTree` functions.

I'm pretty sure your approach works as well. I think this refactoring would still clarify things a lot. Would you be up for it or shall I give it a try and open a PR against this one?

lib/Service/PageService.php
max-nextcloud: What I imagine would look something like this inside `getPagesFromFolder`:

```php
$builder = new PageInfoTreeBuilder($collectiveId, $folder, $userId, $recurse, $forceIndex);
$pageInfos = [];
$builder->build($folder->getId(), 0, $pageInfos);
return $pageInfos;
```

The `build` function would then recurse if needed with:

```php
$this->build($subFolderId, $indexPageId, $pageInfos);
```

I hope that clarifies things a bit.

max-nextcloud: Just learned that this would require a factory to combine the dependency injection with the args to the constructor like `$collectiveId`. So more complex than I imagined.

lib/Service/PageService.php
```diff
@@ -434,81 +438,225 @@ public static function getIndexPageFile(Folder $folder): File {
...
+			$pageInfos[] = $this->buildPageInfo($pageFileInfo, $indexPageId, $pagesByFileId, $linkedPageIdsByFileId, $displayNames, $collectivePath);
 		}
 
-		// Add markdown files from this folder
-		$folderPageInfos = [];
-		foreach ($pageFiles as $pageFile) {
-			if (NodeHelper::isIndexPage($pageFile)) {
+		foreach ($subFolderIds as $subFolderId) {
+			if ($recurse) {
+				$this->buildPageInfoTree($collectiveId, $collectiveFolder, $subFolderId, $indexPageId, false, true, $childrenByParent, $pagesByFileId, $linkedPageIdsByFileId, $displayNames, $collectivePath, $userId, $pageInfos);
```
max-nextcloud: Are you setting the parameter for `$forceIndex` to `false` on purpose? Does that only apply on the root level?

max-nextcloud:
Just tested this locally and performance wise it's really great. Managed to handle a collective with 2k pages well. Became somewhat slow from 3k on.
In comparison - on the same machine 700 pages were getting problematic before. So this seems to provide for approx 4-5 times as many pages.

Koc:
@max-nextcloud @mejo- WDYT about adding a composite index to `filecache` table to columns `storage_id, path`? This can improve performance even more or large instances with millions files

max-nextcloud:
@Koc there's the `fs_storage_path_prefix` cache (except for postgres dbs):
https://github.com/nextcloud/server/blob/master/core/Migrations/Version13000Date20170718121200.php#L244-L246

Might be interesting to confirm it's being used though. On my local instance requests for > 1k files were way below 1 second. That seems good enough.

max-nextcloud:
From my point of view the next step would be to fetch the pages for even larger collectives in multiple requests.
I sketched my ideas about that here: https://github.com/nextcloud/collectives/issues/2581

Koc:
@max-nextcloud heh we're using Postgres :confused: Why it's excluded?

I will try to work on your other comments next week

max-nextcloud:
> @max-nextcloud heh we're using Postgres 😕 Why it's excluded?

* Originally explained here: https://github.com/nextcloud/server/pull/28541#issuecomment-906363838
* See the discussion in https://github.com/nextcloud/server/issues/38483

The latter seems to have some discussions on creating the index by hand to avoid the DBAL limitations.

lib/Service/PageService.php
Koc: yeah, according to previous [implementation](https://github.com/nextcloud/collectives/blob/main/lib/Service/PageService.php#L457)

image

lib/Service/PageService.php
Koc: well, I did requested refactoring. Tests are still green, please re-review. I've retested basic scenarios, LGFM

lib/Model/PageInfo.php
mejo-: I imagined a private helper function (e.g. `fromData()` that gets all parameters and sets them. Then the two functions `fromFile()` and `fromFileInfo()` would become pretty small:

```php
private function fromData(
    int $id,
    string $name,
    string $dirName,
    int $mtime,
    int $size,
    bool $isIndexPage,
    int $parentId,
    ?string $collectivePath,
    ?string $lastUserId,
    ?string $lastUserDisplayName,
    ?string $emoji,
    ?string $subpageOrder,
    ?bool $fullWidth,
    ?string $slug,
    ?string $tags,
    ?array $linkedPageIds,
): void { ... }

public function fromFileInfo(CollectiveFileInfo $fileInfo, int $parentId, ?string $collectivePath = null, …): void {
    $this->fromData(
        $fileInfo->fileId,
        $fileInfo->name,
        dirname($fileInfo->path),
        $fileInfo->mtime,
        $fileInfo->size,
        $fileInfo->isIndexPage(),
        $parentId,
        $collectivePath,
        $lastUserId, $lastUserDisplayName, $emoji, $subpageOrder,
        $fullWidth, $slug, $tags, $linkedPageIds,
    );
}

public function fromFile(File $file, int $parentId, …): void {
    $mountPoint = explode('/', $file->getMountPoint()->getMountPoint(), 4);
    $collectivePath = count($mountPoint) >= 4 ? rtrim($mountPoint[3], '/') : null;

    $this->fromData(
        $file->getId(),
        $file->getName(),
        dirname($file->getInternalPath()),
        $file->getMTime(),
        (int)$file->getSize(),
        $file->getName() === self::INDEX_PAGE_TITLE . self::SUFFIX,
        $parentId,
        $collectivePath,
        $lastUserId, $lastUserDisplayName, $emoji, $subpageOrder,
        $fullWidth, $slug, $tags, $linkedPageIds,
    );
}
```

Koc: that's exactly what I said :smile: 

> but with this approach we will have large method with all possible parameters.

ok, as you wish, extracted method

max-nextcloud:
Thanks a lot for your patience and commitment @Koc 
I'll take another look to be sure and hope I can approve the PR then.

max-nextcloud (approved):
Thanks a lot! 💙

lib/Mount/CollectiveFolderManager.php
max-nextcloud: Opened a follow up issue in https://github.com/nextcloud/collectives/issues/2613

max-nextcloud:
Tracking behat failure in #2614

PR #2568, opened 2026-06-14
chore/configure dep upgrades

max-nextcloud:
- **chore(deps): wait 7 days before upgrades except for @nextcloud.**
- **chore(renovate): drop outdated version limitations**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,8 @@ updates:
     day: monday
     time: "03:00"
     timezone: Europe/Paris
+  cooldown:
+    default-days: 4
   open-pull-requests-limit: 10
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -10,6 +10,8 @@ updates:
     day: monday
     time: "03:00"
     timezone: Europe/Paris
+  cooldown:
+    default-days: 7
```
mejo-: Let's use 4 as default value for composer + npm like server and talk do. Let's keep it at 7 for github-actions.

mejo- (changes requested):

mejo- (approved):

PR #2577, opened 2026-06-16
enh: improve fulltext search

### LINKED ISSUE #2559: Sentence search does not return anything useful
**Describe the bug**
Search in collectives looks broken

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to a collective
2. Click on search
3. Start typing one word => search Ok
4. Start typing a sentence => Garbage search.

Reduced results with a specific sentence => lots of results, but they all look completely wrong there is 0 occurrences in most of them.

**Screenshots**

**Server details:**
- Collectives app version: 4.4.1
- Nextcloud version: 33.0.5
- PHP Version: [e.g. 8.3]
- Database: [e.g. MariaDB 10.6]

**Client details:**
 - OS: [e.g. Windows/macOS/Ubuntu]
 - Browser: [e.g. Firefox, Chrome]
 - Browser version: [e.g. 112]
 - Device: [e.g. iPhone13, desktop]

<details>
<summary>Logs</summary>

```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

benjaminfrueh:
Fixes #2559

### 📝 Summary
Collectives fulltext search improvements with a focus on sentence search, stem distance limiting and result ranking.

### Changes
- AND search for multi-token queries, only return documents containing all search terms
- improved multi-token ranking using a match type scoring across all matched tokens
- stem distance limit, stems too far from the original token are not indexed anymore, reducing noise
- improve snippet preview, shows the exact phrase if found, otherwise the section containing the most matched tokens
- prefix match always runs alongside exact match for the last token, to prevent short tokens from blocking results

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/lib/Search/FileSearch/FileSearcher.php
+++ b/lib/Search/FileSearch/FileSearcher.php
@@ -20,6 +20,9 @@ class FileSearcher {
 	private const SEARCH_LIMIT = 50;
 	private const FUZZY_PREFIX_LENGTH = 3;
 	public const FUZZY_MAX_DISTANCE = 1;
+	private const MATCH_TYPE_EXACT = 3;
+	private const MATCH_TYPE_PREFIX = 2;
+	private const MATCH_TYPE_STEM = 1;
 
 	public function __construct(
 		private readonly SearchWordMapper $wordMapper,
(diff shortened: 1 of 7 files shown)


lib/Search/FileSearch/FileSearcher.php
```diff
@@ -160,4 +124,55 @@ private function fuzzySearchWord(int $collectiveId, string $term): array {
...
+		// step 2: prefix match
+		$prefixWords = $this->wordMapper->findByCollectiveAndPrefix($collectiveId, $token, self::SEARCH_LIMIT);
+
+		if ($exactWord !== null || !empty($prefixWords)) {
+			$wordIds = array_unique(array_map(fn ($w) => $w->getId(), $prefixWords));
+			if ($exactWord !== null && !in_array($exactWord->getId(), $wordIds)) {
+				$wordIds[] = $exactWord->getId();
+			}
+			$matchType = $exactWord !== null ? self::MATCH_TYPE_EXACT : self::MATCH_TYPE_PREFIX;
+			return ['wordIds' => $wordIds, 'matchType' => $matchType];
```
max-nextcloud, line 147: This mixes the exact match with the prefixes and counts both as exact matches. Not sure if that's an issue though.

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -31,82 +34,43 @@ public function __construct(
...
-			$prefixWords = $this->wordMapper->findByCollectiveAndPrefix($collectiveId, $token, $remainingLimit);
-			foreach ($prefixWords as $word) {
-				if (!in_array($word->getId(), $exactWordIds)) {
-					$prefixWordIds[] = $word->getId();
-				}
+			$result = $this->findWordIdsForToken($collectiveId, $token, $languages);
+			if (empty($result['wordIds'])) {
+				continue;
 			}
+			$docs = $this->docMapper->findDocumentsByWords($collectiveId, $result['wordIds'], self::SEARCH_LIMIT);
```
max-nextcloud: Maybe we can find documents that contain all words creating a query that has an `andWhere` clause for each set of `wordIds`.

That would remove the need for an intersection later and make sure we get all the results even if one word occurs in many documents and therefore the limit to 50 results kicks in.

max-nextcloud (commented):
All in all looks good. Just some possible improvements. Not sure if for now or for later.

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -20,6 +20,9 @@ class FileSearcher {
 	private const SEARCH_LIMIT = 50;
```
max-nextcloud: Maybe have two distinct limits here?
* One for the search per word so that the intersection of multiple word searches does not become to small
* One for the default number of results returned.

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: I changed it so it only does the additional prefix match now for the last token if a exact match was found for it.
Commit: 5e2f2fd65543f964e956a4be50e1478d5c198fc0

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: I split it in two limits, `TOKEN_CANDIDATE_LIMIT` and `SEARCH_RESULT_LIMIT`
Commit: 5e2f2fd65543f964e956a4be50e1478d5c198fc0

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: Workaround this for now by having a bigger `TOKEN_CANDIDATE_LIMIT`, so we can keep the ranking functional.
Commit: 5e2f2fd65543f964e956a4be50e1478d5c198fc0

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -31,82 +35,45 @@ public function __construct(
...
-			foreach ($prefixWords as $word) {
-				if (!in_array($word->getId(), $exactWordIds)) {
-					$prefixWordIds[] = $word->getId();
-				}
+		foreach ($tokens as $index => $token) {
+			$isLastToken = $index === $lastIndex;
+			$result = $this->findWordIdsForToken($collectiveId, $token, $languages, $isLastToken);
+			if (empty($result['wordIds'])) {
+				continue;
 			}
```
max-nextcloud, line 61: This means the token did not occure in any document, right? Not as a direct match, not as a prefix, also non of the stems.

I wonder if we can send an empty response right here. If I search for some words and one of them is in none of the pages I would expect an empty result or at least a warning.

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -31,82 +35,45 @@ public function __construct(
...
+			$docs = $this->docMapper->findDocumentsByWords($collectiveId, $result['wordIds'], self::TOKEN_CANDIDATE_LIMIT);
+			$tokenResults[] = ['matchType' => $result['matchType'], 'docs' => $docs];
```
max-nextcloud, line 68: maybe also `return []` if `$docs` is empty?

lib/Search/FileSearch/FileSearcher.php
```diff
@@ -160,4 +127,57 @@ private function fuzzySearchWord(int $collectiveId, string $term): array {
...
+
+		foreach ($tokenResults as $tokenResult) {
+			foreach ($tokenResult['docs'] as $doc) {
+				$fileId = $doc['file_id'];
+				if (!in_array($fileId, $matchingFileIds)) {
+					continue;
+				}
+				$fileScores[$fileId] = ($fileScores[$fileId] ?? 0)
+					+ $tokenResult['matchType']
+					+ log1p((int)$doc['total_hits']);
```
max-nextcloud: I think right now the `matchType` does not matter for the ranking at all. Let me explain why...

We have a bunch of token results. They have different match types. For each file we add the match type of the token results they are in to their score.

But then we only consider the fileIds that occure in all token results. So each file will get the same match types added - those of all token results.

# Alternatives

## Multiplication

```suggestion
					+ $tokenResult['matchType'] * log1p((int)$doc['total_hits']);
```

So we multiply the hits score by the match type. That is if one file has 5 exact matches and 1 stem match it will be ranked before the one that has 1 exact match and 10 stem matches.

## Do not consider match types

We already prefer exact matches over prefixes over stems of the same word. If we find a direct match we don't even look for stems of the same word.

So in the example for Multiplication - say we are searching for 'Nextcloud runs' and we find exact matches for 'Nextcloud' but not for 'runs'. Is it important to rank a document that contains Nextcloud 5 times and 'run' one time higher than one that contains Nextcloud 1 time and 'run' 10 times?

I don't know. I understand it's important that an exact match of 'runs' is better than the stem match of 'run' - but that's not what our current match types reflect as they are the same for all the matches for one word in all files.

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: You are right, with the AND search it makes much more sense to return early here. With the continue you get results even with unknown words if the known words are found in the file.

lib/Search/FileSearch/FileSearcher.php
benjaminfrueh: Theoretically orphaned words should not exist but I think it makes sense as a safety net to return here too.

max-nextcloud (approved):
Tested locally. Works great! Thanks a lot!

PR #2592, opened 2026-06-24
fix(share) search subpages in page share

### LINKED ISSUE #2345: Exception type mismatch in error handling

max-nextcloud:
Fixes #2345 
Superseeds #2346

### DIFF
--- a/.github/workflows/behat-sqlite-encryption.yml
+++ b/.github/workflows/behat-sqlite-encryption.yml
@@ -121,7 +121,7 @@ jobs:
 
       - name: Behat integration
         working-directory: apps/${{ env.APP_NAME }}
-        run: ./vendor/bin/behat --config=${{ env.BEHAT_CONFIG }} --profile ci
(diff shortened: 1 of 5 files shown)


mejo- (approved):
I adjusted the tests: now the page content is written and 'occ collectives:index' called before trying to search for the page content.

mejo-:
Interesting, the new Behat test revealed that full-text search seems to be broken with server-side encryption. I can reproduce this locally. For now I'll look into a way to skip the test in the `behat-sqlite-encryption.yml` CI GitHub workflow as this is a separate issue. But definitely something we should look into later on.

mejo-:
I created a new issue for broken full-text search with server-side encryption: https://github.com/nextcloud/collectives/issues/2681

PR #2603, opened 2026-06-27
chore(test) mention search for members more stricly

max-nextcloud:
Make sure the members are actually top of the list.

Before this test was passing randomly
even though the functionality was missing on stable32.

With few users on the instance, chances were all members would be in the search result.

Tests for https://github.com/nextcloud/text/pull/8784.


max-nextcloud:
Failed as expected during first run:
grafik

max-nextcloud:
Second run with the text backport merged passed.

mejo- (approved):

PR #2604, opened 2026-06-27
fix: show absolute timestamp on hover

### LINKED ISSUE #2597: Show absolute timestamp on hovering relative date display (e.g. "last edited")
Description of the problem:

The display of edit timestamps in Collectives (e.g. "2 days ago") is shown only in relative format. There is no option to switch to an absolute date/timestamp.
I'm aware that this display behavior (relative time) is not specific to Collectives, but occurs in several areas of Nextcloud (e.g. Files app, Activity). However, since Collectives manages its own timestamp display in the page info bar, I'm filing this request separately here.

### Desired solution:

Just like in many other Nextcloud apps provide the absolute timestamp by hovering over the relative one:

Image

### Alternative:

Option to toggle between relative display and absolute date/timestamp (e.g. via click or setting), similar to comparable feature requests in other Nextcloud apps.

Decided against because it would add a config option which we try to avoid if other solutions exist.

Saithej2k:
Summary

- Add an absolute timestamp title to the relative last-updated display.
- Allow the timestamp text in the action menu item to receive hover events for the native tooltip.

Testing

- git diff --check
- npx eslint --ext .vue src/components/LastUserBubble.vue src/components/Page/NcActionLastUser.vue --max-warnings=0
- npx stylelint src/components/LastUserBubble.vue src/components/Page/NcActionLastUser.vue

Closes #2597

### DIFF
--- a/src/components/Page/NcActionLastUser.vue
+++ b/src/components/Page/NcActionLastUser.vue
@@ -62,7 +62,6 @@ export default {
 	border-radius: 0;
 	background-color: transparent;
 	box-shadow: none;
-	cursor: default;
 	font-weight: normal;
 	font-size: var(--default-font-size);
 	line-height: var(--default-clickable-area);
--- a/src/components/Page/NcActionLastUser.vue
+++ b/src/components/Page/NcActionLastUser.vue
@@ -5,13 +5,13 @@
 
 <template>
 	<li v-if="lastUserDisplayName" class="action action--user-bubble">
-		<button class="action-button action-button--user-bubble" type="button">
+		<div class="action-button action-button--user-bubble">
 			<ClockIcon :size="20" />
 			<LastUserBubble
 				:lastUserId
 				:lastUserDisplayName
 				:timestamp />
-		</button>
+		</div>
 	</li>
 </template>
 
@@ -47,14 +47,6 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-.action--user-bubble {
-	pointer-events: none;
-}
-
-.action--user-bubble :deep(.timestamp) {
-	pointer-events: auto;
-}
-
 .action-button--user-bubble {
 	display: flex;
 	align-items: flex-start;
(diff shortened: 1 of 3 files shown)


Saithej2k:
I checked the failing jobs and they look unrelated to this timestamp change. The migration job is failing while fetching origin/fix/timestamp-hover-title, which does not exist on the upstream remote for this fork PR, and the Playwright/Cypress failures are in editor/member specs. Could a maintainer rerun or confirm whether these should block this PR?

max-nextcloud:
Hi @Saithej2k 

Thanks a lot. That looks like a good start. I will look into the CI failures.

Right now the changes to the pointer behavior turn the cursor into a pointer when hovering the time.

Here's a comparison between the old and the new behavior:

https://github.com/user-attachments/assets/ecee3a84-405b-4bcb-976e-b3acdcc0da8e

I think the best fix would be to get rid of the button element in `NcActionLastUser`. This is not a button - it cannot be clicked. I think this would also remove the need for the `pointer-events` css classes all together.

But I would like to hear what @mejo- thinks about that.

max-nextcloud:
Might also be nice to have the same behavior in the recent pages list:


https://github.com/user-attachments/assets/969a871e-2679-4926-8284-e1f09c6dfa89

max-nextcloud:
https://github.com/nextcloud/collectives/pull/2612 should fix the `Static analysis of migrations with latest release`

src/components/Page/NcActionLastUser.vue
```diff
@@ -66,6 +62,7 @@ export default {
 	border-radius: 0;
 	background-color: transparent;
 	box-shadow: none;
+	cursor: default;
```
max-nextcloud: I don't think this is needed. Seems to still work fine if I disable it in the browser. Or am I missing something?

max-nextcloud (commented):
Thanks a lot for the quick update. Looks good to me now. I only have one small remark left.

src/components/Page/NcActionLastUser.vue
Saithej2k: Good catch, you’re right. I removed it in 81008220. Thanks!

max-nextcloud (approved):
🚀

PR #2661, opened 2026-07-28
chore(npm): allow scripts of known NPM dependencies

mejo-:
Starting with NPM v12, scripts defined by dependencies need to be allowed explicitely to protect against supply-chain attacks.

### 🏁 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,5 +81,14 @@
   "engines": {
     "node": "^24.0.0",
     "npm": "^11.3.0"
+  },
+  "allowScripts": {
+    "@parcel/watcher": true,
+    "core-js": true,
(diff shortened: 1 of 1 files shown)


max-nextcloud (commented):
I think I only use cypress and esbuild. Do we actually need the others?

mejo-:
> I think I only use cypress and esbuild. Do we actually need the others?

The others are indirect dependencies and we have them since a long time. Until now (with NPM < v12), their scripts get executed all the time you run `npm i` or `npm ci`. If we don't add them, `npm` will mourn and show a warning each time. I think it's acceptable to add them.

max-nextcloud (approved):

PR #2679, opened 2026-08-04
fix(files): Collectives banner activation for matching folder paths

joshtrichards:
### 📝 Summary

* Resolves: # 

Fixes a potential issue where the Collectives Files banner could be activated for paths that only partially matched the configured Collectives folder.

The previous `startsWith()` check could treat a folder such as `/Collectives-old` as being inside `/Collectives`. 

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI tools
- [ ] The AI-generated content was reviewed, comprehended and tested by a human

### DIFF
--- a/src/views/FileListInfo.vue
+++ b/src/views/FileListInfo.vue
@@ -89,7 +89,12 @@ export default {
 		},
 
 		setActive() {
-			this.active = collectivesFolder && collectivesFolder !== '/' && this.internalPath.startsWith(collectivesFolder)
+			const folder = collectivesFolder?.replace(/\/$/, '')
+			this.active = Boolean(
+				folder
+				&& folder !== '/'
+				&& (this.internalPath === folder || this.internalPath.startsWith(folder + '/')),
+			)
 		},
 	},
 }


src/views/FileListInfo.vue
```diff
@@ -89,7 +89,12 @@ export default {
 		},
 
 		setActive() {
-			this.active = collectivesFolder && collectivesFolder !== '/' && this.internalPath.startsWith(collectivesFolder)
+			const folder = collectivesFolder?.replace(/\/$/, '')
+			this.active = Boolean(
+				folder
+				&& folder !== '/'
```
mejo-, line 95: To my understanding this now no longer can match, as the replace above would change `/` to an empty string, no?

joshtrichards: The `replace` should just safely remove any trailing slash right away; creating a predictable baseline string.

mejo-: Sure, and i think it makes sense. My point was that `folder !== '/'` will now always be true, as `folder` will never again be `/` given that the trailing slash is removed beforehands.

PR #2682, opened 2026-08-06
Simplify navigation

### LINKED ISSUE #2347: Simplify navigation
It would be nice if we could simplify navigation in collectives. After some discussion the option that is preferred by the design team would be in the form of a select element at the top of the page navigation. Very much like what notion is doing.

Image

@max-nextcloud & @mejo- I think that this simplification approach is better than what we discussed in https://github.com/nextcloud/collectives/issues/2315. What are your thoughts?  

--  
This would also solve the awkward positioning of the collective name in the pages list, which when searching it now sits in between the search input and the search results.

Image

(if we adopt the proposed solution, we could move the searchbar underneath the select)

silverkszlo:
### 📝 Summary

* Resolves: #2347 

In the sidebar navigation: 
- move the list of collectives into a NcPopover
- place the search bar underneath it
- remove the old NavigationBar
- apply some refactoring.

#### 🖼️ Screenshots

Screenshot from 2026-08-19 16-43-13

Screenshot from 2026-08-19 16-43-05

### 🏁 Checklist

### 🤖 AI (if applicable)


jancborchardt (changes requested):
Very nice! :)

The new menu and the first 2 screenshots look great.

The state on the 3rd screenshot though is an empty state that people should never see. Could we remember the last used Collective and just load that on app open?

So then the only fully empty state we need to cover if someone decided to delete all collectives.

mejo- (commented):
This looks really great already ❤️ 

One quick note: I always had in mind that it would be better to migrate the whole merged collective and page list to NcAppNavigation. That means replace the current page list with NcListItem elements with NcAppNavigationItem elements. That's the intended components for app navigation and our solution to use NcListItem elements for page list was more a workaround because we had the need for two separate navigation panes.

silverkszlo:
> Very nice! :)
> 
> The new menu and the first 2 screenshots look great.
> 
> The state on the 3rd screenshot though is an empty state that people should never see. Could we remember the last used Collective and just load that on app open?
> 
> So then the only fully empty state we need to cover if someone decided to delete all collectives.

Good point, @jancborchardt ! I have adjusted the code, so now:
- the user always first sees the last collective they had visited
- or simply the first collective in the list (if it's their first time in that Collectives app)
- the empty state is displayed when there are no collectives at all

jancborchardt (changes requested):
Very cool @silverkszlo! Opening the last used Collective is a super nice quality of life feature. :) 

Some more design feedback, really mostly polishing:

https://github.com/user-attachments/assets/01dc7f93-4108-4099-91da-94b9fd58b8ba

- [x] When in an existing collective, clicking the collective name does nothing. Only clicking the emoji or the dropdown arrow opens the dropdown
- [x] The padding of the switcher is too much, seems double as much as it should be? Also a bit misaligned with the 3-dot menu when a collective was created
- [x] There are 2 clickable elements which are separately highlighted, and only the arrow has the clicky cursor
- [x] The icon on the left is misaligned / touches the edges of the clickable element
- [x] Unlike as usual in menus, the Collectives entries in the list have no gaps inbetween:
   Image
- [x] There is a bigger bottom whitespace at the end of the dropdown. Should be the same as on the left and right:
   Image

So basically all details. :) The first item I would say is quite important though as it can make it seem buggy.

(Probably separate, but we shouldn’t use the Collectives app icon as fallback for the emoji/icon, as it repeats too much. A more generic icon would be better. FYI @marcoambrosini )

silverkszlo:
Hey @jancborchardt , thank you for testing and feedback, I will get it in! One thing about the first remark: clicking the collective name actually brings you to the landing page of the collective.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+				return
+			}
+
+			const isModifierClick = event.ctrlKey || event.metaKey || event.shiftKey
+			const clickedLabel = event.target.closest('.collective-selector-label')
+
+			if (clickedLabel || isModifierClick) {
+				this.showSelector = false
+				return
+			}
```
mejo-: What's the reason for this? If I understand this correctly, when clicking on the collective name with Ctrl/Shift/Meta key pressed, this is supposed to close the selector and then continue with the browser default action for the click. I think that's a bit superfluous. If people have the selector drop-down open and then Ctrl/Shift-click on the title, I think it's acceptable that the selector drop-down stays open.

The advantage would be that `onTriggerClick()` could be simplified to just the last two lines (220+221 currently), no?

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+								:disabled="!networkOnline"
+								variant="tertiary"
+								class="new-collective-button"
+								@click="onNewCollective">
+								<template #icon>
+									<PlusIcon />
+								</template>
+							</NcAppNavigationNew>
+						</template>
+						<CollectivesTrash
```
mejo-: Nitpick, but maybe it's better to make this component the last child of the top-level `.collective-selector` div element (next to `CollectiveMembersModal`), instead of having it so deeply nested down in the default slot of the NcPopover component. It renders a full-screen modal anyway.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+							<template v-if="currentCollective?.emoji">
+								{{ currentCollective.emoji }}
+							</template>
+							<CollectivesIcon v-else :size="20" />
+						</span>
+						<span class="collective-selector-label">
+							{{ currentCollective ? currentCollective.name : t('collectives', 'Select a collective') }}
+						</span>
+					</component>
+					<button
```
mejo-: Any reason why you use a `<button>` component here instead of `NcButton`?

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+							v-if="displayTrash"
+							:networkOnline
+							@restoreCollective="onRestoreCollective"
+							@deleteCollective="onDeleteCollective" />
+						<CollectivesGlobalSettings v-if="!isPublic" :networkOnline class="collective-selector-global-settings" />
+					</template>
+				</div>
+			</template>
+		</NcPopover>
+		<NcActions
```
mejo-: From UX perspective I find it a bit irritating, that the dropdown button that allows to switch collectives, create new ones, open trash (i.e.: do actions not related to the current collective) is close to the current collective name than the three-dot action menu button that allows to do actions **on** the current collective. How about moving that one into `.collective-selector-trigger-row` in order to have it next to the collective name?

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+						:tabindex="currentCollective ? undefined : 0"
+						class="collective-selector-trigger"
+						@click="onTriggerClick"
+						@keydown.enter="onTriggerKeydown"
+						@keydown.space="onTriggerKeydown">
+						<span class="collective-selector-icon">
+							<template v-if="currentCollective?.emoji">
+								{{ currentCollective.emoji }}
+							</template>
+							<CollectivesIcon v-else :size="20" />
```
mejo-: I don't understand why, but for collectives without an emoji (you can un-set the emoji in the settings of a collective), there's no hover effect for the whole collective in the dropdown list.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+									<PlusIcon />
+								</template>
+							</NcAppNavigationNew>
+						</template>
+						<CollectivesTrash
+							v-if="displayTrash"
+							:networkOnline
+							@restoreCollective="onRestoreCollective"
+							@deleteCollective="onDeleteCollective" />
+						<CollectivesGlobalSettings v-if="!isPublic" :networkOnline class="collective-selector-global-settings" />
```
mejo-: Nitpick: after closing the FilePicker dialog to select the collectives `user_folder` (under "Collectives Settings in CollectiveSelector), the selector should stay open and collectives settings stay expanded, so one sees the change in the setting. It's unintuitive that closing the FilePicker also closes the CollectiveSelector dropdown (and with it the Collectives settings).

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+						<NcAppNavigationCaption :name="t('collectives', 'Select a collective')" />
+						<ul class="collective-selector-list">
+							<CollectiveListItem
+								v-for="collective in sortedCollectives"
+								v-show="!collective.deleted"
+								:key="collective.id"
+								:collective />
+						</ul>
+						<template v-if="!isPublic">
+							<hr class="collective-selector-divider">
```
mejo-, line 69: I just saw there's `NcAppNavigationSpacer`. Maybe that one should be used instead?

src/components/Nav/CollectiveSelector.vue
mejo-: These two should also move into the `v-if="!isPublic"` block, no? Neither the collectives trash nor the collectives settings make sense in a public share.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+		<NcPopover
+			class="collective-selector-popover-trigger"
+			:shown="showSelector"
+			:triggers="[]"
+			placement="bottom-start"
+			popoverBaseClass="collective-selector-popover"
+			@update:shown="showSelector = $event">
+			<template #trigger="{ attrs }">
+				<div class="collective-selector-trigger-row">
+					<component
```
mejo-: I don't understand why, but when any other collective but the first one is selected, there's no hover effect for the first collective in the dropdown list.

Edit: I think it's because it's focussed by default for keyboard navigation.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+	flex-grow: 1;
+	min-width: 0;
+}
+
+.collective-selector-trigger-row {
+	display: flex;
+	align-items: center;
+	min-width: 0;
+	padding-left: 4px;
+	margin-top: 5px;
```
mejo-: The `margin-top` of `collective-selector-trigger-row` breaks alignment with the collectives actions three-dot menu. I also think it looks cleaner to just remove the `margin-top`.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+:deep(.collective-selector-global-settings > div:first-child) {
+	margin-inline: 0;
+}
+
+:deep(.collective-selector-global-settings .button-vue) {
+	padding-inline: 0 !important;
+}
+</style>
+
+<style lang="scss">
```
mejo-: This is the second unscoped `style` block in this file. Maybe better merge them into one?

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+	// component is always rendered as its sibling above the page list.
+	--collective-selector-height: calc(var(--default-clickable-area) + 8px);
+}
+</style>
+
+<style lang="scss" scoped>
+.collective-selector {
+	display: flex;
+	align-items: center;
+	margin-top: 4px;
```
mejo-: I would say this can go, looks cleaner without the `margin-top`.

Also, what's the purpose of `min-height` and `justify-content`? Both don't seem to have any effect for me.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+	:deep(.app-navigation-entry) {
+		margin-inline: 4px;
+		width: calc(100% - 8px);
+	}
+
+	:deep(.app-navigation-entry-icon) {
+		flex-basis: auto;
+		width: auto;
+		justify-content: flex-start;
+	}
```
mejo-: All of these seem to get overwritten by more specific rules in the component, so they don't have any effect for me.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+
+	:deep(.app-navigation-entry-link),
+	:deep(.app-navigation-entry__name) {
+		font-weight: normal !important;
```
mejo-, line 321: Has no effect for me.

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	:deep(.app-navigation-entry:focus-within) {
+		background-color: transparent !important;
```
mejo-: This rule seems to break the hover effect on the focussed (per default: first) collectives list item. What was the purpose of it?

src/components/PageList/SubpageList.vue
```diff
@@ -4,40 +4,37 @@
...
+		:timestamp="page.timestamp"
+		:lastUserId="page.lastUserId"
+		:lastUserDisplayName="page.lastUserDisplayName"
+		:emoji="page.emoji"
+		:level
+		:canEdit="currentCollectiveCanEdit"
+		:hasVisibleSubpages
+		:filteredView
+		:networkOnline
+		@click="show('details')">
```
mejo-, line 22: `show('details')` is dead code now. `showing('details')` used to be used to determine if the page list or page content are displayed on mobile displays. All instances of the call should be removed, not only from this component.

src/CollectivesApp.vue
```diff
@@ -16,20 +16,30 @@
 			id="sharingToken"
 			type="hidden"
 			:value="shareTokenParam">
-		<NavigationBar v-if="!printView" />
+		<NcAppNavigation v-if="!printView">
```
mejo-: The padding (`--app-navigation-padding`) of `.app-navigation-list` looks a bit too much. In talk, they overwrite it with `.app-navigation__list { padding: 0 !important }`. I suggest we do the same.

src/components/PageList/DraggableElement.vue
mejo-, line 1: Moving pages via drag and drop feels a bit more brittle now than before. I think one aspect is that when dragging a page1 over another page2 (which would move page1 into page2), the ghost element that shows the new order place of the new page is not made transparent, which was the case in the old page list. That way it's less obvious that dropping will move page1 into page2 instead of placing it as a sibling of page2 next to page2.

src/components/PageList.vue
```diff
@@ -4,216 +4,225 @@
 -->
 
 <template>
-	<NcAppContentList :showDetails="showing('details')">
```
mejo-, line 7: The page list is no longer resizable horizontally. This used to be possible and was very handy especially for page lists with many nesting levels and long page titles.

src/components/PageList/PageListItem.vue
```diff
@@ -4,52 +4,40 @@
...
-					<div class="item-icon-emoji" :class="{ 'root-page': isRootPage }">
+					<div class="item-icon-emoji">
 						{{ emoji }}
 					</div>
 				</template>
-				<template v-else-if="isLandingPage">
-					<CollectivesIcon :size="22" fillColor="var(--color-main-text)" />
-				</template>
 				<template v-else>
 					<PageIcon :size="22" fillColor="var(--color-background-darker)" />
```
mejo-: The page icon has wrong color in light mode now as the background usually is brighter. In my tests, using `fillColor="var(--color-text-maxcontrast)` worked well in both light and dark mode.

src/components/CollectiveContainer.vue
```diff
@@ -145,10 +145,12 @@ export default {
 		...mapActions(useVersionsStore, ['selectVersion']),
 
 		async initCollective({ closeNavDelay = false }) {
-			if (closeNavDelay && !this.isPublic) {
```
mejo-, line 148: I think that the whole `closeNav` logic can go, it should be obsolete. At least `closeNavDelay` is no longer needed and can be dropped.

src/components/PageList.vue
```diff
@@ -4,216 +4,225 @@
...
+			:pageId="rootPage.id"
+			:parentId="0"
+			:title="currentCollectiveIsPageShare ? rootPage.title : currentCollective.name"
+			:timestamp="rootPage.timestamp"
+			:lastUserId="rootPage.lastUserId"
+			:lastUserDisplayName="rootPage.lastUserDisplayName"
+			:emoji="currentCollectiveIsPageShare ? rootPage.emoji : currentCollective.emoji"
+			:level="0"
+			:canEdit="currentCollectiveCanEdit"
+			:networkOnline />
```
mejo-: This (complete commit 7d65a161fe60b4d5a6a3c793bf0688e6ae392b22) seems superfluous. The CollectiveSelector is also loaded in public shares and this addition makes the landing page being listed twice 🤔

src/components/Page/PageActionMenu.vue
```diff
@@ -5,7 +5,10 @@
 
 <template>
 	<div>
-		<NcActions forceMenu @click.stop>
+		<NcActions
+			forceMenu
+			:container="inPageList ? '#app-navigation-vue' : undefined"
```
mejo-: container #app-navigation-vue seems superfluous. You removed a similar container attribution on other components after adding them in an earlier commit. Maybe this can go as well?

src/components/PageList/PageListItem.vue
```diff
@@ -58,74 +46,61 @@
...
+				<PageActionMenu
+					:pageId
+					:pageUrl="to"
+					:parentId
+					:timestamp
+					:lastUserId
+					:lastUserDisplayName
+					inPageList
+					:networkOnline />
+				<NcActions container="#app-navigation-vue">
```
mejo-: container #app-navigation-vue seems superfluous. You removed a similar container attribution on other components after adding them in an earlier commit. Maybe this can go as well?

src/components/PageList.vue
```diff
@@ -4,216 +4,225 @@
...
+	<div class="page-list-headerbar">
+		<!-- Tag selection popover -->
+		<NcPopover
+			popupRole="listbox"
+			class="page-filter"
+			popoverBaseClass="page-filter-popover"
+			:shown="showTagSelection"
+			:triggers="[]"
+			placement="bottom-start"
+			container="#app-navigation-vue"
```
mejo-: container #app-navigation-vue seems superfluous. You removed a similar container attribution on other components after adding them in an earlier commit. Maybe this can go as well?

src/components/PageList.vue
```diff
@@ -4,216 +4,225 @@
...
+							:tag
+							@select="onSelectFilterTag(tag.id)" />
+					</ul>
+				</div>
+			</template>
+		</NcPopover>
+
+		<NcActions
+			class="toggle"
+			container="#app-navigation-vue"
```
mejo-: container #app-navigation-vue seems superfluous. You removed a similar container attribution on other components after adding them in an earlier commit. Maybe this can go as well?

src/components/Nav/CollectiveSettings.vue
```diff
@@ -346,13 +350,13 @@ export default {
...
-				await this.$router.push('/')
-				emit('toggle-navigation', { open: true })
-			}
 			try {
 				await this.trashCollective(this.collective)
 				emit('collectives:navigation:collective-trashed')
+				if (this.collectiveId === this.collective.id) {
+					await this.$router.push('/')
+					emit('toggle-navigation', { open: true })
+				}
```
mejo-, line 359: This (commit bb138a639ff42e59bb985a3a1c004ffc779fdb93) looks like hiding a bug to me.

From my understanding trashing a collective should result in the next collective being loaded now, given that we don't want to open the home screen. And you already added code to HomeView.vue that loads the next available collective automatically, what error is this supposed to catch?

src/components/Nav/NewCollectiveModal.vue
```diff
@@ -277,9 +277,12 @@ export default {
...
 			.catch(displayError('Could not get list of teams'))
-		this.$nextTick(() => {
-			this.$refs.collectiveName.$el.getElementsByTagName('input')[0]?.focus()
-		})
+		const focusCollectiveName = () => {
+			this.$refs.collectiveName?.$el.getElementsByTagName('input')[0]?.focus()
+		}
+		focusCollectiveName()
+		const focusInterval = setInterval(focusCollectiveName, 50)
+		setTimeout(() => clearInterval(focusInterval), 500)
```
mejo-: Trying to steal the focus again and again every 50ms doesn't look like the right approach 😉

Wouldn't it be enough to do this once with a delay of 50ms (i.e call the callback via `setTimeout` instead of `this.$nextTick`)?

src/components/Nav/CollectiveSelector.vue
```diff
@@ -0,0 +1,428 @@
...
+			this.showSelector = false
+			this.$router.push(this.currentCollectivePath)
+		},
+
+		onChevronClick() {
+			this.showSelector = !this.showSelector
+		},
+
+		onRestoreCollective(collective) {
+			return this.restoreCollective(collective)
```
mejo-, line 200: Maybe we could navigate to the restored collective straight away? Otherwise one doesn't see that the restore was successful, as the collectives list is no longer visible.

src/components/Nav/CollectiveSettings.vue
```diff
@@ -336,6 +338,8 @@ export default {
 
 			// Push new router path if currentCollective was renamed
 			if (redirect) {
+				// Clear pages so a reload can't render a page with a stale pre-rename collectivePath
+				this.clearPagesForCollective(this.collective)
```
mejo-, line 342: This again seems like it hides a bug. I don't see how this was not necessary before but would be necessary now 😉

mejo- (changes requested):
Overall this looks really great and is a really nice improvement in my opinion. I found quite some smaller and some bigger issues, see my comments.

src/components/Nav/CollectiveSelector.vue
silverkszlo: Yes you're right. The only difference is:
- without the remaining code clicking on the collective name with Ctrl/Shift/Meta key pressed, it does not open in a new tab, but opens the popover
- then inside the popover you would have to click on the respective collective name with Ctrl/Shift/Meta key pressed to open it in a new tab

But that is not a big impediment I think and we have less code. So I agree with keeping only the last two loc.

silverkszlo: Well okay and what also happens if we only leave the two last loc:
- when you're on a different page than landing page of the collective and click on the selector, it brings you to the landing page AND opens the popover, which is a bit annoying, because you then have to close the popover first, if you want to navigate to another specific page

src/components/Nav/CollectiveSelector.vue
silverkszlo: oversight, will change it!

src/components/Nav/CollectiveSelector.vue
silverkszlo: Screenshot from 2026-08-18 15-36-06

I can do that, but it's just an empty space. When I give it a border-top or border-bottom, I still have to give it some margin, so it has equal distance to the upper or lower content. So not sure if that is really better than before.

src/components/Nav/CollectiveSelector.vue
silverkszlo: probably styling relicts from when I started developing that ;)

src/components/Nav/CollectiveSelector.vue
silverkszlo: relicts

src/components/Nav/CollectiveSelector.vue
silverkszlo: Screenshot from 2026-08-18 16-12-42
Screenshot from 2026-08-18 16-15-12

the difference is discreet but it's there

src/components/Nav/CollectiveSelector.vue
silverkszlo: Screenshot from 2026-08-18 16-20-49
That is actually hiding another bug: that the first element is selected/focues by default. I will try to resolve that differently. Thanks!

src/components/PageList/SubpageList.vue
silverkszlo: ok, but that should maybe happen in a separate PR?

src/components/PageList.vue
silverkszlo: Yes, that is because the `CollectiveSelector` and the `PageList` live inside `NcAppNavigation` now, which has no built-in resize mechanism. Navigation was moved in there to avoid the split panes and have a UI that is more consistent with other NC apps. Also two panes would not make sense with the new `NcPopover` being used for the nav. I'll add a commit, that adds custom resizing.

src/components/PageList.vue
silverkszlo: you're right, `container #app-navigation-vue` actually only has to stay in the outer `NcPopover` in `CollectiveSelector`, as `CollectivesTrash` and `CollectivesGlobalSettings` are rendered inside this same `NcPopover`. Without this container prop there though, when on mobile view, opening `CollectivesTrash` and `CollectivesGlobalSettings`, would not work. The popover ends up rendered on top of them, making them unreachable.

silverkszlo: Although it might be that I had added some of these because of failing tests. I will see how that goes :*)

src/components/Nav/CollectiveSettings.vue
silverkszlo: I don't think it's hiding a bug. Navigating to `/` before `trashCollective()` resolves means `HomeView`'s watcher fires while the store still lists the collective being trashed, so it redirects right back into it. The commit is supposed to be the adjustment needed to make the new redirect feature work correctly with the existing trash flow. I reverted it locally and reproduced the crash to confirm (console full of 404s/500s and `currentCollective` is undefined, all tracing back through `HomeView`'s redirect handler).

src/components/Nav/CollectiveSettings.vue
silverkszlo: Without this the test `"Allows to rename the collective"` in `collective-settings.spec.js` fails.

What has changed compared to before this PR: the navigation restructuring changed mount-order/timing, so that most probably creates race conditions. `CollectiveSelector` and `PageList` were mounted inside `CollectiveView` (via `NcAppContent`'s `#list` slot), as part of the same component subtree and the same mount cycle as` CollectiveContainer/PageContainer/TextEditor`.

Now in this PR they are mounted in `CollectivesApp`'s persistent `NcAppNavigation`, a structurally earlier, independent part of the tree. On a full reload that changes the relative order in which concurrent async work gets queued on the microtask/promise queue.

That's what I understand from Claude's reasoning, but it's hard to pinpoint the exact cause for race conditions..

src/components/Nav/CollectiveSelector.vue
silverkszlo: That does not seem to be such a nitpick, as I can not find an easy way to move the `CollectivesTrash` component without visually moving it out from its row in the dropdown/popover-list. The trigger button would be relocated out of the dropdown and I don't think that's the intent? I would leave this as is for now and we can discuss refactoring this in a follow-up issue.

silverkszlo:
@jancborchardt I think I've addressed all of your remarks. Only the issue with the Collectives app icon as fallback stays open :)

blizzz (approved):
As discussed, merging with green CI 🥳

PR #2683, opened 2026-08-07
feat: improve PHP rector integration

mejo-:
### 📝 Summary

* chore(deps): move rector to a dedicated vendor-bin directory
* chore(rector): update config, use nextcloud/rector
* chore(php): run 'rector process'
* test(ci): run rector in CI to enforce compliance

### 🏁 Checklist

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -9,7 +9,6 @@ declare(strict_types=1);
 
 namespace OCA\Collectives\AppInfo;
 
-use Closure;
 use OCA\Circles\Events\CircleDestroyedEvent;
 use OCA\Circles\Events\EditingCircleEvent;
 use OCA\Collectives\CacheListener;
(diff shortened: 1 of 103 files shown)


nickvergessen (approved):

lib/AppInfo/Application.php
```diff
@@ -149,13 +148,11 @@ public function register(IRegistrationContext $context): void {
...
 
-		$context->registerService(SluggerInterface::class, function () {
-			return new AsciiSlugger();
-		}, false);
+		$context->registerService(SluggerInterface::class, fn () => new AsciiSlugger(), false);
 	}
 
 	public function boot(IBootcontext $context): void {
-		$context->injectFn(Closure::fromCallable([$this, 'registerMountProvider']));
+		$context->injectFn($this->registerMountProvider(...));
```
CarlSchwan, line 155: This is afaik broken, ether registerMountProvider need to be public or you need to keep the `Closure::fromCallable`

mejo-: @CarlSchwan mhh, interesting. At least all tests passed, and if registering the mount provider would be broken now, I would have expected many failing Behat, Cypress and Playwright tests.

nickvergessen: Within the same class protected works just fine. I just tested this with all the injectFn() functions in Talk :thinking:

PR #2685, opened 2026-08-08
feat: Optimize users fetching by moving member management to the Circles

Koc:
## 📝 Summary

This is a continuation of the work started in #2518 and #2443 to optimize the members list in Collectives.

`nextcloud/circles` v35 now provides its own [members management UI](https://github.com/nextcloud/circles/pull/2561), so Collectives no longer needs to duplicate that functionality. This PR detects that capability and, when it is available, turns the `MembersWidget` members button into a link that opens the Circles team settings in a new tab. For older Circles versions the existing member management popup is preserved unchanged.

### Benefits
- Smaller member-management surface in Collectives.
- Faster member fetches because `fullDetails=true` is skipped when the Circles app already provides the management UI.
- No behavior change for users on older Circles versions.

### 🖼️ Screenshots
🏚️ Before | 🏡 After
---|---
image | image

## :question: Open question
- [ ] Should we preserve members list popup for members that has no access to the member management (e.g. moderators)?

image

## 🚧 TODO
- [ ] Manually test more scenarios (WIP)
- [ ] Fix Cypress test for a new behavior

## 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/src/components/Member/CurrentMembers.vue
+++ b/src/components/Member/CurrentMembers.vue
@@ -101,7 +101,7 @@ export default {
 		isCurrentUser() {
 			return (item) => {
 				return item.userId === this.currentUser
-					|| item.singleId === item.circle.initiator.singleId
+					|| item.singleId === item.circle?.initiator?.singleId
 			}
 		},
 	},
--- a/src/components/Page/LandingPageWidgets/MembersWidget.vue
+++ b/src/components/Page/LandingPageWidgets/MembersWidget.vue
@@ -38,6 +38,8 @@
 					variant="tertiary"
 					:title="showMembersTitle"
 					:aria-label="showMembersAriaLabel"
+					:href="circlesMembersUrl"
+					:target="circlesMembersUrl ? '_blank' : undefined"
 					@click="openCollectiveMembers()">
 					<template #icon>
 						<AccountMultiplePlusIcon v-if="isAdmin" :size="16" />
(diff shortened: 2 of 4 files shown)


max-nextcloud:
Thanks @Koc 

That's definitely matching what the current teams efforts are going for.

@marcoambrosini Is this what you had in mind for managing members from within an app?

mejo- (changes requested):
As far as I understood it, the plan was to provide a shared members management widget that can be injected in other apps, but to keep the possibility to manage members in Collectives without having to leave the app. And that's also what I personally would prefer.

Koc:
@mejo- I can try to do that, but need some more info:
1. do we have something similar already? where can I inspire from?
2. how UI should look like? Because for now we have quite different representation of popup in Collectives and page in Circles

mejo-:
@Koc thanks, much appreciated. I think the best would be to align with the design team (Cc @marcoambrosini), maybe even in a quick coordination call?

marcoambrosini:
> As far as I understood it, the plan was to provide a shared members management widget that can be injected in other apps, but to keep the possibility to manage members in Collectives without having to leave the app. And that's also what I personally would prefer.

Sorry for the late reply. Yes, this is exactly what I think is best.

Koc:
@marcoambrosini can you please answer to my questions, then?

> - do we have something similar already? where can I inspire from?
> - how UI should look like? Because for now we have quite different representation of popup in Collectives and page in Circles

PR #2741, opened 2026-09-01
Feat: Add possibility do deny Collectives download per role

Koc:
### 📝 Summary

This is reincarnation of the #1163. In our organization™ we are really need to forbid download of the collectives data. So this PR aims to make it configurable in the same manner like we can configure sharing permissions.

:construction: WIP

#### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🚧 TODO

- [ ] ...

### 🏁 Checklist

### 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI tools
- [ ] The AI-generated content was reviewed, comprehended and tested by a human


tests/Integration/features/bootstrap/FeatureContext.php
```diff
@@ -74,8 +74,10 @@ public function userCreatesCollective(string $user, string $collective, ?string
 		if ($fail === 'fails') {
 			$this->assertStatusCode($user === 'guest' ? 403 : 400);
 		} else {
-			$this->assertStatusCode(200);
-			$this->assertCollectiveLevel($collective, 9);
+			$this->assertStatusCode([200, 400]);
```
Koc: it seems like we need to remove 400 from here

PR #2451, opened 2026-04-28
fix(versions): fix duplicate entry for key 'c_p_versions_uniq_idx'

mejo-:
A race condition in requests could lead to this error as two requests tried to create version entries for the same fileId + timestamp combination.

Fix the bug by catching these errors and handling them gracefully.

Fixes: #1886
Fixes: #2405

### 🏁 Checklist

- [x] Code is properly formatted (`npm run lint` / `npm run stylelint` / `composer run cs:check`)
- [x] [Sign-off message](https://probot.github.io/apps/dco/) is added to all commits

Issue #2405, opened 2026-04-08
Unique violation error in oc_collectives_p_versions during Desktop Client synchronization


avsob:
**Describe the bug**
Every time the Nextcloud Desktop Client synchronizes changes to a file within a Collective (either creating a new file or modifying an existing one), a DbalException (Unique violation) is triggered in the server logs.
This happens for both .md and other file types (e.g., .docx) placed inside the Collectives directory.
The issue does not occur when creating or editing files via the Web UI. It seems like the Desktop Client triggers multiple rapid requests (PUT/PROPPATCH), causing a race condition where the Collectives app tries to insert the same version metadata into the database simultaneously.

**To Reproduce**
1. Open a local folder synchronized with a Collective.
2. Create a new file (e.g., test.docx or test.md).
3. Wait for the Desktop Client to sync the file.
4. Check the Nextcloud server logs and PostgreSQL logs.

**Expected behavior**
The synchronization process should complete without triggering database exceptions in the server logs.

**Server details:**
- Collectives app version: 4.2.0
- Nextcloud version: 32.0.8 (Hub 25 Autumn)
- PHP Version: 8.3.30
- Database: PostgreSQL 15/16/17

**Client details:**
 - OS: Windows
 - Desktop Client version: 33.0.2

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
DbalException An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "c_p_versions_uniq_idx" DETAIL: Key (file_id, "timestamp")=(XXXXXX, XXXXXXXXXX) already exists.
```

#### PostgreSQL log
```
INSERT INTO "oc_collectives_p_versions" ("file_id", "timestamp", "size", "mimetype", "metadata") VALUES ($1, $2, $3, $4, $5);
ERROR: duplicate key value violates unique constraint "c_p_versions_uniq_idx"
DETAIL: Key (file_id, "timestamp")=(XXXXXX, XXXXXXXXXX) already exists.

```

</details>

avsob:
Updating Collectives app version to 4.3.0 - the problem persists

Schubi2024:
Hello,

I have a similar problem. When Collectives is enabled, the Nextcloud client (33.0.3) checks for folder changes almost every second. Then it says “Everything is synchronized,” and immediately starts checking for folder changes again.

[...]

PR #2743, opened 2026-09-03
Fix: fix duplicate entry during indexation (alternative approach)

### LINKED ISSUE #1886: Duplicate entry '...' for key 'c_p_versions_uniq_idx'
**Describe the bug**

The following error log appears in nextcloud.log:

> Error: An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '364233-1752587859' for key 'c_p_versions_uniq_idx'

**Server details:**
- Collectives app version: 3.0.1

Koc:
### 📝 Summary

This is alternative approach to fix #1886, #2405. Even after #2451 we observe tons of logs on PostgreSQL side related to duplication. In this PR this logs are gone

### 🏁 Checklist

### 🤖 AI (if applicable)

### DIFF
--- a/lib/Db/TInsertIgnoreConflict.php
+++ b/lib/Db/TInsertIgnoreConflict.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Collectives\Db;
+
+use OCP\AppFramework\Db\Entity;
+use OCP\AppFramework\Db\SnowflakeAwareEntity;
+
+/**
+ * Helper trait to insert entities while ignoring conflicts on unique constraints
+ */
+trait TInsertIgnoreConflict {
+	/**
+	 * Creates a new entry in the db from an entity, ignoring conflicts on
+	 * unique constraints
+	 *
+	 * @param Entity $entity the entity that should be created
+	 * @return int number of inserted rows (0 if a conflicting row already exists)
+	 * @throws \OCP\DB\Exception
+	 */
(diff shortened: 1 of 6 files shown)


mejo-:
@Koc thanks, this looks very promising. I didn't even know about `insertIgnoreConflict()`. Could you adjust `VersionsBackend.php` as well and replace the try/catch with it?

Koc:
@mejo- sure, give me few days and I will finalize current PR

lib/Db/TInsertIgnoreConflict.php
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * Creates a new entry in the db from an entity, ignoring conflicts on
+	 * unique constraints
+	 *
+	 * @param Entity $entity the entity that should be created
+	 * @return int number of inserted rows (0 if a conflicting row already exists)
+	 * @throws \OCP\DB\Exception
+	 */
+	public function insertIgnoreConflict(Entity $entity): int {
+		if ($entity instanceof SnowflakeAwareEntity) {
```
mejo-, line 28: Can you elaborate on why you introduced this trait instead of just using `$this->db->insertIgnoreConflict()` directly within the mapper functions? As far as I know, we don't use snowflake IDs yet in Collectives, and I'd prefer to not add code handling this specific case before we introduce snowflake IDs.

mejo- (changes requested):
Thanks @Koc. One question/comment.

lib/Db/TInsertIgnoreConflict.php
Koc: `$this->db->insertIgnoreConflict()` works with plain tables/arrays of columns=>values. But I want to continue using entities like it was before. That's why I've introduced `insertIgnoreConflict` on a mapper level that accepts entity object and uses `$this->db->insertIgnoreConflict()` under the hood.

PR #2765, opened 2026-09-09
chore(test): text with contexts

max-nextcloud:


blizzz:
AttachmentResolver needs to be adjusted yet, according to Claude:

> 1. Collectives renders preview/print through window.OCA.Text.createEditor({ fileId, useSession: false }) (src/composables/useReader.ts:118) → Text's MarkdownContentEditor, which does new AttachmentResolver({ fileId }).
> 2. AttachmentResolver (unchanged on the branch) POSTs { documentId: <fileId> } to /apps/text/attachments.
> 3. Old middleware read that param as a file id: $fileId = (int)$request->getParam('documentId') → FileService::getDocumentIdForUser($fileId, $userId).
> 4. New middleware (SessionMiddleware::assertUserOrShareToken) treats it as a primary key: documentService->getDocument($documentId), and bails out with InvalidSessionException → HTTP 403 when it's null.
> 5. Document ids are now snowflakes, so that lookup can never succeed with a file id.

PR #2767, opened 2026-09-09
fix(pages): don't crash and don't lose fresh page state on stale page list responses

### LINKED ISSUE #2167: Adding new page results in "Page with ID xxxx not found"
**Description**

While creating a new Page in an existing collective the described error "Page with ID xxxx not found" is shown up.

Image

After some page reloads the page exists and is editable.

**To Reproduce**
1. Go to an existing Collective
2. Click on '+' for "add an page"
3. See error

**Expected behavior**
The new Page should be available immediately.

**Screenshots**

**Server details:**
Nextcloud version: 32.0.2.2
collectives: 3.4.0
PHP-Version: php8.3
mariadb: 10.11.14

**Client details:**
 - OS: Debian Linux
 - Browser: Firefox
 - Browser version: 140.6-esr
 - Device: desktop

```
no errors
```

#### Browser log
The interesting thing: as soon as I have opened the browser development console it's not reproducable anymore.

patches-bniemczyk:
## Problem

Opening, creating, renaming or moving pages could intermittently flip the UI to "Page with ID xxxx not found" (#2167), even though the page exists. Root cause is two-fold:

1. **Racing page list requests.** Concurrent/stale `GET /pages` responses (CollectiveView background poller, CollectiveView mount, TextEditor mount) replaced `allPages[index]` wholesale, dropping pages that had just been created/renamed/moved. The last-landed response won, even when it was initiated *before* a local mutation.

2. **Null-unsafe current-page getter chain.** `currentPageIds`, `pagePath`, `pageFilePath`, `pageDavPath`, `isIndexPage` and `title` dereferenced pages without null checks, so any transient gap turned into a hard `TypeError` (e.g. "can't access property 'id', e.currentPage is undefined") and rendered the "not found" empty view until the next poll repopulated the list.

## Fix

- Sequence page list requests per index and drop responses that were superseded by a newer request or invalidated by a local mutation (create/rename/move/trash/restore and single-page updates). Bumping all known indices on invalidation is intentional: worst case it costs one extra refetch. Dropped responses still release the loading flag.
- Make the current-page getter chain fail soft while the page list is (re)loading, so a transient gap can no longer throw or render the "Page with ID xxxx not found" empty view.

## Verification

Logic-tested with a standalone Pinia-emulated harness (7/7): `createPage`/`renamePage`/`trashPage` survive stale page-list responses; the current-page getter chain never throws on transient gaps; happy-path behavior unchanged. Reproduces the #2167 create-page race scenario.

Fixes #2167

### DIFF
--- a/src/stores/pages.js
+++ b/src/stores/pages.js
@@ -18,6 +18,17 @@ const STORE_PREFIX = 'collectives/pinia/pages/'
 // used in `PagePicker.vue`
 export const ALL_PAGES_STORE_NAME = STORE_PREFIX + 'allPages'
 
+// Request sequences per page list index. Page list responses of superseded requests
+// must not be applied, otherwise they overwrite newer local state (e.g. a page that
+// was just created or renamed) until the next refetch.
+const getPagesSequences = {}
+
+function invalidatePagesRequests() {
+	for (const index of Object.keys(getPagesSequences)) {
+		getPagesSequences[index] = (getPagesSequences[index] ?? 0) + 1
+	}
+}
+
 export const usePagesStore = defineStore('pages', {
 	state: () => ({
 		// Uses `collectiveId` as index for internal collectives and `share_<shareToken>` for public ones
@@ -115,7 +126,7 @@ export const usePagesStore = defineStore('pages', {
 				|| rootStore.pageParam === INDEX_PAGE
 		},
 		isIndexPage(state) {
-			return state.currentPage.fileName === INDEX_PAGE + PAGE_SUFFIX
+			return state.currentPage?.fileName === INDEX_PAGE + PAGE_SUFFIX
 		},
 
 		rootPage(state) {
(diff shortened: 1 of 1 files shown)


patches-bniemczyk:
Closing temporarily while we complete live testing of the fix against a production instance; will reopen with test results.

patches-bniemczyk:
**Live verification added:** we reproduced this bug's root cause on a production instance and validated the fix end-to-end.

- Reproduction fingerprint from the browser console on Collectives 3.4.1 / NC 30.0.17: `TypeError: Cannot read properties of undefined (reading 'fileId')` at `pages.ts:202/214/124/248` (via `currentPagePath` → `currentPageId`) — same crash chain as reported here.
- Server-side was verified innocent: page file etag/mtime unchanged through the failures, API polls always clean.
- The patched build (3.4.1 + this backported fix, built with the project's vite pipeline) was installed on the production instance: rapid create/rename/trash page sequences that previously flipped the UI to "Page with ID xxxx not found" within seconds now run indefinitely with a clean console.

Happy to provide anything else that helps.

patches-bniemczyk:
**Additional verification (post-rollout):** after installing the patched build on the production instance, the full server-side write path was also exercised — page file create / overwrite / delete via the API (HTTP 204s). One note for transparency: during rollout we initially hit 500s on writes caused by an incomplete `vendor/` directory in *our hand-built deployment package* (an interrupted composer run) — a packaging mistake on our side, unrelated to this patch; copying the vendor directory from the official 3.4.1 appstore package resolved it. The patched app code itself is byte-identical to the release except for `src/stores/pages.js`.

PR #2786, opened 2026-09-16
ci(collectives): test comparisons on Nextcloud 35

hweihwang:
### Summary

Run the full comparison suite on Nextcloud 35 in one CI job, alongside the existing stable32 and master jobs.

Check the server and Text versions, rebuild Text assets for comparison tests, and keep browser reports separate from server diagnostics. The server checks also apply to the general Chromium tests. An explicit base URL still uses an external server.

Allow ten minutes to build the pinned Text revision and cancel superseded Playwright runs. Wait for editor readiness and apply password-share settings through HTTP so the running server sees them immediately.

Text version updates follow in #2789.

### Tests

The complete stack passed TypeScript checks, lint, the production build and all 116 unit tests locally. All 69 comparison checks and all three page-share checks passed on both NC35 and master with retries disabled.

### AI

- [x] The content of this PR was partly or fully generated using AI tools.
- [x] The AI-generated content was reviewed, comprehended and tested by a human.


mejo- (changes requested):
@hweihwang I don't understand the purpose of this PR. As far as I understand it, it adds a lot of complexity to run the tests against Nextcloud 35. If we'd want to do this, we could just add 'stable35' to the `server-branch` matrix in `playwright.yml`, no? But I don't think this is necessary, as there's nothing special to test on stable35 that we don't already test when running the tests against the server master branch (which becomes Nextcloud 36), or am I missing something?

